PR #4, opened 2020-02-20
Added content to selectContext

lagartoverde:
I added the option to filter on content on the selectContext function, it tries all the other properties it already matched and then the content.
I also added a way to include regex in the attributes matching


lagartoverde:
I tested it with different properties and different values, and all the plugins are working (I think) but my knowledge of the editor is limited so maybe I missed something

nvdk:
merged into development

PR #6, opened 2020-03-12
added delete handler

lagartoverde:
Added delete handler, just to make it not work and prevent bubbling


lagartoverde:
I fixed everything :)

PR #13, opened 2020-03-30
Add helper function to easen hints removal

madnificent:
plugins remove all of the hints for the region they are currently scanning.  Doing this with the old API requires some basic logic, but it's repeated everywhere.

## What is this?

This PR introduces a new method to easen the burden.  It removes the hints for all the regions which an array of rdfaBlocks cover.  The
implementation itself is rather silly and could be improved over time, yet plugins do become easier to read by it.

The commit originated from the cleaning effort of lblod/ember-rdfa-editor-wikipedia-slug-plugin#1 and from simplifications in the [ember-rdfa-editor-plugin-generator](https://github.com/lblod/ember-rdfa-editor-plugin-generator).

## Example

The standard way of implementing a plugin service would be transformed into something that looks much easier to thing about.  There are less surprises in the following code because the natural flow of the code is also the one which works:

    execute(hrId, rdfaBlocks, hintsRegistry, editor) {
      hintsRegistry.removeHintsInRdfaBlocks( rdfaBlocks, hrId, "hello-hint-scope" );

      for( const rdfaBlock of rdfaBlocks ){
        if( ... ) {
          hintsRegistry.addHint( hrId, "hello-hint-scope", {
            location: absoluteLocation,
            card: "editor-plugins/hello-hint-card",
            info: { ... }
          });
        }
      }
    }

## To discuss

  - The name of this method is fairly long.  If we could find a shorter or easier to remember name that would be nice.
  - Also introducing a `removeHints` method which detects the received argument and is smart about region vs rdfaBlock would make the code a lot easier to read.  Can we extend the current PR to include this or is that over-the-top?

### DIFF
--- a/addon/utils/rdfa/hints-registry.js
+++ b/addon/utils/rdfa/hints-registry.js
@@ -292,6 +292,32 @@ export default class HinstRegistry extends EmberObject {
   }
 
   /**
+   * Removes hints which occur in a set of rdfaBlocks.
+   *
+   * This is a helper method which makes functional plugins a lot
+   * easier to write.  These plugins remove all hints as a starting
+   * point, calculate where the hints should be, and supply all of the
+   * new hints to the HinstRegistry.  In case there are no performance
+   * issues to recalculate the hints, this approach is easy to
+   * understand.  This particular call cleans up such plugins.
+   *
+   * @method removeHintsInRdfaBlocks
+   *
+   * @param {Array} rdfaBlocks Objects for which the region is in the
+   * `region` property.
+   * @param {Object} [hrIdx] Index of the registry else registry at
+   * current index is provided
+   * @param {string} [identifier] Name of the plugin to filter on
+   *
+   * @public
+   */
+  removeHintsInRdfaBlocks(rdfaBlocks, hrId, identifier) {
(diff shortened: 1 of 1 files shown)


nvdk:
short summary of our out-of-band discussion:

We need to check and see if overloading arguments is not confusing to javascript developers. It's common in other languages, but not so much in javascript. I concur it does seem like a nice thing to have. I suggest to merge this as-is for now and we create a new PR for removeHints if and when we decide overloading is common enough in javascript. If you'd like to keep track of this, perhaps we should create an issue?

madnificent:
Let's ask for suggestions on the naming first, then merge it tomorrow as is if nothing clean comes out of the discussion.

nvdk:
a suggestion based on a pattern that I have seen more in javascript (and something we already do with update):

```
removeHints({ rdfaBlocks, region, hrId, pluginId})
```

so using an object as argument, where either rdfaBlocks or region needs to be specified.

madnificent:
That looks like a nice option. Will probably need to be rather informative to the user when content is lacking or unexpected content is supplied.  I'll draft a tryout implementation and push it on this branch to see how it feels.

PR #15, opened 2020-04-16
Filter outer on blocks and parents

claire-lovisa:
When using the `outer` scope of the `selectContext` method, we were only checking the blocks given by marawa's context scanner. Those blocks are strictly contained in the range given by the `selectContext` method. It can be an issue as we could want the parents of those blocks when filtering with an outer scope.

This PR uses the current blocks and their parents when filtering with an `outer` scope.

/!\ Needs to be merged after [this PR](https://github.com/lblod/marawa/pull/4) is merged and bumped in the editor. /!\


nvdk:
closing this, since we haven't merged it after 13+ months

PR #24, opened 2020-08-04
Feature/semantic model info

lagartoverde:


nvdk:
going to close this, as I assume we might go for a different approach with vdom in the future

PR #36, opened 2020-10-30
Feature/escape handler

lagartoverde:
solves https://binnenland.atlassian.net/browse/GN-1706.


addon/components/ce/content-editable.js
```diff
@@ -159,7 +160,8 @@ export default class ContentEditable extends Component {
                                    new IgnoreModifiersHandler({rawEditor}),
                                    new UndoHandler({rawEditor}),
                                    new BoldItalicUnderlineHandler({rawEditor}),
-                                   new FallbackInputHandler({rawEditor})
+                                   new FallbackInputHandler({rawEditor}),
```
cecemel, line 163: By not being the last one in the list, will it still full fill what it says it does?

cecemel (changes requested):
I think unit tests would be handy. Please test different scenarios, given what you now know how the editor may be used. (multiple editors in on one screen)

nvdk (approved):

PR #37, opened 2020-11-11
Feature/list delete

abeforgit:
Improves the behavior of the delete button in and around lists
https://binnenland.atlassian.net/browse/GN-1847


addon/utils/dom-helpers.ts
```diff
@@ -413,6 +413,31 @@ function isVisibleElement(element: HTMLElement): boolean {
...
+  if (!editor) throw new Error("Editor element not found in dom");
+  return editor;
+}
+
+/**
+ * Utility to get the selection in a type-safe way. A null selection only happens when called on a
+ * hidden iframe in Firefox, so it is ok to throw an error here instead of nullchecking everywhere
+ * see https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection
+ */
+function getWindowSelection(): Selection {
```
nvdk, line 431: love this! we might want to move these kind of helpers to it's own selection helper file

addon/utils/dom-helpers.ts
```diff
@@ -413,6 +413,31 @@ function isVisibleElement(element: HTMLElement): boolean {
   return !!(element.offsetWidth || element.offsetHeight || element.getClientRects().length);
 }
 
+/**
+ * Utility to get the editor element in a type-safe way
+ * This avoids having to nullcheck everywhere where a null editor would be an error anyway.
+ */
+function getEditorElement(): Element {
```
nvdk, line 420: if this is only used in tests (guessing here), this file is probably not the right location

nvdk (commented):
I'd prefer not changing existing code styles when they're used consistently. E.g. I think we use the following consistently.
```
if {

} 
else {

}
```
It makes code reviews more tedious. 

For the case syntax I quite like the `case: { }` syntax you're using, I was not familiar with it. Do note that in one case you've removed `break` statements whereas in the other you kept it, I assume you can remove it from both? General question: does this syntax also imply the variables within the brackets are locally scoped?

Will review/test the delete plugin itself in a next comment :)

abeforgit:
yes, this makes the variables locally scoped, that's the reason for the brackets
It's a recommendation from eslint
https://eslint.org/docs/rules/no-case-declarations

I removed break statements only where the previous statement was a return statement (so they would never be reached)
In all other scenarios we should always put breaks. These situations were detected by the typescript language server (unreachable code warning). I think it's best to maintain the habit of always putting breaks, and relying on tooling to tell us when it's not needed.

The if-else syntax was rearranged by the prettier autoformatter, apologies for not being more careful with it
However I must say I strongly dislike that if-else style... not that it really matters

But I think it would be good to setup an autoformat config so that we dont have to worry about this stuff anymore... case in point: while maybe ruining the if-else convention, it also did at lots of missing semicolons and made spacing consistent everywhere

addon/utils/dom-helpers.ts
abeforgit: well so far it is, but I thought it could also be useful elsewhere

nvdk: I'm a bit wary, because typically it would be best to retrieve the element using rawEditor.rootNode . the selection is a bit brittle imo.

abeforgit: ah yeah now that I look at it again its not ideal
I'll make a testutils file

addon/utils/plugins/lists/delete-plugin.ts
```diff
@@ -0,0 +1,243 @@
...
+      return this.guidanceForRemoveEmptyElement(manipulation);
+    } else if (manipulation.type == "removeEmptyTextNode") {
+      return this.guidanceForRemoveEmptyTextNode(manipulation);
+    }
+    return null;
+  }
+
+  detectChange(): boolean {
+    if (this.hasChanged) {
+      this.hasChanged = false;
```
nvdk, line 56: this is the first time I've seen state in a plugin, not necessarily an issue but we have to make sure the calling code takes this into account.

abeforgit: yeah I don't like it, but man the change detection was driving me nuts
it's much easier to detect the change within the function thats doing the changes
Definitely to be improved though, this kind of implicit messaging is asking for trouble

abeforgit:
oh..... well I wish I'd have found those sooner

cecemel:
I am sorry, I made some more general overview, because for such big changes looking for commits made the image in my head to fragmented. This is only a partial review. Tomorrow I hope to find some extra time

# general
- Cool it was a good read!
- Please bear in my, I just make remarks, but I am often wrong too. sorry :-)
- Personally, I like more line length in the code. Is it here 80? (I don't think we have a convention though)

# specific
- Why is state introduced here? Someting unexpected for me. (could be sensible, but now I need more scrunity)
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L39
  I am afraid if we have some unexpected errors, and there is no cleaning logic, we might end up with broken plugin
  Why can't we rely on what is there, or similar to other plugin?
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L54
  (look e.g. at what backspace does)
- So suggestion, think we have a helper for this:
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L74
- Do we miss a case when the next node after a list is not an element?
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L68
- I think the same here, what if last li and no element immedialty afterwards
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L93
  Tested it:
  ```
  <ul><li>|</li><ul>felix<div>ruiz</div>
  ```
   is funky whehn you press delete
- I am unsure why we need a specific flow for this:
   https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L111
   Should this not just be the normal handler doing this and you end up in the next iteration in moveCursorAfterElement?
- I am unsure wether empty lists are legal.
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L145
- Why is RemoveElementWithNoVisibleChildren not a special case? <li>[invisibletextspaces]</li>
- Will this not fail for case
  [emptyTextNode][fooTextNode]<li>... (a bit the same as previous remark)
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L145
- I don't think we want in a guidance already DOM manipulation
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L130
- Does your linter complain if ```nextElement!``` is put in the if check above?
  https://github.com/lblod/ember-rdfa-editor/blob/feature/list-delete/addon/utils/plugins/lists/delete-plugin.ts#L134
-

abeforgit:
why did your comment show up only now :o
added numbers to you comment to make it easier to talk about

# General

3. linelength is the default from prettier, so 80
personally I hate having to scroll horizontally, and linelength 100 is somehow always just a bit too much for 2 files next to eachother
my eyes are not great so I need a relatively large font size
but I could live with 100

# specific

1. why state:
change detection is difficult. delete usually doesn't delete the node itself, but it's next sibling, so the backspace strategy doesn't work
detecting changes in the amount of children of the parent list also doesnt work, because of magicspans
I had a whole bunch of logic in the detectChange function before,  and it still didn't work properly. In the interest of actually getting something done I decided to detect the change where is was actually detectable, aka next to the code that was actually changing things...

2. im in the process of converting things to use helpers
3. see below
4. I'll look at that case, yes I didnt consider it

5 - 10: ill check these

11. I dont see where I'm doing dom manip in the guidance
12. we dont have to check nextElement, tagName returns an empty string when the element is null so we know that it's there
however typescript does not, so thats why theres an exclamation mark

abeforgit:
5. this case:
```
<ul>
<li>|</li>
</ul>
<div>abc</div>
```
expected:
```
<ul>
<li>|abc</li>
</ul>
```
actual when no specific handling is added:
```
<ul>*</ul>
<div>abc</div>
```
with * = zero-width space

abeforgit:
3 and 4 are handled now
6 Im not sure what you mean exactly
7-10 Ill try to test this tomorrow

abeforgit:
I think all the comments are handled now
it still behaves weird when there are newlines in the source html, but afaict that's an issue with the delete handler itself and not with the list plugin

nvdk (approved):

PR #38, opened 2020-11-17
added standard/extended paste tags

Asergey91:


addon/components/ce/content-editable.js
```diff
@@ -322,9 +322,14 @@ export default class ContentEditable extends Component {
     const clipboardData = (event.clipboardData || window.clipboardData);
 
     //TODO: if no clipboardData found, do we want an error?
-    if (this.features.isEnabled('editor-html-paste') && this.hasClipboardHtmlContent(clipboardData) ) {
+    if ((this.features.isEnabled('editor-html-paste')||
+         this.features.isEnabled('editor-extended-html-paste'))&&
```
nvdk, line 326: shouldn't the check on extended paste be removed here?

Asergey91: wouldn't that cause the user to supply 2 options for the extended functionality?
It seems to me more intuitive to have either options turn on html paste.

nvdk: that makes sense, I misread at first and thought you used the && operator :+1:

PR #41, opened 2020-11-26
Bugfix/lump node is editable

cecemel:
See commit message

### DIFF
--- a/addon/utils/ce/movement-observers/lump-node-movement-observer.js
+++ b/addon/utils/ce/movement-observers/lump-node-movement-observer.js
@@ -57,7 +57,7 @@ function jumpOverLumpNode(node, rootNode, editor) {
     nextNode = document.createTextNode('');
     element.after(nextNode);
   }
-  nextNode = ensureValidTextNodeForCaret(nextNode);
+  nextNode = ensureValidTextNodeForCaret(nextNode); //TODO: move this in set.caret or other uniform API somehow
   editor.updateRichNode();
   editor.setCaret(nextNode, 0);
 }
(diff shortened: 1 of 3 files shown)


nvdk (changes requested):
as discussed we should only use the moveCaret helper if we specifically don't want to change editor state, such as in the delete and backspace handlers where we do consecutive moves before updating editor state. This is rare and not the case here. `setCaret` both moves the caret and sets editor state and should be used here.

addon/utils/ce/movement-observers/lump-node-movement-observer.js
```diff
@@ -1,43 +1,78 @@
...
+  const element = getParentLumpNode(node, rootNode);
+  let textNode;
+  if(element.previousSibling && element.previousSibling.nodeType == Node.TEXT_NODE){
+    textNode = element.previousSibling;
+  }
+  else {
+    textNode = document.createTextNode('');
+    element.before(textNode);
+  }
+  textNode = ensureValidTextNodeForCaret(textNode);
```
nvdk, line 75: there is no need to call ensureValidTextNodeForCaret, as setCaret should take care of that.

cecemel: I think there are still issue related to the implementation with this logic in setCaret.
 Mainly related to whitespace textnodes next to elements displayed as block and the caret not wanting to appear in there. Very visible if the lump node is the only node in the editor, and start clicking in it. Cursor will not follow (or display weirdly in FF). (Toggle the line, you'll see)

And setCaret relies heavily on richnode and fixing it there,might potentially break other stuff, I consider this as out of scope to merge both, and I suggest of wrapping this in a TODO. (I guess the approach will also change quite a bit with native handling and the virtual DOM limiting the possible cases)

PR #42, opened 2020-11-27
Improve cursor behavior after deletes in lists

abeforgit:


addon/utils/dom-helpers.ts
```diff
@@ -1,6 +1,7 @@
 import { A } from '@ember/array';
 import { PernetSelection, PernetSelectionBlock } from '@lblod/ember-rdfa-editor/editor/pernet'
 import { RichNode } from '@lblod/ember-rdfa-editor/editor/raw-editor'
+import { isEmptyRdfaAttributes } from './rdfa/rdfa-rich-node-helpers';
```
nvdk, line 4: was this import missing or where is it coming from?

addon/utils/dom-helpers.ts
```diff
@@ -442,6 +443,119 @@ function getWindowSelection(): Selection {
...
+}
+/**
+ * Get the bounding clientRect for the caret.
+ * This also handles the cases where the caret is in a weird place
+ * and therefore has no clientRect by default.
+ * The returned rect should always be interpreted as being in front of the caret.
+ * So when using this to store a location and later resetting it
+ * use the "right" property as the x value.
+ * */
+function getCaretRect(): ClientRect {
```
nvdk, line 487: I'm very exited about this bit, this could possibly really enhance our change detection in both the backspace and delete handler. bit worried about some state changes it introduces though, such as introducing  an extra invisibleSpace.

addon/utils/plugins/lists/delete-plugin.ts
```diff
@@ -149,21 +153,23 @@ export default class ListDeletePlugin implements DeletePlugin {
...
+    const mergeNode = this.getDeepestLastDescendant(node);
+
+    const cursorPos = getCaretRect();
+    const selection = getWindowSelection();
+    selection.removeAllRanges();
     const nodeToMerge = this.getDeepestFirstDescendant(targetNode);
     this.mergeNodes(mergeNode, nodeToMerge);
 
-    this.repositionCursor(mergeNode, nodeToMerge, cursorPosition);
+    setCaretOnPoint(cursorPos.right, cursorPos.bottom - cursorPos.height/2 )
```
nvdk, line 171: I suppose the delete handler is responsible for making sure the editor state is up to date with this change.

nvdk (approved):
functionally this is working very well in both chrome and firefox! If you or felix can confirm the editor state is updated at the end of a delete loop, feel free to merge as far as I'm concerned

addon/utils/dom-helpers.ts
abeforgit: I have no idea, probably an accidental auto-import

addon/utils/dom-helpers.ts
abeforgit: yeah I hate that part, but it's the most consistent way I found to get the coordinates when the caret is in a weird position (which happens about 50% of the time)

abeforgit:
Quite a few bugs popped up, this is not ready for merging

abeforgit:
might be made obsolete by #45

PR #43, opened 2020-11-30
don't indent/unindent when not in list

nvdk:
this only addresses (un)indenting for collapsed selections, as true selections do not throw an error at this time.


abeforgit (approved):
Looks sensible

PR #44, opened 2020-12-02
Feature/lumpnode delete

abeforgit:
Also has changees from #42 so best to wait for that one before merging


abeforgit:
included in #45

PR #45, opened 2020-12-09
Feature/native handling

abeforgit:
So this is the magic pr that will fix all our delete worries.

The aim of this is to let the browser handle the delete key in almost all cases, and use the plugin architecture we already have to handle any exceptions we might want. 

Having done it both the manual way and this way, I can say that this way is a million times easier. The browsers we support are actually very smart in how they handle things like lists and other nested nightmares. In fact it does exactly what we want for lists, to the point the entire list plugin is just gone. The lumpnode plugin was kept with only minor changes needed (and those might have been bugs with the old handler as well).

The delete handler itself also got about half it's code thrown out. I've never been a huge fan of the "just keep deleting until something changes" approach (although it is a clever way to look at it, it just never really worked _for_ me in my delete struggles)

I believe that with a sensible fallback as we have now, we can avoid the whole hassle of looping to begin with. When we press delete, the handler runs exactly once, and does what it needs to do. Running once in this case means: Run it by the plugins, which can decide whether the event has been fully handled or not, then run it by the fallback delete handler code (there is a small remnant of our old code here for dealing with the removeEmptyTextnode manipulation, but it makes sense to have this step in general in my opinion) and then let the browser do what it likes. Each layer should be able to signal that the event has been handled and that the browser is not allowed to do anything.

This has some really nice benefits as the delete handler is now entirely synchronous, but mainly all the change detection logic can just dissapear, as we now commit to making sure the handler does the right thing on the first go.

This is quite a drastic change and not a drop-in replacement for handlers that already work. With the delete I had the benefit of starting from something which could hardly be made worse. I do think this approach has huge value going forward, especially in those areas where the current behavior is very broken or nonexistent (e.g. the arrow keys)

I think even the functioning handlers can benefit from this approach, but with those of course we have to be more careful not to introduce regressions.

One other thing is that at the moment basically all tests concerning delete are broken. Kind of funny, since visually it works so much better. This shows how hard these things are to test accurately. I will go through the tests tomorrow.


addon/editor/input-handlers/delete-handler.ts
```diff
@@ -674,101 +325,102 @@ export default class DeleteHandler implements InputHandler {
     // check where our cursor is and get the deepest "thing" after
     // the cursor (character or node)
     const thingAfterCursor: ThingAfterCursor = this.getThingAfterCursor();
-    console.log(thingAfterCursor)
+    console.log(thingAfterCursor);
```
nvdk, line 328: should probably replace this with a call to `editorDebug` or remove the statement all together

addon/editor/input-handlers/delete-handler.ts
```diff
@@ -869,7 +521,7 @@ export default class DeleteHandler implements InputHandler {
               const textNode = child as Text;
               if (stringToVisibleText(textNode.textContent || "").length == 0) {
                 return {type: "emptyTextNodeStart", node: textNode};
-              } else {
+
```
nvdk, line 524: an else branch disappeared here ;)

nvdk (changes requested):
This looks really promising and indeed seems to work quite well on firefox (need to do some more extensive testing though). 

+ I love that we can get rid of the async handler, this has always felt risky. 
+ the amount of complexity that we now externalize is probably a good thing for the short to mid long term
- I think the current implementation leaves the editor state (currentSelection etc) out of sync if the handler takes care of the delete. I don't see a call to setCaret and updateRichNode (or the fallback updateSelectionAfterComplexInput() which also takes care of these things) and these are required if a handler handles an event.

addon/utils/plugins/lump-node/delete-plugin.ts
```diff
@@ -0,0 +1,124 @@
...
+} from "../../ce/lump-node-utils";
+import {
+  getCaretRect,
+  setCaretOnPoint,
+  findFirstAncestorWhichSatisfies,
+  findDeepestFirstDescendant,
+} from "../../dom-helpers";
+import { RawEditor } from "@lblod/ember-rdfa-editor/editor/raw-editor";
+
+const SUPPORTED_MANIPULATIONS = [
```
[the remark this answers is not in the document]
abeforgit: not mine, this was stolen from the backspace lumpnode plugin :)

addon/editor/input-handlers/delete-handler.ts
abeforgit: no idea why that else branch is gone :o

abeforgit:
not relevant anymore because of the model

PR #47, opened 2021-01-19
Bugfix/first use css

Dietr:
This should fix the issue with the CSS not being included in the `vendor.css` when installing the addon.
I think this was because the `index.js` was missing the CSS import include statement.

Took some inspiration from https://github.com/cibernox/ember-power-select/blob/master/index.js to check if Sass is used.

My initial tests seem to be working but some further testing and code review seems necessary to check if the blueprint and import is done correctly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-  name: require('./package').name;
+  name: require('./package').name,
 
   included() {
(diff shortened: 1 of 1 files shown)


nvdk (approved):
that is quite an awesome find, I would not have found this myself I think :). I've pushed a small fix that you may want to check out, but other than that looks to be working correctly.

PR #49, opened 2021-01-21
Feature/improved selection

abeforgit:
The selection now has a 1-to-1 mapping to the domselection. Positions are represented by paths from the root, in anticipation for future expansion.

Completes https://binnenland.atlassian.net/browse/GN-2037

As an added bonus, also seems to fix https://binnenland.atlassian.net/browse/GN-2036
This was expected, but still nice to see.


abeforgit:
This also reverts back to using the dom elements as keys for the nodeMap, since that didnt seem to be the cause of the issue.
We use a WeakMap now for the nodemap, which should prevent memory leaks coming from destroyed nodes that we still have in the nodeMap

abeforgit:
> This also reverts back to using the dom elements as keys for the nodeMap, since that didnt seem to be the cause of the issue.

@nvdk @lagartoverde  please verify this when reviewing

abeforgit:
Also added https://binnenland.atlassian.net/browse/GN-2038 here cause its just a one-line change

addon/utils/ce/model-selection-tracker.ts
```diff
@@ -35,117 +21,68 @@ export default class ModelSelectionTracker {
...
-      return {
-        textNode,
-        offset: type === 'anchor' ? 0 : textNode.length
-      };
-    }
-  }
-
 }
 
+// if (currentSelection.type != 'Caret' && currentSelection.focusNode !== this.model.rootNode && currentSelection.focusOffset === 0 && currentSelection.anchorNode) {
```
nvdk, line 38: commented out code should probably go

addon/utils/ce/pernet-raw-editor.ts
```diff
@@ -138,12 +138,7 @@ export default class PernetRawEditor extends RawEditor {
    * @param args
    */
   executeCommand(commandName: string, ...args: any[]) {
-    const command = this.registeredCommands.get(commandName);
-    if(!command) {
-      throw new Error(`Unrecognized command ${commandName}`);
-    }
-    command.execute(...args);
-    this.updateRichNode();
+    super.executeCommand(commandName, ...args);
```
nvdk, line 141: it seems we lost updateRichNode here and I think we need to call it before performing generateDiffEvents ?

nvdk (changes requested):
very happy with the tests and in some user testing I didn't find any immediate regressions other than the "triple click" selection behaving a bit weirdly if you apply a text property afterwards. I think this probably needs some more extensive testing, but so far (apart from the small inline remarks) seems to work quite well and can be merged.

addon/utils/ce/pernet-raw-editor.ts
abeforgit: no, the super method also calls updateRichnode

abeforgit:
So I seem to have solved the paragraph selection problems for now
The issue was the ambiguity of dom selections, as usual. I'm not very happy with the solution, it feels too ad-hoc. 
I see two better for the long term:
- blindly follow the dom selection anchor/offset choices, and expect commands to deal with edge cases such as the triple-click (but provide helpers to do so relatively easily)
- guarantee a disambiguous and well-specified modelSelection

My gut feeling is to go for the latter, but I do not have such a specification at the ready, it would require some though and experimenting. It may become more clear what we need when more commands are implemented and selection patterns emerge naturally.

nvdk:
I've addressed the small bug we discussed, seems to work ok still (tested on firefox). Merging now.

PR #50, opened 2021-01-28
Feature/list insertion topdown

abeforgit:
Adds the make-list and remove-list commands

completes
https://binnenland.atlassian.net/browse/GN-2040
https://binnenland.atlassian.net/browse/GN-2041
https://binnenland.atlassian.net/browse/GN-2042


abeforgit:
Testing needed to see if it also fixes the other list-related stories

abeforgit:
- if you start typing right after creating a list from an empty line, it inserts above the list and not inside of it
- extra newlines being added
- complex nested lists not handled well in list removal
- sometimes selection doesn't recognize it's no longer in a list after deletion
- broken ul elements are sometimes left after intense list creation/deletion sessions
- sometimes the entire document ends up inside of a list

abeforgit:
theres a weird bug where the first character you type after making a list appears in the wrong location, I think this has something to do with the text-insertion handler being out of sync? cause visually the caret is in the right location

nvdk:
> theres a weird bug where the first character you type after making a list appears in the wrong location, I think this has something to do with the text-insertion handler being out of sync? cause visually the caret is in the right location

That's somewhat odd, because I believe the text-insertion handler works on the dom selection, see also https://github.com/lblod/ember-rdfa-editor/blob/master/addon/editor/input-handlers/text-input-handler.ts#L184

abeforgit:
this is the case where the li does have children though

nvdk:
I'm going to merge this as is in dev, we can track improvements in a new PR if that's ok with you

addon/components/rdfa/editor-toolbar.hbs
```diff
@@ -46,15 +46,15 @@
       </button>
       {{/if}}
       {{#if @showListButtons}}
-      <button type="button" class="say-toolbar__button" {{on "click" this.insertUL}} data-test-button-id="unordered-list">
+      <button type="button" class={{concat "say-toolbar__button" (if this.isInList "active")}} {{on "click" this.toggleUnorderedList}} data-test-button-id="unordered-list">
```
nvdk, line 49: it's probably best to not use concat here. there proper way of doing this is:
`<button class="say-toolbar__buton {{if this.isInList "active"}}" ...>`

abeforgit: ah ok ill change it, I actually thought concat was better

abeforgit:
> I'm going to merge this as is in dev, we can track improvements in a new PR if that's ok with you

sounds good to me

PR #53, opened 2021-02-04
Improve list removal behavior

abeforgit:
Vastly improved list removal logic, now properly handles nested lists
Probably still missing edge cases, but I think this is in the "usually does what it should, and otherwise doesnt blow up" stage

But please track any issues you find with list removal in the comments here

### DIFF
--- a/addon/components/rdfa/editor-toolbar.hbs
+++ b/addon/components/rdfa/editor-toolbar.hbs
@@ -64,7 +64,7 @@
       </button>
       {{/if}}
       {{#if @showIndentButtons}}
-      <button type="button" class="say-toolbar__button" {{on "click" this.insertUnindent}} data-test-button-id="insert-unindent">
+      <button type="button" class="say-toolbar__button {{if this.isInList "is-active"}}" {{on "click" this.insertUnindent}} data-test-button-id="insert-unindent">
         <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 44 44" class="say-icon say-icon--large" aria-hidden="true">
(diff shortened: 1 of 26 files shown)


abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2152

nvdk:
Did some testing and this is definitely an improvement! I'm just testing list removal as this is the main concern for this PR. A general remark is that the pattern "remove list" -> "reapply list" -> "remove list" is pretty broken for me. Currently testing in firefox.

Now on to the issues I still encounter:

given the list of `case 20` under lists in the dummy app:
* select part the second list item, all of it's sublist items and the final list item. 
remove list will not remove the item from the top list, this was unexpected to me, but may be by design. at least it behaves very consistently and I could argue this to be expected behaviour :).

* putting the cursor on  "subitem 1" and pressing remove list
results in undesirable behaviour for sub item 2, which I don't think would be resolved by merging sibling lists.

addon/components/rdfa/editor-toolbar.hbs
```diff
@@ -64,7 +64,7 @@
       </button>
       {{/if}}
       {{#if @showIndentButtons}}
-      <button type="button" class="say-toolbar__button" {{on "click" this.insertUnindent}} data-test-button-id="insert-unindent">
+      <button type="button" class="say-toolbar__button {{if this.isInList "is-active"}}" {{on "click" this.insertUnindent}} data-test-button-id="insert-unindent">
```
nvdk, line 67: This seems counter intuitive, the button is not active when in a list. I think you want `class={{unless this.isInList "disabled"}}`, as well as a disabled state on the button itself

abeforgit: I think this is changed in the indent/unindent prs, so I'd leave it as is and see how it behaves after all merges

Asergey91:
not sure that I have anything meaningfull to add to be honest.
cases like these seem to be wierd
when removing list item 6 
it seems to unindent item 7 to the most parent ul
this seemed strange to me but then i checked on ck editor and it seems to do the same.
Screenshot from 2021-02-08 15-00-35

abeforgit:
thanks for trying that out, not a case I tested. But yeah the way it handles it now seems sensible, I don't think there is a "natural" behavior for this case which everyone can agree on.

abeforgit:
@nvdk I think all the cases you mentioned are handled now, and it's now adding br's where needed much more consistently. I think it's ready to be merged

nvdk (approved):

PR #54, opened 2021-02-04
Feature/list indent

abeforgit:
This implements list indentation as it functions in CK editor: namely that any selection which contains the first item of a list cannot be indented. This prevents for example that a list with only one item can be indented.

I think it's as sensible a strategy as any, and it makes things a lot simpler to implement.

contains and should be merged after #53 

https://binnenland.atlassian.net/browse/GN-2096


abeforgit:
Oh in order to check if the indent button should be displayed, it uses a new `editor.canExecuteCommand` function, which in turn uses `command.canExecute` which every command implements (the default implementation is to return `true`)


To me it makes sense that a command should be able to define in which environment it makes sense, and that this information can be queried to inform the user.

Ive also modified `editor.executeCommand` to check if `canExecute` returns true before executing, I'm not sure this is the right way to go. We might want to allow the option to ignore this check and try to execute anyway, although by definition this should basically never produce any sensible results.

abeforgit:
Another note: this implementation can end up in some less than ideal (but not actually incorrect) html

In particular, indenting a selection of 2 listitems vs indenting them one by one gives different results. 
I think this should be fixed, but for that I want to implement a `mergeAdjacentLists` utility as per discussion in the lblod-core-editor channel

PR #55, opened 2021-02-05
Feature/unindent list

Asergey91:


addon/commands/unindent-list-command.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  execute(selection: ModelSelection = this.model.selection): void {
+    if(!ModelSelection.isWellBehaved(selection)) {
+
+      throw new MisbehavedSelectionError();
+    }
+    //start of selection
+    const anchorNode = selection.lastRange.start.parent;
+    //end of selection
+    const focusNode = selection.lastRange.end.parent;
```
abeforgit, line 23: these are not used so they should be removed

addon/commands/unindent-list-command.ts
```diff
@@ -0,0 +1,87 @@
...
+        if(node && parent && grandParent && greatGrandParent){
+          //remove node
+          const nodeIndex=node.index!;
+          parent.removeChild(node);
+
+          //remove parent if node is only child
+          if(parent.length==0){
+            if(parent.index){
+              grandParent.addChild(node, parent.index);
+              this.model.write();
```
abeforgit, line 66: this shouldn't be necessary here since you are writing the whole model at the end already

addon/commands/unindent-list-command.ts
```diff
@@ -0,0 +1,87 @@
...
+          if(parentSplitIndex){
+            grandParent.addChild(node, parentSplitIndex+1)
+          }
+        }
+      }
+    }
+
+    this.model.write();
+    this.model.readSelection();
+    return;
```
abeforgit, line 85: the return here is unnecessary

abeforgit (commented):
Works well overall. Try to stay consistent with formatting, especially around operators
we should add spaces either side of an operator
so `const nodeIndex = node.index!` instead of `const nodeIndex=node.index!`

addon/commands/unindent-list-command.ts
Asergey91: this is an escape clause it returns if this "if" is reached hence i am writing the model. maybe there is a way to break out of nested if inside a loop in ts im am unaware of though?

Asergey91: oh i forgot you can break =) in for loops

nvdk:
Seems to work well indeed and I think you addressed all outstanding remarks!

PR #56, opened 2021-02-09
Feature/undo functionality

lagartoverde:


nvdk:
@lagartoverde perhaps a less generic name would be better for the model write event, so we're sure it doesn't overlap with existing libraries?

lagartoverde:
Maybe editorModelWrite?

PR #64, opened 2021-03-02
Updated ember to 3.20

lagartoverde:
I spent a bunch of time testing and neither seems to be broken. 
I don't know how to test the embedded editor do you think this could be broken when we change the ember version? @nvdk If is possible please comment it here and how to test it


nvdk:
for testing in embeddable: you can just npm link the library in frontend-embeddable and that should work?

lagartoverde:
I tested on embeddable and it works perfectly fine :)

PR #65, opened 2021-03-17
Feature/table insertion

lagartoverde:
First version of the table insertion, it adds 5 buttons to the toolbar (Sorry they are all equal waiting for dieter icons :P).
The first that appears adds a 2x2 table.
Then when you are inside a table 4 more appear on the toolbar
From left to right:
- Adds a row below the cursor
- Adds a row above the cursor (note you can't add a row above if you are on the first row as that's the header)
- Adds a column after the cursor
- Adds a column before the cursor


nvdk (changes requested):
I've tested in firefox and due to selection issues I could often not add rows or columns. 90% of the time I would get a "position is null" error. I've traced this to the check
```
    if(!position || !position.x)
```
that throws an error if x is zero, so that check should be improved.

2 other remarks:

1. I would prefer insert-table-column-after instead of insert-column-after (etc), as it makes it more explicit this is about a table
2. I wonder if the static methods on ModelTable should not just be utils/helpers (though I guess then the question becomes, where to put them :D).

addon/commands/insert-column-after-command.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const table = ModelTable.getTableFromSelection(selection);
+
+    if(!table) {
+      throw new Error('The selection is not inside a table');
+    }
+
+    const position = ModelTable.getCellIndex(cell);
+    if(!position || !position.x) {
```
nvdk, line 37: this check is broken

nvdk:
most other editors also place the cursor in the first cell after inserting

nvdk:
also perhaps for now, just add text instead of repeating the same icon over and over? :)

lagartoverde:
I quickly solved the problem with position is null, I added that bug trying to avoid the typescript warning and I forgot about the 0

lagartoverde:
I didn't know either where to put the functions, I just added them as static methods but will need to discuss where they should be. 
About the methods I don't have any preference, I didn't put table because I thought it might be too long, but I will add it if it's more clear :)
About the cursor, I tried not to make it more complex than it should be as we are aiming for a simple version of the table, but yeah we can move the cursor to the added row, what should happen when I add a column?
And yeah, I will add text instead of the icon :)

nvdk:
ah for adding rows/colums I would keep the cursor at the current position, I was referring to when a table was inserted. should have made that more clear :).

Dietr (commented):
Added default styling for the table and a dropdown menu for all the table functions. I didn't add icons to these yet as I'm not sure we need them in a dropdown menu.

addon/commands/insert-table-command.ts
```diff
@@ -0,0 +1,109 @@
...
+    const leftParent = leftOfStart.parent;
+    if (!leftParent) {
+      throw new NoParentError();
+    }
+
+    const table = new ModelTable(2,2);
+
+    //handle zero length selection
+    if (selection.isCollapsed) {
+      // add the break to the right of the node before the cursor
```
abeforgit, line 64: try to keep the comments up to date when copying code

addon/model/model-element.ts
```diff
@@ -25,6 +25,19 @@ export default class ModelElement extends ModelNode implements Cloneable<ModelEl
     this._type = value;
   }
 
+  set className(value: string) {
```
abeforgit, line 28: good idea! I like this

addon/model/model-table.ts
```diff
@@ -0,0 +1,117 @@
...
+    const row = cell.parent;
+    if(!row) throw Error('Table doesn\'t have the expected structure');
+    const xIndex = row.getChildIndex(cell);
+    const rowParent = row.parent;
+    if(!rowParent) throw Error('Table doesn\'t have the expected structure');
+    let yIndex;
+    if(rowParent.type === 'thead') {
+      yIndex = 0;
+    } else {
+      yIndex = rowParent.getChildIndex(row) + 1;
```
abeforgit, line 86: you can add a `!` after getChildIndex if you are sure it will never be null
if you are not sure, you should check for it and throw an error

addon/model/model-table.ts
```diff
@@ -0,0 +1,117 @@
...
+    return {x: xIndex, y: yIndex};
+  }
+
+  static getCellFromSelection(selection: ModelSelection) {
+    const commonAncestor = selection.getCommonAncestor();
+    if(!commonAncestor) return null;
+    const cell = commonAncestor.parentElement.findAncestor((node) => {
+      if(node.modelNodeType === 'ELEMENT') {
+        const element = node as ModelElement;
+        if(element.type === 'th' || element.type === 'td') return true;
```
abeforgit, line 97: try to avoid single-line if statements

abeforgit (commented):
Looks good overal! Really just some nitpicking comments, nothing important
- when cursor is in the table header, the add row above does not work, probably best to hide it
- maybe disable insert table when inside a table? It does seem to work but not sure we want to support nested tables

nvdk (approved):

addon/commands/insert-table-command.ts
lagartoverde: What's wrong about those commands? I keep them because I didn't change any code and I think they describe good what the code is doing

addon/model/model-table.ts
lagartoverde: Didn't know that trick thanks

PR #66, opened 2021-03-18
Feature/table deletion methods

lagartoverde:
Table deletion methods plus the selection handling that was asked in the other PR, I decided to do it here because I also had to do the selection handling of the removal methods.

The selection works like this:
- If you add a table it selects the first cell
- If you remove a column or row it tries to keep the selection static, it moves up or left the selection so the user doesn't notice the cursor jump too much and can keep removing
- When you remove a table the parent of that table gets a cursor in the position the table was


addon/model/model-table.ts
```diff
@@ -0,0 +1,152 @@
...
+    this.className = 'say-table';
+  }
+
+  getDimensions() {
+    const tHead = this.children[0] as ModelElement;
+    const tBody = this.children[1] as ModelElement;
+    const y = tBody.children.length + 1;
+    const firstRow = tHead.children[0] as ModelElement;
+    const x = firstRow.children.length;
+    return {x, y}
```
abeforgit, line 39: missing semicolon

abeforgit (commented):
Looks good and seems to work well

nvdk (approved):
Looks good, probably want to merge dieters changes from inserting into this branch and update the menu accordingly.

PR #67, opened 2021-03-18
Feature/offsets

abeforgit:
Changes the selection to work on offsets as defined here: https://github.com/lblod/ember-rdfa-editor/wiki/Design-Proposal-Offsets-and-Operations

Opening this as a WIP to make reviewing less crazy

Important places to look at: 
- SelectionReader.readDomPosition
- SelectionWriter.writeDomPosition
- tree-walker.ts
- set-property-command.ts
- ModelSelection.getTextPropertyStatus

currently broken:
- all list commands
- old tests

untested:
- plugins
- lumpNodes

The NodeFinder and related utility methods have been deprecated in favor of the ModelTreeWalker

Following a well-tested and documented specification is in my opinion much better than relying on my own ad-hoc algorithms
On top of that, the interface is much more flexible and allows one to implement a custom walking algorithm should they want to.


abeforgit:
cool sidenote
check out the modelText.getTextProperty method

rather than maintaining two different maps for properties and textProperties, there is only one map with two sets of get/set methods to maintain compatibility

this is also as a preparation for operations, so that we can have a single AttributeOperation which can set any attribute

addon/model/model-node.ts
```diff
@@ -156,6 +213,9 @@ export default abstract class ModelNode {
   setTextAttribute(_key: TextAttribute, _value: boolean) {
   }
 
+  /**
+   * @deprecated use {@link ModelTreeWalker} instead
```
nvdk, line 217: perhaps (though I guess that would make the PR explode) we can just remove and replace with modeltreewalker where it's used

addon/model/readers/selection-reader.ts
```diff
@@ -51,68 +54,70 @@ export default class SelectionReader implements Reader<Selection, ModelSelection
...
-      // ignore breaks to the left
-      // TODO: what about other block elements? Do we actually want to do this here, or should we expect
-      // commands to handle breaks themselves?
-      while (leftNode && tagName(leftNode) === "br") {
-        leftNode = leftNode.previousSibling;
+    return rslt;
+  }
+  private findPositionForTextPopertyNode(container: Node, domOffset: number): ModelPosition {
+    if(container.childNodes.length === 0) {
+      // this is fallback behavior in case the dom does something really weird
```
nvdk, line 88: not sure if this can never happen, are we assuming container is an element and excluding void elements here?

addon/model/util/tree-walker.ts
```diff
@@ -0,0 +1,329 @@
...
+ * Extensions include support for stopping at an end node, the implementation of the Iterable interface
+ */
+import ModelElement from "@lblod/ember-rdfa-editor/model/model-element";
+import ModelNode from "@lblod/ember-rdfa-editor/model/model-node";
+import ModelPosition from "@lblod/ember-rdfa-editor/model/model-position";
+import {NotImplementedError} from "@lblod/ember-rdfa-editor/utils/errors";
+import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
+
+export enum FilterResult {
+  FILTER_ACCEPT,
```
nvdk, line 15: it would be nice to at least document the usage of accept, skip, reject and filter here. my first assumption switched the meaning of skip and reject.

addon/model/readers/selection-reader.ts
abeforgit: this function is only called when container is a textpropertynode, i.e.. <strong>, <em>, etc

addon/model/model-node.ts
abeforgit: yeah too much cognitive load for this pr, but we should do it soon

addon/model/util/tree-walker.ts
```diff
@@ -0,0 +1,329 @@
...
+}
+
+export type ModelNodeFilter = (node: ModelNode) => FilterResult;
+
+export interface ModelTreeWalkerConfig {
+  filter?: ModelNodeFilter;
+  range: ModelRange;
+}
+
+export class ModelTreeWalker implements Iterable<ModelNode> {
```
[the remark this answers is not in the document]
abeforgit: huh, yeah it should be

abeforgit:
Trying to integrate the DOM in tests has completely defeated me
just ship the thing at this point I'm too frustrated to fix the tests

abeforgit:
nevermind, tables are broken

abeforgit:
Well that was easier than expected

nvdk:
> When inserting a table with a collapsed selection the dropdown triggers an ember error, something about "value already used in a previous computation"

we should address this, but not in this branch :)

nvdk (approved):
tested quite a bit while comparing with the dev branch, found no immediate regressions. also seems to be working ok when included in GN now.

Absolutely love the complexity that has now disapeared from the table insertion :).

PR #68, opened 2021-03-19
Close table dropdown when an action inside is clicked

Dietr:
Added a modifier to toggle the table dropdown when an action inside is clicked.

Not sure if this is the best way to do this so suggestions are welcome.


lagartoverde (approved):
Looks good to me 👍

nvdk:
Perhaps we can use the regular click modifier? e.g.
```
  <div id="{{this.dropdownId}}" class="say-dropdown__menu {{if this.dropdownOpen "is-visible" ""}}" role="menu" tabindex="-1" {{focus-trap isActive=this.dropdownOpen shouldSelfFocus=true focusTrapOptions=(hash onDeactivate=this.closeDropdown clickOutsideDeactivates=true)}} {{on "click" this.closeDropdown}}>
  ```

nvdk:
I've simplified the component a bit, I don't think we need a new modifier as the regular "click" modifier seems to work as we want. @Dietr please check if it's still in line with your expectations.

nvdk (approved):

PR #69, opened 2021-03-22
Feature/table backspace handler

lagartoverde:
Made a new backspace plugin to prevent the user from deleting the table cells

### DIFF
--- a/addon/utils/ce/raw-editor.ts
+++ b/addon/utils/ce/raw-editor.ts
@@ -41,7 +41,7 @@ import RemoveTableCommand from "@lblod/ember-rdfa-editor/commands/remove-table-c
 @classic
 class RawEditor extends EmberObject {
   registeredCommands: Map<string, Command> = new Map()
-  protected model: Model;
+  model: Model;
   protected tryOutVdom = true;
 
   /**
--- a/addon/utils/ce/raw-editor.ts
+++ b/addon/utils/ce/raw-editor.ts
@@ -27,6 +27,7 @@ import InsertColumnBeforeCommand from "@lblod/ember-rdfa-editor/commands/insert-
 import RemoveTableColumnCommand from "@lblod/ember-rdfa-editor/commands/remove-table-column-command";
 import RemoveTableRowCommand from "@lblod/ember-rdfa-editor/commands/remove-table-row-command";
 import RemoveTableCommand from "@lblod/ember-rdfa-editor/commands/remove-table-command";
+import ModelSelection from '@lblod/ember-rdfa-editor/model/model-selection';
 /**
  * Raw contenteditable editor. This acts as both the internal and external API to the DOM.
  * Any editing operations should be implemented as {@link Command commands}. External plugins can register their own commands.
@@ -93,6 +94,9 @@ class RawEditor extends EmberObject {
    */
   get rootNode() : HTMLElement {
     return this.model.rootNode;
+  }
+  get selection(): ModelSelection {
+    return this.model.selection;
   }
   set rootNode(rootNode: HTMLElement) {
     this.model.rootNode = rootNode;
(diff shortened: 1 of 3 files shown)


lagartoverde:
I made this PR based on the other PR I made so that one should be merged earlier

lagartoverde:
Nevermind, I saw the other PR was already approved so I merged it

lagartoverde:
Made a draft because I found a bug with the enter handler

lagartoverde:
I found a bug that when you press enter multiple times, and try to delete the new lines sometimes it jumps back to the line you deleted, but I found the same bug in an empty editor so I guess this PR shouldn't be in charge of fixing that

addon/utils/plugins/table/backspace-plugin.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @class TableBackspacePlugin
+ * @module plugins/table
+ */
+export default class TableBackspacePlugin implements BackspacePlugin {
+  label = 'backspace plugin for handling table nodes'
+
+  guidanceForManipulation(manipulation: Manipulation, editor: RawEditor) : ManipulationGuidance | null {
+    const voidExecutor = {
+      allow: true,
```
nvdk, line 15: why not just disallow backspace and not provide an executor?

nvdk (changes requested):

addon/utils/plugins/table/backspace-plugin.ts
lagartoverde: I don’t fully know how this plugin system works, if I pass false there then nothing will be executed?

nvdk: See https://github.com/lblod/ember-rdfa-editor/blob/master/addon/editor/input-handlers/backspace-handler.ts#L394 , yes it should stop
(in general I think having a good grasp of that backspace method will help understand the plugin system a bit more)

addon/utils/ce/raw-editor.ts
```diff
@@ -41,7 +41,7 @@ import RemoveTableCommand from "@lblod/ember-rdfa-editor/commands/remove-table-c
 @classic
 class RawEditor extends EmberObject {
   registeredCommands: Map<string, Command> = new Map()
-  protected model: Model;
+  model: Model;
```
nvdk, line 44: I believe we don't want model to be public, this probably means we should provide some wrappers on editor so the handler can get the selection.

lagartoverde: This was suggested by @abeforgit, we can just provide a getter for the selection as I think is a common use case

addon/utils/plugins/table/backspace-plugin.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    return null;
+  }
+
+  /**
+   * If the handler has been executed we had done nothing so we should always return true.
+   * @method detectChange
+   */
+  detectChange() : boolean {
+    return true;
```
nvdk, line 40: detectChange overrides default backspace behaviour, so by returning true here you're letting the backspace handler know there is a change even if it did not detect a change itself. note that detectChange is called even if we have not provided guidance.

in this case it's better to always return false I think

nvdk (changes requested):

addon/utils/plugins/table/backspace-plugin.ts
lagartoverde: I thought it only got called when we handled the case. If I return false, it will run in an infinite loop (well, not infinite because it stops after 50 iterations I think) there's no better solution than trying defaulting to false?

nvdk: ah I see! Yes in that case you will need to recognize the specific case (e.g. we're at the start of a cell) and return true only in that case.

PR #71, opened 2021-03-22
Refactor/highlights

nvdk:
builds upon #67, which should be merged first.

apply/remove highlights on the virtual dom instead of on the actual dom. removes some legacy api's from pernet-raw-editor which should no longer be in use.

outstanding work:
- [x]  (optional) use the insert-html command from rdfa-document to set content on virtual dom
- [x] for the getter of htmlContent on rdfa-document, provide a custom writer that removes highlights
- [x] currently removing highlights and immediately after modifying the dom from a plugin is broken, find a workaround see https://github.com/lblod/ember-rdfa-editor/pull/71#issuecomment-810896652


nvdk:
I should note this  PR also renamed the execute method of the SetProperty command to `setProperty`, which causes a diff in all text property related commands.

My main reason is that I needed to override the signature in the highlight command, but it also feels right because rather than a default implementation this needed to be overwritten and called in each subclass. Another option was getting rid of the sub classes all together.

nvdk:
cause currently unknown, but I did notice that in GN when inserting a "besluit" template, something is off. I end up not replacing the entire placeholder (in a draft decision). I don't really understand how that could happen atm.

nvdk:
further: it seems my element reader/writer for spans drops other properties which was unintentional

addon/model/model-text.ts
```diff
@@ -28,24 +27,28 @@ export default class ModelText extends ModelNode {
...
-  get textAttributeMap(): Map<TextAttribute, boolean> {
-    return this._textAttributeMap;
+  get isBlock() {
+    return false;
   }
 
-  set textAttributeMap(value: Map<TextAttribute, boolean>) {
-    this._textAttributeMap = value;
+  get offsetSize() {
+    return this.length;
```
nvdk, line 35: should we consider odd cases like "   abc  " here (which has only 4 visible positions, starting right before the a and ending right after c).

abeforgit: no it has 6 visible positions: 
before: space,a,b,c,space,end

nvdk: note that's not true for real text nodes though, " abc " renders as "abc", per https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace

addon/utils/rdfa/event-processor.ts
```diff
@@ -13,59 +22,36 @@ import { isEmpty } from '@ember/utils';
...
-  },
+  profile: string;
+
+  constructor({ registry, profile, dispatcher, editor} : { registry: HintsRegistry, profile: string, dispatcher: RdfaEditorDispatcher, editor: RawEditor}) {
+    this.cardsLocationFlaggedNew = [];
+    this.cardsLocationFlaggedRemoved = [];
+    this.modifiedRange = [];
+    this.registry = registry;
+    this.profile = profile;
+    this.dispatcher = dispatcher;
```
nvdk, line 52: given that this is a service, perhaps we should just `@inject` it.

addon/utils/rdfa/rdfa-document.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  get htmlContent() {
+    // TODO: this no longer removes the highlights and should be fixed
+    return this._editor.rootNode.innerHTML;
+  }
+
+  set htmlContent(html: string) {
+    // const selection = this._editor.createSelection();
+    // selection.selectRange(this._editor.createRangeFromPaths([], []));
+    // this._editor.executeCommand("insert-html", html, selection);
```
nvdk, line 26: this should be updated once the insert-html command has been merged

nvdk:
> cause currently unknown, but I did notice that in GN when inserting a "besluit" template, something is off. I end up not replacing the entire placeholder (in a draft decision). I don't really understand how that could happen atm.

As far as I can tell at the moment this is caused by new zero width text nodes that are introduced but that the hints registry is unaware off (no text difference generated?).

addon/model/model-text.ts
abeforgit: oh good god

addon/utils/text-offset-to-position.ts
```diff
@@ -0,0 +1,49 @@
+import Model from "@lblod/ember-rdfa-editor/model/model";
+import ModelNode from "@lblod/ember-rdfa-editor/model/model-node";
+import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
+import ModelTreeWalker, { FilterResult } from "@lblod/ember-rdfa-editor/model/util/model-tree-walker";
+import ModelText from "../model/model-text";
+import RawEditor from './ce/raw-editor';
+
+const VOID_NODES = ["area","base","br","col","command","embed","hr","img","input","keygen","link","meta","param","source","track","wbr"];
+
+function selectTextNodes(node: ModelNode) {
```
abeforgit, line 10: maybe call this selectTextAndVoidNodes?

addon/utils/text-offset-to-position.ts
```diff
@@ -0,0 +1,49 @@
...
+  const treeWalker = new ModelTreeWalker({ filter: selectTextNodes, range });
+  let startOffset = 0;
+  for (const node of treeWalker) {
+    if (ModelNode.isModelText(node)) {
+      let endOffset = startOffset + node.length;
+      if (offset >= startOffset && offset <= endOffset) {
+        const path = node.getOffsetPath();
+        const delta = offset - startOffset;
+        path[path.length -1] = path[path.length-1]+delta;
+        return path
```
abeforgit, line 39: missing semi
some other minor formatting in this loop
also endOffset can be const

addon/model/readers/wrapped-attribute-reader.ts
```diff
@@ -17,16 +18,25 @@ export default class WrappedAttributeReader implements Reader<HTMLElement, Fragm
       ["em", "italic"],
       ["u", "underline"],
       ["del", "strikethrough"],
+      ["span", "highlighted"]
```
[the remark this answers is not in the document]
nvdk: yes indeed, I tried it in a different way with it's own reader but really struggled :/

addon/model/model-text.ts
nvdk: That sounds like a plan, we should take care though when normalizing. Take splitting text nodes for example. If the resulting nodes start or end with a space we probably want to convert that a non breaking space (if we don't already do that) so it shows up. (e.g. split "I walk my dog" into "I walk&nbsp;" "my dog"). on normalizing I suppose we want to convert that non breaking space back to a normal space so we have "I walk my dog" instead of "I walk&nbsp;my dog")

nvdk:
found the issue with removing highlight, I was not passing the selection :facepalm: . I still believe it's a bit brittle (in that if remove-highlight or make-highlight accidentally introduce textual changes things will break), but seems to work as well as the old implementation for now. 

Definitely needs extensive testing!

abeforgit:
do you think it's testable?
we could test the highlight command and make sure no textual changes are made there but ideally the plugin wouldnt break too hard when it does happen?

nvdk:
> do you think it's testable?
> we could test the highlight command and make sure no textual changes are made there but ideally the plugin wouldnt break too hard when it does happen?

Should be testable atm and I think we can have it merged as is. What will happen when remove-highlight makes textual changes is that in some cases make-highlight might be called with an incorrect range (and thus the highlight will be off a bit).

addon/utils/ce/raw-editor.ts
```diff
@@ -1,34 +1,37 @@
...
-import MakeStrikethroughCommand from "@lblod/ember-rdfa-editor/commands/text-properties/make-strikethrough-command";
-import RemoveStrikethroughCommand from "@lblod/ember-rdfa-editor/commands/text-properties/remove-strikethrough-command";
-import MakeUnderlineCommand from "@lblod/ember-rdfa-editor/commands/text-properties/make-underline-command";
-import RemoveUnderlineCommand from "@lblod/ember-rdfa-editor/commands/text-properties/remove-underline-command";
-import MakeListCommand from '@lblod/ember-rdfa-editor/commands/make-list-command';
-import RemoveListCommand from '@lblod/ember-rdfa-editor/commands/remove-list-command';
-import UnindentListCommand from "@lblod/ember-rdfa-editor/commands/unindent-list-command";
-import InsertNewLineCommand from "@lblod/ember-rdfa-editor/commands/insert-newLine-command";
 import IndentListCommand from "@lblod/ember-rdfa-editor/commands/indent-list-command";
+import InsertHtmlCommand from '@lblod/ember-rdfa-editor/commands/insert-html-command';
```
nvdk, line 4: sorry looks like it sorted on auto import

addon/utils/plugin-editor-api.js
```diff
@@ -18,6 +20,17 @@ export default class PluginEditorApi {
     this._hrId = hrId;
   }
 
+  executeCommand(name, ...args) {
```
nvdk, line 23: already added these so plugins can use the virtual dom in the future, not used anywhere at the moment

addon/utils/rdfa/event-processor.ts
```diff
@@ -128,24 +114,35 @@ export default EmberObject.extend({
...
    */
   handleRegistryChange(/*registry*/) {
-    const editor = this.get('editor');
-    editor.clearHighlightForLocations(this.get('cardsLocationFlaggedRemoved'));
-
-    this.get('cardsLocationFlaggedNew').forEach(location => {
-      editor.highlightRange(location[0], location[1]);
-    });
+    let region;
+    while (region = this.cardsLocationFlaggedRemoved.shift()) {
```
nvdk, line 118: this was important as calling remove-highlight might in some cases trigger textual differences, if that happens `handleRegistryChange` will be called before this loop finishes. so need to make sure the card is removed from the array to prevent infinite loops.

abeforgit:
> > do you think it's testable?
> > we could test the highlight command and make sure no textual changes are made there but ideally the plugin wouldnt break too hard when it does happen?
> 
> Should be testable atm and I think we can have it merged as is. What will happen when remove-highlight makes textual changes is that in some cases make-highlight might be called with an incorrect range (and thus the highlight will be off a bit).

I mean testable as in with unit tests, since its about plugins and async im a bit uncertain

nvdk:
ah in that way, currently that seems hard at least.

abeforgit:
First thing I noticed is that backspace breaks easily in and around highlights (cursor gets to a state where backspace doesn't do anything, usually clicking resolves it)

This is probably not really in scope for this PR though

addon/model/writers/html-export-writer.ts
```diff
@@ -0,0 +1,39 @@
...
+      }
+    } else if (ModelNode.isModelText(modelNode)) {
+      result = this.htmlTextWriter.write(modelNode);
+    } else {
+      throw new WriterError("Unsupported node type");
+    }
+
+    if (!result) {
+      result = new Text();
+      this.model.bindNode(modelNode, result);
```
abeforgit, line 34: should we bind here when we are exporting html? seems weird

addon/utils/global-text-offset-to-path.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+/**
+ * This matches text offsets as found in the TextNodeWalker to a position in our virtual dom
+ * as such it assumes void nodes have a "width" of 1, being either a newline for breaks and
+ * a space for all other void nodes
+ *
+ */
+export default function globalTextOffsetToPath(editor: RawEditor, offset: number): number[] {
```
abeforgit, line 25: Now that I look at this again it seems a bit weird for this to take in a whole RawEditor
a root and an offset make more sense I think? also makes it way easier to test

addon/utils/rdfa/event-processor.ts
abeforgit, line 118: Eslint doesn't like this trick, personally I'm indifferent towards it

abeforgit: also, pop is O(1) and shift is O(n), so if the order doesn't matter pop is better

addon/model/writers/html-export-writer.ts
nvdk: hmm no that's definitely an oversight

addon/utils/global-text-offset-to-path.ts
nvdk: hmm yes, the helper can just import ModelRange to create a range from paths.

addon/utils/rdfa/event-processor.ts
nvdk: I'm removing them in the order they were added I think (so first in first out), for the size of the array there will be no difference in O(1) vs O(n). IIRC the difference between shift and pop only shows up on arrays in excess of 100.000 elements.

nvdk:
>     1. First thing I noticed is that backspace breaks easily in and around highlights (cursor gets to a state where backspace doesn't do anything, usually clicking resolves it)
> 
> 
> This is probably not really in scope for this PR though

This could be related to https://binnenland.atlassian.net/browse/GN-2476 . While I would not like to block this PR because of backspace bugs, we should make sure we don't heavy to many regressions.
> 
>     1. Changing the besluittype seems to break highlights, but only visually (the yellow dissapears, but clicking inside "voeg nieuw artikel in" still triggers the hint)

A quick check on dev.gelinkt-notuleren.lblod.info tells me this was already the case, so not related to this PR.

addon/utils/rdfa/event-processor.ts
nvdk: wrt to the assignment in the loop predicate, I'm fine with having a different predicate and doing the assignment inside the loop as well if that makes the linter happy :).

abeforgit:
I took the liberty of implementing the suggestions we both seemed to agree on in the interest of moving this thing forward

abeforgit:
This commit fixes the outer cause of the backspace not working. With this gone we can see what I think is the "real" error which has to do with a Fragment element staying in the model tree somehow. I think this is really the point where a rewrite of the reader logic becomes a priority.

abeforgit:
> This commit fixes the outer cause of the backspace not working. With this gone we can see what I think is the "real" error which has to do with a Fragment element staying in the model tree somehow. I think this is really the point where a rewrite of the reader logic becomes a priority.

I will tackle this is a separate branch however, since I want to use the improvements to the tests to make sure we get this right

PR #72, opened 2021-03-26
Feature/insert html command

Asergey91:
shouldn't be merged until feature/offsets is merged


addon/commands/insert-html-command.ts
```diff
@@ -0,0 +1,39 @@
...
+    super(model);
+  }
+  execute(htmlString: string, selection: ModelSelection = this.model.selection) {
+    const range = selection.lastRange;
+    if(!range){
+      return;
+    }
+    const parser = new DOMParser();
+    const html = parser.parseFromString(htmlString, "text/html");
+    const bodyContent=html.body.children[0];
```
nvdk, line 32: We should support multiple top nodes and assume some of those topnodes can be text nodes. So

- use body.childNodes to also include textNodes
- loop over all childNodes

nvdk (changes requested):
This works for simple html, but throws an error if the selected html consist of only textnodes. It also only considers the first element of the html if we have multiple top nodes.

nvdk:
@Asergey91 can you have a look?

addon/commands/insert-html-command.ts
```diff
@@ -0,0 +1,39 @@
...
+import ListCleaner from "@lblod/ember-rdfa-editor/model/cleaners/list-cleaner";
+import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
+import {ModelTreeWalker} from "@lblod/ember-rdfa-editor/model/util/tree-walker";
+import ModelPosition from "@lblod/ember-rdfa-editor/model/model-position";
+import HtmlReader from "@lblod/ember-rdfa-editor/model/readers/html-reader"
+import ModelText from "../model/model-text";
+
+
+/**
+ * command will convert all nodes in the selection to a list if they are not already in a list
```
nvdk, line 17: this comment is out of line with the implementation

Asergey91: yep fixed

nvdk (approved):
Looks good to me! When called from paste I see my html appearing... and then disappearing again, but that seems to be a separate issue :)

PR #73, opened 2021-03-30
Feature/table tab handler

lagartoverde:


addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @class TableBackspacePlugin
+ * @module plugins/table
+ */
+export default class TableTabInputPlugin implements TabInputPlugin {
+  label = 'backspace plugin for handling table nodes'
+
+  guidanceForManipulation(manipulation: Manipulation, editor: RawEditor) : ManipulationGuidance | null {
+    const selection = editor.selection;
+    if(selection.isInTable) {
```
nvdk, line 17: this is always truthful as it has the string "disabled" as value when disabled. As a result the tab handler executor will throw an error every time the plugin is called outside of a table

nvdk (changes requested):
This was not properly tested outside of a table, it breaks the tab handler outside of tables.

nvdk:
perhaps consider allowing the tab handler to jump out of the table when you press tab in the last cell?

lagartoverde:
Now the problem I see with allowing the table to go outside is that we cannot go inside, so it feels like the navigation is broken, maybe the solution is to fix the tab handler outside of the table to jump to the first cell when it encounters a table. I have uploaded the code so you can test what it happens

nvdk (approved):
Looks ok to me, in firefox at least it is jumping in the table as well. though perhaps not to where we want it to jump? This could be a further improvement.

lagartoverde:
In firefox it looks like it's jumping inside the table at the correct position but it's not if you press tab again it kinda breaks but I agree the objective of the task was to provide navigation inside the table we can fix that later on another task

nvdk:
@lagartoverde  can you resolve the merge conflicts and merge?

PR #74, opened 2021-03-30
Hide list button when you are in a table

lagartoverde:


addon/components/rdfa/editor-toolbar.hbs
```diff
@@ -45,7 +45,7 @@
         </span>
       </button>
       {{/if}}
-      {{#if @showListButtons}}
+      {{#if (and @showListButtons canInsertList)}}
```
nvdk, line 48: Please make sure to properly scope variables in templates. It should be `this.canInsertList` in ember octane.

nvdk (approved):
Looks good, tiny syntax remark.

PR #75, opened 2021-03-30
Feature/fix tests

abeforgit:
Doing this in a separate pr cause there could be some hidden issues

This fixes the non-determinism in the tests, or rather avoids it completely by slightly adjusting the way the model is set up.
The problems all centered around the window.getSelection() method, and by extension the modification of the testing-dom in the tests. Qunit and ember-qunit documentation is extremely sparse on this topic. (can you tell I'm really not a qunit fan)

The good news is that we don't really need the selection at all to perform model tests. We can work with a Range instead, which we can create on detached document tree (i.e. one that is not rendered at all, the result of a simple `document.createElement()` without attaching the node anywhere). Since the logic for converting the domselection immediately and trivially drops down to its ranges, we can be confident that this will be enough for almost all usecases.

The changes needed for this were mainly the ability to create a model without a tracker, and to do a model.read without also reading the selection. Both are added in this PR.

Because this changes the way the model is initialized, and also moves the selectionTracker into the RawEditor instead of the model, it's possible something broke on the boundary between pernet and vdom code.

Sidenote: I am still not happy with the way the `RawEditor` and the model are initialized, but I think it will do for now.

depends on #67

### DIFF
--- a/addon/utils/ce/raw-editor.ts
+++ b/addon/utils/ce/raw-editor.ts
@@ -28,6 +28,7 @@ import RemoveTableColumnCommand from "@lblod/ember-rdfa-editor/commands/remove-t
 import RemoveTableRowCommand from "@lblod/ember-rdfa-editor/commands/remove-table-row-command";
 import RemoveTableCommand from "@lblod/ember-rdfa-editor/commands/remove-table-command";
 import ModelSelection from '@lblod/ember-rdfa-editor/model/model-selection';
+import ModelSelectionTracker from "@lblod/ember-rdfa-editor/utils/ce/model-selection-tracker";
 
 /**
  * Raw contenteditable editor. This acts as both the internal and external API to the DOM.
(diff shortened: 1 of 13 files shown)


addon/utils/ce/raw-editor.ts
```diff
@@ -122,10 +139,15 @@ class RawEditor extends EmberObject {
...
-      this.updateRichNode();
+    try {
+      const command = this.getCommand(commandName);
+      if (command.canExecute(...args)) {
+        this.getCommand(commandName).execute(...args);
+        this.updateRichNode();
+      }
+
+    } catch (e) {
+      console.log(e);
```
nvdk, line 150: can perhaps be a console.error, don't know what we prefer here.

nvdk (approved):
tested on GN, doesn't seem more broken than dev right now. So OK to merge for me!

nvdk:
ah the error in the enter handler does seem to happen a bit more, might be a coincidence.

abeforgit:
yea the enter handler is p much broken atm, im fixing it in a separate branch

PR #77, opened 2021-04-02
Bugfix/fix enter

abeforgit:
not really a "fix" more like a bandaid
newline seems to work fine and uses operations
newLi is way too complex to handle
added tests to document and confirm current behavior
one test breaks because the actual behavior is so far off from desired I can't let it pass in good conscience
Unsure if this is a regression or if the case was already broken
in any case the whole command needs a rewrite


abeforgit:
Sidenote this adds a move operation, which works but is probably not useful yet, and is not tested properly
I left it in because the work to abstract the operation algorithms into their own utility class is useful

PR #79, opened 2021-04-08
Refactor/reader

abeforgit:
- Adds tests to the reader
- Adds the dom and domStripped template tags for easy test writing
- reworks the reader logic to be fully functional (each reader invocation is responsible for its entire subtree, and readers maintain no state). State is managed in a context class that is passed along with every read() call. Readers are responsible for cleaning up context they set themselves if necessary.
- fixes the selectionReader so it recognizes a highlight span as a textPropertyNode (so it handles it the same way as strong, em, etc)
- reworks the selectionReader logic to hopefully be more robust in finding a suitable position when it encounters textPropertyNodes

This is all in service of trying to get backspacing into highlighted sections to work properly. That goal is not fully realized in this PR, but I think I went as far as I could without rewriting the backspace handler. It seems backspacing is more consistent, and in most cases "just keep hitting backspace" eventually leads to deleting what you want.
builds on #71 and #78


lagartoverde (approved):
Looks good 👍

nvdk (approved):

PR #80, opened 2021-04-12
Bugfix/init problems

abeforgit:
Solves https://binnenland.atlassian.net/browse/GN-2514
builds on #79


lagartoverde (commented):
This will remove the crashing, but will still show an error each time it gets initialized, couldn't we detect somehow when it's not yet initialized so we can avoid throwing errors that doesn't mean anything?

abeforgit:
it doesnt show an error, it shows a warning
and yes, detecting initialization is the better option, but afaict that's basically impossible at the moment

I still want to keep the warning at least so we are aware

lagartoverde (approved):
Okay, then we can merge it :)

PR #81, opened 2021-04-13
Implement property cleaner

abeforgit:
Implements https://binnenland.atlassian.net/browse/GN-2524


abeforgit:
will also convert the listcleaner to operate on ranges (or at least give it a decent try) so we are consistent

addon/model/util/model-tree-walker.ts
```diff
@@ -26,14 +26,26 @@ enum TraverseType {
...
+
 export interface ModelTreeWalkerConfig {
   filter?: ModelNodeFilter;
   range: ModelRange;
   descend?: boolean;
   visitParentUpwards?: boolean;
 }
 
-export default class ModelTreeWalker implements Iterable<ModelNode> {
+export default class ModelTreeWalker<T extends ModelNode = ModelNode> implements Iterable<T> {
```
nvdk, line 48: ah I can use this in my offset to range mapper :dancers:

abeforgit: its still a cast though so be careful
but yea very useful when your filter only outputs one type

abeforgit: its why the nodefinder used to have both a filter and a predicate
(filter was for types, predicate for other stuff)
but I found it too confusing and usually redundant

addon/commands/make-list-command.ts
```diff
@@ -45,10 +45,12 @@ export default class MakeListCommand extends Command {
 
     this.model.change(mutator => {
       mutator.insertNodes(range, list);
+      const resultRange = mutator.flush();
+      const fullRange = ModelRange.fromInElement(this.model.rootModelNode, 0, this.model.rootModelNode.getMaxOffset());
```
nvdk, line 49: I guess we should optimize this range at some point.

addon/model/cleaners/property-cleaner.ts
```diff
@@ -0,0 +1,30 @@
...
+import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
+import ModelTreeWalker, {toFilterSkipFalse} from "@lblod/ember-rdfa-editor/model/util/model-tree-walker";
+import ModelText from "@lblod/ember-rdfa-editor/model/model-text";
+import ModelNodeUtils from "@lblod/ember-rdfa-editor/model/util/model-node-utils";
+
+export default class PropertyCleaner {
+  clean(range: ModelRange) {
+
+    const textNodes = new ModelTreeWalker<ModelText>({range, filter: toFilterSkipFalse(ModelNode.isModelText)});
+    // careful, we are modifying the nodes in the iterator inside the loop
```
nvdk, line 11: appreciate the comment here!

addon/model/util/model-node-utils.ts
```diff
@@ -0,0 +1,30 @@
+import MapUtils from "@lblod/ember-rdfa-editor/model/util/map-utils";
+import ModelNode from "@lblod/ember-rdfa-editor/model/model-node";
+import ModelElement from "@lblod/ember-rdfa-editor/model/model-element";
+import {LIST_CONTAINERS} from "@lblod/ember-rdfa-editor/model/util/constants";
+
+export default class ModelNodeUtils {
+  static DEFAULT_IGNORED_ATTRS: Set<string> = new Set(["__dummy_test_attr", "__id"]);
```
nvdk, line 7: probably best to add `data-editor-position-level` and `data-editor-rdfa-position-level` to this list.

Both are used to make it possible to style the current `active` subtree based on your position, e.g. if you're cursor is inside an element it will have `data-editor-position-level`=1, it's parent will have 2 etc. but should not block two nodes from merging

addon/model/util/model-tree-walker.ts
```diff
@@ -26,14 +26,26 @@ enum TraverseType {
 
 export type ModelNodeFilter = (node: ModelNode) => FilterResult;
 
+export function toFilterSkipFalse(func: (node: ModelNode) => boolean): ModelNodeFilter {
+  return function (node: ModelNode) {
+    return func(node) ? FilterResult.FILTER_ACCEPT : FilterResult.FILTER_SKIP;
```
nvdk, line 31: some heredoc on these would be nice

nvdk (approved):
Some minor remarks, but should be ok to merge.
 
I ran into quite a few list issues while testing, but I don't think it's related to the list cleaner as the behaviour is mostly the same on the dev branch.

nvdk:
@abeforgit can you have a look at the outstanding remarks and fix the merge conflict?

PR #82, opened 2021-04-13
added ember intl and translated the table commands

lagartoverde:


nvdk (approved):
looks good to me!

abeforgit (commented):
For me even if my browser is set to english it still shows the dutch version. Same thing whether I run inside GN or not

nvdk:
> For me even if my browser is set to english it still shows the dutch version. Same thing whether I run inside GN or not

Was that in chrome? In firefox it did seem to work correctly for me, though I did have to refresh after changing my preferred language.

lagartoverde:
How are you setting your language?, for me in chrome I had to use a browser extension because it picked up the system language by default

nvdk:
on chrome it does seems broken when you have multiple languages set:
Screenshot from 2021-04-15 17-13-11

nvdk (changes requested):
does not seem to work when multiple locales are set in chrome

nvdk (approved):

PR #83, opened 2021-04-15
Feature/simplify annotation css

Dietr:
Quick fixes on the annotation hover design. These are temporary fixes but should improve selection/deleting of text and provide a more relevant annotation hover for decisions.
- Show less annotations on hover (if no content value/translation is set, nothing is shown > easier to override in GN)
- Make the annotation sidebar more compact
- Remove block styling on inline elements
- Small fixes to the flagged annotations: more relevant translations and a fix on selecting the text (after element was blocking selection)


nvdk:
image

@Dietr The "vul in" is quite prominent now and wasn't before, not sure if you intended this?

nvdk (approved):
don't see the "vul in" issue in ember-rdfa-editor itself, so ok for me!

PR #84, opened 2021-04-15
Feature/table limit selections

Asergey91:
- added limited selection inside tables
- there was a lot of code written using vdom stuff but then most of it was scrapped and the solution was reversed engineered from redactor wysiwyg editor
- the only thing really left from the vdom code are some utility functions, they also have tests


nvdk:
Looks ok to me, also seems to solve a tab issue that kevin reported.

PR #85, opened 2021-04-16
Fix/make list in empty document

nvdk:
this PR fixes inserting a list from a text node that's a direct child from the rootNode. For example when immediately creating a list.

I'm unsure whether the parent.parent logic in `getBlocksFromRange` is still required, but I left it as is since it does seem to work as expected. 

One annoyance I haven't addressed is that when pressing the list button at the end of a textNode it's not always consistent (sometimes the text is included in the li, but not always).

### DIFF
--- a/addon/commands/make-list-command.ts
+++ b/addon/commands/make-list-command.ts
@@ -64,7 +64,13 @@ export default class MakeListCommand extends Command {
         range.start.parentOffset = cur.getOffset();
       }
       if (range.start.parentOffset === 0) {
-        range.start = ModelPosition.fromInElement(range.start.parent.parent!, range.start.parent.getOffset());
+        if (range.start.parent === this.model.rootModelNode) {
+          //expanded to the start of the root node
+          range.start = ModelPosition.fromInElement(this.model.rootModelNode, 0);
+        }
+        else {
+          range.start = ModelPosition.fromInElement(range.start.parent.parent!, range.start.parent.getOffset());
+        }
       }
     }
     cur = range.end.nodeBefore();
@@ -77,7 +83,13 @@ export default class MakeListCommand extends Command {
         range.end.parentOffset = cur.getOffset() + cur.offsetSize;
       }
       if (range.end.parentOffset === range.end.parent.getMaxOffset()) {
-        range.end = ModelPosition.fromInElement(range.end.parent.parent!, range.end.parent.getOffset() + range.end.parent.offsetSize);
+        if (range.end.parent === this.model.rootModelNode) {
+          // expanded to the end of root
+          range.end = ModelPosition.fromInElement(this.model.rootModelNode, this.model.rootModelNode.getMaxOffset());
+        }
+        else {
+          range.end = ModelPosition.fromInElement(range.end.parent.parent!, range.end.parent.getOffset() + range.end.parent.offsetSize);
+        }
       }
     }
 
--- a/tests/unit/commands/make-list-command-test.ts
+++ b/tests/unit/commands/make-list-command-test.ts
@@ -1,12 +1,9 @@
 import {module, test} from "qunit";
 import ModelTestContext from "dummy/tests/utilities/model-test-context";
 import ModelElement from "@lblod/ember-rdfa-editor/model/model-element";
-import ModelText from "@lblod/ember-rdfa-editor/model/model-text";
 import MakeListCommand from "@lblod/ember-rdfa-editor/commands/make-list-command";
-import ModelPosition from "@lblod/ember-rdfa-editor/model/model-position";
 import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
 import {vdom} from "@lblod/ember-rdfa-editor/model/util/xml-utils";
-import ModelNode from "@lblod/ember-rdfa-editor/model/model-node";
 
 module("Unit | commands | make-list-command", hooks => {
   const ctx = new ModelTestContext();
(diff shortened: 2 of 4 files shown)


abeforgit (approved):
Looks good

I took the liberty of just pushing the changes to the test, we can revert if you disagree but otherwise it takes quite a bit of overhead for something not super critical

It did give me some ideas on how to improve the xml testing utils cause I agree it's a bit wonky now with the dummy elements

It would be cleaner to include the root I think, and have a special <modelRoot> element which the reader will expand into the div with all the correct attributes so we dont have to specify it each time

abeforgit:
I'm also pretty sure the current behavior is not what we want so with these cases im always struggling a bit: should we let the test pass to document the current behavior, or should we  let it fail to signal that it needs work?

nvdk:
> I'm also pretty sure the current behavior is not what we want so with these cases im always struggling a bit: should we let the test pass to document the current behavior, or should we let it fail to signal that it needs work?

I struggled with this as well, it's how I ended up with the somewhat vaguer test with `find`. Ideally I'd like something like `warn`, so the test can fail visually (and indicate the behaviour is not what we want) but we also know it's not blocking.

nvdk:
general remark: Something our xml comparisons do (and previous integrations tests as well) is focus quite a bit on the specific implementation of the command. e.g. if at some point we decide an empty list item starts out with a `br` child we'll need to revisit all make-list tests.

abeforgit:
Yes but I think that's a good thing
I think no matter what we do editing operations will always be brittle and depend heavily on us knowing exactly what they do and how they interact.
Breaking tests will then signal us that a previous assumption was broken and force us to think if that will be a problem or not.

abeforgit:
> > I'm also pretty sure the current behavior is not what we want so with these cases im always struggling a bit: should we let the test pass to document the current behavior, or should we let it fail to signal that it needs work?
> 
> I struggled with this as well, it's how I ended up with the somewhat vaguer test with `find`. Ideally I'd like something like `warn`, so the test can fail visually (and indicate the behaviour is not what we want) but we also know it's not blocking.

I guess the closest that we have available is letting it fail but skipping the test

This will get marked visually as well (see the next-text-node tests for an example)

Come to think of it thats not a bad workflow:
- failing test signifies an urgent bug we should fix asap
- failing but skipped test signifies potentially problematic behavior that we can live with for the moment

abeforgit:
There is also test.todo()
which passes when the test fails and vice versa (signalling that you should remove the todo when the behavior is implemented)
it also visually marks the failing test

it could serve as a third level:
- behavior we eventually want but which is not an immediate priority

abeforgit:
thing is in this case im not exactly sure what we want
I guess for the newline to be replaced by a list with a single li (and possibly an empty textnode inside the li?)

nvdk:
> thing is in this case im not exactly sure what we want
> I guess for the newline to be replaced by a list with a single li (and possibly an empty textnode inside the li?)

What I expected to happen (as in consistent with behaviour when there is text and you click the button): 
- that the contents of the textnode is moved into the li, possibly normalized.

nvdk:
> This will get marked visually as well (see the next-text-node tests for an example)

I did notice that triggers a linting error, though I did not look at it yet.

abeforgit:
> > This will get marked visually as well (see the next-text-node tests for an example)
> 
> I did notice that triggers a linting error, though I did not look at it yet.

yes, true. There are also other areas where eslint is confused, e.g. when you use a function type as a type for a parameter (complains about unused params) so we need to have a look at the config there

abeforgit:
> > thing is in this case im not exactly sure what we want
> > I guess for the newline to be replaced by a list with a single li (and possibly an empty textnode inside the li?)
> 
> What I expected to happen (as in consistent with behaviour when there is text and you click the button):
> 
> * that the contents of the textnode is moved into the li, possibly normalized.

Situations like this really prove the value of feature passports, love the direction we're going with those :+1:

abeforgit:
feature passport -> encode those into unit tests marked "todo" -> start implementing

see now we're cooking with TDD :smiley:

abeforgit:
> > > This will get marked visually as well (see the next-text-node tests for an example)
> > 
> > 
> > I did notice that triggers a linting error, though I did not look at it yet.
> 
> yes, true. There are also other areas where eslint is confused, e.g. when you use a function type as a type for a parameter (complains about unused params) so we need to have a look at the config there

made a jira story: https://binnenland.atlassian.net/browse/GN-2570

nvdk:
> > > I'm also pretty sure the current behavior is not what we want so with these cases im always struggling a bit: should we let the test pass to document the current behavior, or should we let it fail to signal that it needs work?
> > 
> > 
> > I struggled with this as well, it's how I ended up with the somewhat vaguer test with `find`. Ideally I'd like something like `warn`, so the test can fail visually (and indicate the behaviour is not what we want) but we also know it's not blocking.
> 
> I guess the closest that we have available is letting it fail but skipping the test
> 
> This will get marked visually as well (see the next-text-node tests for an example)
> 
> Come to think of it thats not a bad workflow:
> 
>     * failing test signifies an urgent bug we should fix asap
> 
>     * failing but skipped test 

I think I'd prefer `todo` for 
> signifies potentially problematic behaviour that we can live with for the moment
and use `skip` for 
>    behaviour we eventually want but which is not an immediate priority

As todo still runs your test, but skip doesn't this feels a closer match to me

nvdk:
@abeforgit what do you think, do we merge as is?

PR #86, opened 2021-04-19
refine say modal and card styles

Dietr:
Update of the say-modal, say-card and say-link-list to adapt to the new citaten-plugin.


nvdk (approved):
looks good to me!

PR #87, opened 2021-04-19
Bugfix/br in br

abeforgit:
Fixes a bug where a br could appear inside a br

Was a bit of a fundamental issue with the way we convert dom positions.
Unclear how the dom selection ended up inside the br in the first place, but this can 
be just browser behavior (a position inside a br is perfectly valid in the dom spec) so we need to be able to handle this case anyway.

Update:
it is now clear how the dom selection ended up there: the selectionWriter was just plain wrong. Kind of amazing that we didn't notice before
Could possibly have been the cause for many other headaches weve head.

The new algorithm seems a lot more sound. It hinges on the fact that a position in front of an element is visually identical to a position _inside_ that element at (dom) offfset 0

### DIFF
--- a/addon/commands/insert-newLine-command.ts
+++ b/addon/commands/insert-newLine-command.ts
@@ -25,9 +25,6 @@ export default class InsertNewLineCommand extends Command {
     const br = new ModelElement("br");
     this.model.change(mutator => {
       mutator.insertNodes(range, br);
-    }, false);
-    const cursorPos = ModelPosition.fromAfterNode(br);
-    this.model.selection.selectRange(new ModelRange(cursorPos, cursorPos));
-    this.model.writeSelection();
+    });
   }
 }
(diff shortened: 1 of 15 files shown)


nvdk:
tbh it's kind of funny to see you walk the same path wrt to selections as I did previously :).

> It hinges on the fact that a position in front of an element is visually identical to a position inside that element at (dom) offfset 0
this is not necessarily true, though it might work. a position in between block elements will look very different from offset 0 at the second element.

```
<div>some text</div>
some text |
<div>other text</div>
```
is very different from
```
<div>some text</div>
some text 
<div>|other text</div>
```

abeforgit:
yeah fair enough

since its in the writer I dont think itll be a problem
as long as the conversion is predictable and logical and any desired visual position can be described by a modelPosition it doesnt really matter how we convert

abeforgit:
the only thing I could see is getting rid of  the nodeBefore is textNode branch

that will make conversion simpler to understand (since it always looks ahead, never behind)

nvdk:
on regression i notice is when inserting the table, the cursor is no longer correctly placed in the first cell

nvdk:
> one regression i notice is when inserting the table, the cursor is no longer correctly placed in the first cell

seems this is also broken on the dev branch

PR #88, opened 2021-04-20
proper typescript linting

nvdk:
probably needs some further tweaking, putting this here so it doesn't get forgotten.

Current config based somewhat on https://github.com/typescript-eslint/typescript-eslint/issues/109


.eslintrc.js
```diff
@@ -18,7 +18,6 @@ module.exports = {
   extends: [
     'eslint:recommended',
     'plugin:ember/recommended',
-    'plugin:@typescript-eslint/eslint-recommended',
```
nvdk, line 21: I don't think this actually existed, at least it didn't seem to enable any rules

abeforgit:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-this-alias.md

This one is a bit annoying
I agree it's bad practice in their examples, but assigning this is quite natural for tree-walking algorithms
case in point: ModelNode>getOffsetPath()

Can be worked around of course so idk

Looks good otherwise, no false positives that I can spot

abeforgit:
wait it as working fine when switching branches with ember live-reloading, but a fresh build crashes.
Ember seems to be of the opinion that extends inside overrides is not allowed

.eslintrc.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
...
+    'plugin:@typescript-eslint/eslint-recommended'
```
nvdk, line 21: don't know where you get `    'plugin:@typescript-eslint/eslint-recommended'` from, I couldn't find any references to it. only to `plugin:@typescript-eslint/recommended`

nvdk:
> wait it as working fine when switching branches with ember live-reloading, but a fresh build crashes.
> Ember seems to be of the opinion that extends inside overrides is not allowed

> wait it as working fine when switching branches with ember live-reloading, but a fresh build crashes.
> Ember seems to be of the opinion that extends inside overrides is not allowed

odd in the linked ticket, it's mentioned 
> Also of note: ESLint also now supports nested overrides and extends in overrides: eslint/eslint#11554 :tada:

perhaps it's just a matter of needing a more recent eslint

nvdk:
it's green :tada:

abeforgit:
forbidden non-null-assertions are the only warnings left :)

PR #89, opened 2021-04-20
Refactor/operations

abeforgit:
This is an effort to make the model.change flow more intuitive
The existing logic hasn't changed, it's just been renamed to make their batched behavior more clear, and a non-batched version has been added 

For the moment, we have no real use for the batched version so it's mostly getting in the way.
depends on #87


abeforgit:
good thing its a draft, immediate obvious mistake spotted

addon/model/model.ts
```diff
@@ -144,6 +144,11 @@ export default class Model {
     modelNode.remove();
   }
 
+  /**
```
nvdk, line 147: huuraah for docs!

addon/commands/insert-xml-command.ts
```diff
@@ -22,7 +22,8 @@ export default class InsertXmlCommand extends Command {
     const range = selection.lastRange;
     const parsed = parseXml(xml);
     this.model.change(mutator => {
-      mutator.insertNodes(range, parsed.root);
+      const newRange = mutator.insertNodes(range, parsed.root);
+      mutator.selectRange(newRange);
```
nvdk, line 26: for my understanding, it's now expected that a command explicitly selects something or do we still (also) provide a "best guess" selection?

nvdk:
hmm accidentally pushed to this branch, apologies

addon/commands/insert-xml-command.ts
abeforgit: yes, thats what insertNodes returns
but the command has to decide to use it indeed, it is not automatically applied

nvdk (approved):
looks good to me!

nvdk:
> looks good to me!

woops looks like this introduces a linting error :)

PR #90, opened 2021-04-23
Bugfix/new li bugfix

Asergey91:


addon/model/model-position.ts
```diff
@@ -302,6 +302,19 @@ export default class ModelPosition {
     return this.parent.childAtOffset(this.parentOffset - 1) || null;
   }
 
+  //this returns true if the position is inside a text node (not right before not right after)
+  isInsideText(): Boolean{
```
nvdk, line 306:
```suggestion
  isInsideText(): boolean{
```
typescript prefers a type here

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
-    } else {
-      selectedIterator = new ModelTreeWalker({
-        range: maximizedRange,
-        filter: node => ModelNode.isModelText(node) ? FilterResult.FILTER_ACCEPT : FilterResult.FILTER_SKIP
+    //collapsed selection case
+    if (isCollapsed) {
+      this.insertNewLi(startPosition);
+    }
+    //single li expanded selection case
+    else if (startParentLi == endParentLi) {
```
nvdk, line 45: what if both startParentLi and endParentLi are `undefined`

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
-    let selectedIterator;
-    if (range.collapsed) {
-      selectedIterator = [];
-      const nodeAfter = range.start.nodeAfter();
-      if (ModelNode.isModelText(nodeAfter)) {
-        selectedIterator.push(nodeAfter);
-      } else if (ModelNode.isModelText(range.start.nodeBefore())){
-        selectedIterator.push(range.start.nodeBefore());
-      }
+    const startParentLi = startPosition.findAncestors(node => ModelNode.isModelElement(node) && node.type === "li")[0];
```
nvdk, line 37: I guess if there's no parentLi this is undefined

addon/commands/insert-newLi-command.ts
nvdk: perhaps the more general question: should this command block being executed when not already in an li?

abeforgit:
First thing I notice: hitting enter at the end of the last non-empty li has several issues

If there is nothing behind it it sometimes doesn't make a new li
if there are empty lis behind it it often makes one new li and then subsequent enters dont do anything
lots of "couldn't find a node to work with errrors"

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
+        if (!node) {
+          throw new Error("couldn't find a node to work with");
+        }
+
+        //find parent li and ul/ol as well as parent li position
+        parentLi = node.findAncestor(node => ModelNode.isModelElement(node) && (node.type === 'li'), false);
+        if (!parentLi) {
+          throw new Error("couldn't find the parent li");
+        }
+        parentUl = parentLi.findAncestor(node => ModelNode.isModelElement(node) && (node.type === 'ul' || node.type === 'ol'), false) as ModelElement;
```
abeforgit, line 271: parentUl has to be the direct parent of li, otherwise we have an invalid model state

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
+
+      //assume that the cursor is at the end
+      if (!node) {
+        node = position.nodeBefore();
+        if (!node) {
+          throw new Error("couldn't find a node to work with");
+        }
+
+        //find parent li and ul/ol as well as parent li position
+        parentLi = node.findAncestor(node => ModelNode.isModelElement(node) && (node.type === 'li'), false);
```
abeforgit, line 267: I think I'll end up undeprecating findAncestor for the time being
it seems too popular

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
+        }
+
+        //add empty li to the parent ul
+        parentUl.addChild(newLi, parentLiPos + 1);
       }
     }
-  }
+    else {
+      //find parent li and ul/ol as well as parent li position
+      parentLi = node.findAncestor(node => ModelNode.isModelElement(node) && (node.type === 'li'), false);
```
abeforgit, line 286: this is duplicated code that can easily be avoided by shuffling the if branches around a bit

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
+  insertNewLi(position: ModelPosition): void {
+
+    //split the nodes at position
+    position.split();
+
+    //variable initialization
+    let parentLi;
+    let parentUl;
+    let parentLiPos;
+    let toBeInserted: ModelNode[] = [];
```
abeforgit, line 232: toBeInserted and newLiNodes can be const and defined closer to where they are used since they are only used in the upcoming else branch

abeforgit (changes requested):
I think we're almost there, mostly small changes
General remarks: please  use triple equal signs everywhere
(I'm kinda surprised we don't have a linting rule for this?)

addon/commands/insert-newLi-command.ts
abeforgit: it won't be, so they "should" never both be undefined

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
   }
 
   canExecute(selection: ModelSelection = this.model.selection): boolean {
-    if (!ModelSelection.isWellBehaved(selection)) {
-      throw new MisbehavedSelectionError();
-    }
-    const commonAncestor = selection.getCommonAncestor().parent;
-    if (commonAncestor?.findAncestor(node => ModelNode.isModelElement(node) && (node.type === 'ul' || node.type === 'ol'), true)) {
+    if (selection.isInside(["ul", "ol"]) === "enabled") {
```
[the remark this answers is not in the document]
abeforgit: ah the enum now also has string values, not sure when that happened
which is why this works I assume
Still I'd prefer to see a comparison to the enum value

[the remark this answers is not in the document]
abeforgit: > ah the enum now also has string values, not sure when that happened

seems like it happened when I defined it to be like that :joy:

addon/commands/insert-newLi-command.ts
```diff
@@ -17,261 +17,334 @@ export default class InsertNewLiCommand extends Command {
...
+    let parentLiPos;
+    let toBeInserted: ModelNode[] = [];
+    let node = position.nodeAfter();
+
+    const newLiNodes = [];
+    const newLi = new ModelElement('li');
+
+    //try to find the first node we can copy if there is no nodeAfter
+    //looks for a right sibling of a direct ancestor untill we reach the parent li
+    if (!node) {
```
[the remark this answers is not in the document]
Asergey91: yeah Kinda, I was kinda running out of mental bandwith at the end

addon/commands/insert-newLi-command.ts
nvdk: @Asergey91 can you add a comment so this assumption is clear when reading the code?

addon/commands/insert-newLi-command.ts
nvdk: I'm fine with that if we make the assumption clear in a comment.

addon/commands/insert-newLi-command.ts
Asergey91: oki

nvdk:
solid work, I could not get this to break in some basic testing. and all the tests are green :dancers:

nvdk:
@Asergey91 / @abeforgit can you have a look at the merge conflict?

PR #91, opened 2021-04-23
Bugfix/table tab handler

lagartoverde:


addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -24,11 +26,58 @@ export default class TableTabInputPlugin implements TabInputPlugin {
...
+        }
+      }
     }
     return null;
   }
 
-  tabHandler = (manipulation : Manipulation, editor: RawEditor) => {
+  isNextElementATable(selection: ModelSelection) {
+    const nextSibling = selection.anchor?.nodeAfter() as ModelElement;
+    if(!nextSibling) return false;
```
nvdk, line 51: I think this is more readable if converted to
```
if (nextSibling) {
    return nextSibling.type === 'table';
}
else {
  return false;
}
```

addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -24,11 +26,58 @@ export default class TableTabInputPlugin implements TabInputPlugin {
         allow: true,
         executor: this.tabHandler as unknown as ManipulationExecutor
       };
+    } else {
+      if(_manipulation.type === 'moveCursorToStartOfElement' || _manipulation.type === 'moveCursorAfterElement') {
+        if(this.isNextElementATable(selection)){
+          return {
+            allow: true,
+            executor: this.selectFirstCell,
+          }
```
nvdk, line 35: missing semi

nvdk (changes requested):
Some linting errors to fix, as well as an issue I ran into in firefox:
starting out after the table, pressing shift+tab twice (so first moving into the cell and then moving to the previous cell) results in an error ` Selection is not inside a cell`

Other than that seems to work well

addon/utils/plugins/table/tab-input-plugin.ts
abeforgit: what about
`return nextSibling?.type === "table";`

abeforgit:
I will make some changes to the collapseOn method
Ill probably rename it to collapseIn and delegate to ModelRange.fromInNode so it will work consistently
Ill link the pr here when ready (will be ready today)

addon/utils/plugins/table/tab-input-plugin.ts
nvdk: I'd say the if/else is more clear if quickly eyeing the code, but I'm OK with both.

addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -7,7 +7,9 @@ import {
 import RawEditor from 'dummy/utils/ce/raw-editor';
 import ModelTable from '@lblod/ember-rdfa-editor/model/model-table';
 import { PropertyState } from '@lblod/ember-rdfa-editor/model/util/types';
-import ModelElement from "@lblod/ember-rdfa-editor/model/model-element";
```
nvdk, line 10: you've removed this type, but it's still used in the file.

abeforgit:
> I will fix the collapseOn method
> Ill probably rename it to collapseIn and delegate to ModelRange.fromInNode so it will work consistently
> Ill link the pr here when ready (will be ready today)

Here we go: #92

lagartoverde:
> Some linting errors to fix, as well as an issue I ran into in firefox:
> starting out after the table, pressing shift+tab twice (so first moving into the cell and then moving to the previous cell) results in an error ` Selection is not inside a cell`
> 
> Other than that seems to work well

I don't get this error nvdk, are you doing something special? I just create a table with some text after, then press shift tab twice when the cursor just after the table

lagartoverde:
I managed to reproduce it, it's only if you are not at the end of the table but you have some text in between, I'm checking what's happening

lagartoverde:
https://drive.google.com/file/d/1LSLmwKvqUZjlJUyYZEgV4KOqzF-MPUAg/view?usp=sharing

addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -32,55 +33,62 @@ export default class TableTabInputPlugin implements TabInputPlugin {
...
-          }
+          };
         }
       }
     }
     return null;
   }
 
-  isNextElementATable(selection: ModelSelection) {
+  isNextElementATable = (selection: ModelSelection) => {
```
nvdk, line 50: I'm personally not a fan of defining these functions this way. We could move them outside of the class, but defining them as a class field and then binding to this feels odd to me.

nvdk:
@lagartoverde I took the time to step through what's happening and I think this is a bug in the tab plugin. It occurs because we currently ignore information that you get from the manipulation. The plugin receives a manipulation of type `moveCursorToEndOfEelement` with the element `table`. So the intent is clearly communicated to the plugin. However our plugin currently ignores this information and works only on the editor.selection, which is why it makes the wrong decision. I think for now, it makes most sense to also check the element received in the plugins callback and translate that element to a dom element.

addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -24,10 +27,67 @@ export default class TableTabInputPlugin implements TabInputPlugin {
         allow: true,
         executor: this.tabHandler as unknown as ManipulationExecutor
       };
+    } else {
+      if(_manipulation.type === 'moveCursorToStartOfElement' || _manipulation.type === 'moveCursorAfterElement') {
```
nvdk, line 31: if we're using the variable, don't start its name with an underscore

addon/utils/plugins/table/tab-input-plugin.ts
```diff
@@ -24,10 +27,67 @@ export default class TableTabInputPlugin implements TabInputPlugin {
...
+        if(this.isNextElementATable(selection)){
+          return {
+            allow: true,
+            executor: this.selectFirstCell,
+          };
+        }
+      } else if(_manipulation.type === 'moveCursorToEndOfElement' || _manipulation.type === 'moveCursorBeforeElement') {
+        if(this.isPreviousElementATable(selection)){
```
nvdk, line 39: this should use the manipulation.element instead of just relying on the selection

nvdk:
> @lagartoverde I took the time to step through what's happening and I think this is a bug in the tab plugin. It occurs because we currently ignore information that you get from the manipulation. The plugin receives a manipulation of type `moveCursorToEndOfEelement` with the element `table`. So the intent is clearly communicated to the plugin. However our plugin currently ignores this information and works only on the editor.selection, which is why it makes the wrong decision. I think for now, it makes most sense to also check the element received in the plugins callback and translate that element to a model element if you need that.

I've merged arne's branch and have made some changes to take the provided element into account, think it works now. if someone else can confirm we can merge!

lagartoverde:
But why would you want to be at the end of the table if you don't want to be inside the table?

addon/utils/plugins/table/tab-input-plugin.ts
lagartoverde: Not a fan either, I did this in order to solve the typescript error generated 
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md

PR #92, opened 2021-04-23
Bugfix/collapse selection

abeforgit:
Long overdue

Clears out some deprecated and unused selection/range/position methods and fixes the ones that were broken.
No more "ah yeah but you shouldn't use this method anymore" from me :see_no_evil: 

Most notably renames the selection.collapseOn method to collapseIn, it makes it more obvious what it does and reflects the names of the methods it delegates to.


nvdk (approved):
Looks good to me, didn't see anything unexpected and the editor still seems to work :dancers:

PR #94, opened 2021-04-28
replace global-text-offset-to-path with global-text-region-to-model-range

nvdk:
In all cases the function was used we're actually mapping a range, so it makes more sense to pass and convert the entire region instead of just a position. To properly map a region to a range and prefer the correct edge, it's also easier if known whether we're searching for the start or end. 

this PR also adds tests for the utility


addon/utils/global-text-region-to-model-range.ts
```diff
@@ -0,0 +1,84 @@
...
+  else if (eligibleNodes.length > 1) {
+    // by definition this should include edges, prefer end
+    const eligibleNode = eligibleNodes.find((textWithOffsets) => textWithOffsets[preferredEdge] === offset);
+    if (! eligibleNode) {
+      throw `no eligble node found`;
+    }
+    return eligibleNode;
+  }
+  else {
+    throw `no eligible node found`;
```
abeforgit, line 63: I'd prefer to see error instances here
If you cant find a suitable error type and dont want to make a new one, you can throw `new CustomError("no eligible node found")`

tests/unit/utils/global-text-region-to-model-range-test.ts
```diff
@@ -0,0 +1,84 @@
...
+import ModelElement from "@lblod/ember-rdfa-editor/model/model-element";
+import globalTextRegionToModelRange from "@lblod/ember-rdfa-editor/utils/global-text-region-to-model-range";
+
+module("Unit | Utility | global-offset-to-path", function () {
+  test("returns the correct range for a simple text", function (assert) {
+    // language=XML
+    const {root, textNodes: {testNode}} = vdom`<div><text __id="testNode">abc</text></div>`;
+    const calculatedRange = globalTextRegionToModelRange(root as ModelElement, 0, 3);
+    const range = ModelRange.fromInTextNode(testNode,0, 3);
+    assert.deepEqual(range, calculatedRange);
```
abeforgit, line 14: For ranges there is also `sameAs` 
I dont think it matters really which one we use but just so you know

abeforgit (changes requested):

addon/utils/global-text-region-to-model-range.ts
nvdk: I've created an IllegalArgumentError

Asergey91 (approved):
makes sense

abeforgit (approved):

PR #95, opened 2021-04-28
Refactor/operations

abeforgit:
This pr fleshes out the operations api 
main improvements:
- remove the unnecessary dependency on the model for the mutators to make testing easier
- introduced split operation
- added useful convenience methods to the mutator
- make the MoveOperation and it's associated mutator methods usable, and test it thoroughly
- add a fully operation-based unwrap method to the mutator
- because of lots of added tests, some bugs in the core algorithms were found and fixed
- adds a toString method on ModelRange for easier debugging
-  reimplement the remove-list command with the new methods

The last point was a way to discover what kind of methods we need on the mutator to make commands feasible to implement. I'm quite happy where we landed. Lots of effort went into making the mutator methods easy to use with no caveats.
This means most of the required position shuffling happens inside the mutator, and we can expect a logical and valid range to be returned after the mutator has done it's thing

This is the way I'd like to see the mutators evolve. Rather than expecting a command implementor to deal with the intricacies of positions that become invalid because of model mutations, provide methods that deal with this automatically and give you both a valid model state and a valid range to continue working with.

It also marks a bit of a shift in the ideas behind the mutator api: initially, everything was supposed to work on ranges. While this is still a good baseline, some useful methods (like unwrap for example) just don't make sense to be defined on a range. Other methods work better on positions. What does remain very useful is the convention of returning a valid and sensible default range /position as a result of the method


nvdk:
random comment: having unwrap makes me dream of a wrap operation, which can be very useful for some commands I'd like to create for plugins :)

abeforgit:
> random comment: having unwrap makes me dream of a wrap operation, which can be very useful for some commands I'd like to create for plugins :)

It wont be an operation, but it will for sure get added to the mutator

abeforgit:
and saying that, feel free to add it yourself :stuck_out_tongue:  it shouldnt be that hard, the unwrap was also fairly simple

addon/model/model-range.ts
```diff
@@ -30,6 +30,11 @@ export default class ModelRange {
     const end = ModelPosition.fromInElement(element, endOffset);
     return new ModelRange(start, end);
   }
+  static fromAroundNode(node: ModelNode) {
```
nvdk, line 33: nitpicking: could benefit from a newline above the method

addon/model/mutators/batched-model-mutator.ts
```diff
@@ -25,10 +26,9 @@ export default class BatchedModelMutator extends ModelMutator<void> {
     this.batch.push(op);
   }
 
-  move(rangeToMove: ModelRange, targetRange: ModelRange) {
-    const op = new MoveOperation(rangeToMove, targetRange);
+  moveToPos(rangeToMove: ModelRange, targetPos: ModelPosition) {
```
nvdk, line 29: in all other instances position is written out in full, so would be more consistent to also do that here?

nvdk (approved):
certainly looks ok, will need to actually use the mutators to get a feel for whether they fit or not.

addon/model/model-range.ts
abeforgit: I love nitpicking, good catch :)

addon/model/mutators/batched-model-mutator.ts
abeforgit: I agree!

PR #96, opened 2021-04-28
Bugfix/insert html

abeforgit:
Add some testcases for insert html in order to try and reproduce wrong behavior when inserting into an empty document
as mentioned here https://github.com/lblod/ember-rdfa-editor-standard-template-plugin/pull/3

Failed to reproduce the problem, but the tests will be useful in any case

depends on #95

### DIFF
--- a/tests/unit/commands/insert-html-command-test.ts
+++ b/tests/unit/commands/insert-html-command-test.ts
@@ -1,0 +1,166 @@
+import {module, test} from "qunit";
+import {vdom} from "@lblod/ember-rdfa-editor/model/util/xml-utils";
+import ModelTestContext from "dummy/tests/utilities/model-test-context";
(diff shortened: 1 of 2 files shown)


nvdk:
fyi when depending on another PR, I'd recommend setting the base to that PR's branch, it makes for a cleaner diff. Once the base PR is merged the base of this PR will automatically be converted to the branch it's been merged to.

nvdk (approved):

abeforgit:
> fyi when depending on another PR, I'd recommend setting the base to that PR's branch, it makes for a cleaner diff. Once the base PR is merged the base of this PR will automatically be converted to the branch it's been merged to.

ah good to know! I knew there had to be a better way to do this super common flow

PR #98, opened 2021-04-29
Bugfix/insert list

abeforgit:
- reworks and fixes bugs in inLIstState and inTableState methods
- renames and reworks isInside selection method to hasCommonAncestorWhere and moves it to modelrange
- renames and reworks contains selection method to containsNodeWhere and moves it to modelrange
- adds tests for and fixes broken algorithm of getCommonAncestor method
- remove some unused and deprecated selection methods
- fixes selection issues in make-list command

the latter is a hotfix rather than the necessary rework, though I don't think the rework would have solved the issue anyway. The problem is that while the operation and mutator methods do return useful ranges to select, when you run a cleaner after them that sort of ruins the whole affair. 

The current fix leaves you with a predictable albeit maybe slightly unexpected selection (when making a list of the line right after an existing list, you end up with the whole merged list selected)
This also fixes the reported issue in an empty document.

### DIFF
--- a/addon/model/model-range.ts
+++ b/addon/model/model-range.ts
@@ -111,7 +111,6 @@ export default class ModelRange {
   * findCommonAncestorsWhere(predicate: Predicate<ModelElement>): Generator<ModelElement, void, void> {
     let commonAncestor: ModelElement | null = this.getCommonAncestor();
     while (commonAncestor) {
-      console.log(commonAncestor.toXml());
       if (predicate(commonAncestor)) {
         yield commonAncestor;
       }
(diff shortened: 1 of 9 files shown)


tests/unit/model-selection-test.ts
```diff
@@ -76,87 +73,5 @@ module("Unit | model | model-selection", hooks => {
     assert.strictEqual(result[3], li012);
 
   });
-  test("collapsed selection inside an element", assert => {
```
abeforgit, line 79: note these are moved to model-range-test, not deleted

addon/model/model-range.ts
```diff
@@ -101,6 +103,41 @@ export default class ModelRange {
...
 
+  hasCommonAncestorWhere(predicate: Predicate<ModelElement>): boolean {
+    const result = this.findCommonAncestorsWhere(predicate).next();
+    return !!result.value;
+  }
+
+  * findCommonAncestorsWhere(predicate: Predicate<ModelElement>): Generator<ModelElement, void, void> {
+    let commonAncestor: ModelElement | null = this.getCommonAncestor();
+    while (commonAncestor) {
+      console.log(commonAncestor.toXml());
```
nvdk, line 114:
```suggestion
```
I guess this is a stray debugging statement, it logs quite often

nvdk (approved):
lists in general work a lot better with these changes! aside from the stray console.log statement I don't see anything off, nice work :+1:

PR #100, opened 2021-05-05
replace ember-cli-release with release it

nvdk:
as a side effect, this removes plain css support for now


package.json
```diff
@@ -118,11 +116,18 @@
...
     "last 1 Firefox version",
-    "last 1 FirefoxAndroid version",
-    "last 1 Safari version",
-    "last 1 iOS version",
     "last 1 Edge version"
-  ]
+  ],
+  "release-it": {
+    "github": {
+      "release": false
```
Windvis, line 124: I think `false` is the default so this can be removed if this isn't a wanted feature.

index.js
```diff
@@ -9,16 +9,6 @@ module.exports = {
   },
 
   included() {
```
Windvis, line 11: The whole method can be removed I think?

package.json
```diff
@@ -118,11 +116,18 @@
...
+    },
+    "npm": {
+      "publish": false
+    },
+    "git": {
+      "requireCleanWorkingDir": true
```
Windvis, line 130: I think this is also the default behavior? https://github.com/release-it/release-it/blob/master/docs/git.md#clean-working-directory

Windvis (approved):
Only some small / nitpicky things, so feel free to ignore that :smile:.

index.js
nvdk: I've kept it, because last time something was added to included we forgot to call super. but you're right it can go.

PR #101, opened 2021-05-05
Rework handler types

abeforgit:
This is more of a discussion point cause the code on this commit doesn't compile

As you can see I shuffled some things around in the type definitions for the manipulations
Basically the previous definitions  were really annoying to extend, for the following reason:

while the BaseManipulation supertype did not require a node to be present (it had a question mark, meaning it was optional)
all the existing manipulations did define one. As such, the code that was using them was depending on that node always being there in many places, even though this was just a coincidence as far as the types were concerned.

This meant that when I wanted to add a manipulation which did not have a node, this assumption broke and type-errors started appearing left and right. Fundamentally the problem was the definition of Manipulation, which was just a tagged union of a bunch of manipulation types with no constraints as to their contents. Typescript is pretty smart about this and can determine whether every member of the tagged union implements a certain property, which is why calling code was allowed to depend on the node being there.

I fixed this by renaming the BaseManipulation type to Manipulation, a simple superclass to sort of enforce that every manipulation should have a type property (since we're working with tagged unions, theres basically no way to _really_ enforce that, but lets not get into that). Then I got rid of the tagged union definition of manipulation, and replaced it with a handler-specific tagged union of manipulations that that specific handler generates and passes to its plugins. This means that plugins can now benefit from typed information about which manipulations can possibly be received, so they can safely ignore the ones that will not be sent. On the handler side, you now have a safeguard when you introduce a new possible manipulation. Plugins that would crash because of the addition will now generate compile time errors, basically.

Its a change that touches a lot of code, but since its really only in typescript land im quite confident (and I did test) that nothing really broke. Still I wanted to open this pr so you can see what happened, rather than just having these changes shoveled in with the upcoming text-input-handler pr

### DIFF
--- a/addon/editor/input-handlers/enter-handler.ts
+++ b/addon/editor/input-handlers/enter-handler.ts
@@ -1,5 +1,5 @@
-import { InputHandler } from './input-handler';
-import RawEditor from "@lblod/ember-rdfa-editor/utils/ce/raw-editor";
+import {InputHandler} from './input-handler';
+import LegacyRawEditor from "@lblod/ember-rdfa-editor/utils/ce/legacy-raw-editor";
 
 
 /**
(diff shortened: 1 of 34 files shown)


addon/editor/input-handlers/enter-handler.ts
```diff
@@ -1,5 +1,5 @@
-import { InputHandler } from './input-handler';
-import RawEditor from "@lblod/ember-rdfa-editor/utils/ce/raw-editor";
+import {InputHandler} from './input-handler';
+import LegacyRawEditor from "@lblod/ember-rdfa-editor/utils/ce/legacy-raw-editor";
```
nvdk, line 2: I think it's quite safe to use PernetRawEditor just about everywhere

PR #102, opened 2021-05-06
Insert text command

abeforgit:
# Changes
## New features
- implement insertText mutator method, since I suspect we'll want this for multiple commands
- implement insert text command using that method
- rewrite the text-input handler to use the new command
- rewrite the existing handler pugins to use the command 
## Code cleanup
- Replace references to legacy raw editor with pernet
- rework typing structure of Manipulations to allow for easier extension and better type information (see #101 for a better overview)
- make the InputHandler basetype an abstract class instead of an interface, since the `checkManipulationsByPlugins` method was identical on all implementations of it.

# Comments

I'm still planning on doing the TODO comments, it just might take a while and I didn't want to block this PR cause it's fine for now (we cant depend on operation-only modifications yet anyway)

### DIFF
--- a/addon/commands/insert-text-command.ts
+++ b/addon/commands/insert-text-command.ts
@@ -1,0 +1,24 @@
+import Command from "@lblod/ember-rdfa-editor/commands/command";
+import Model from "@lblod/ember-rdfa-editor/model/model";
+import ModelRange from "@lblod/ember-rdfa-editor/model/model-range";
+import {MisbehavedSelectionError} from "@lblod/ember-rdfa-editor/utils/errors";
+
+export default class InsertTextCommand extends Command {
+  name = "insert-text";
+
+  constructor(model: Model) {
+    super(model);
+  }
+
+  execute(text: string, range: ModelRange | null = this.model.selection.lastRange): void {
+    if (!range) {
+      throw new MisbehavedSelectionError();
+    }
+    this.model.change(mutator => {
+      const resultRange = mutator.insertText(range, text);
+      resultRange.collapse();
+      this.model.selectRange(resultRange);
+    });
+  }
+
+}
--- a/addon/commands/insert-text-command.ts
+++ b/addon/commands/insert-text-command.ts
@@ -16,8 +16,10 @@ export default class InsertTextCommand extends Command {
     }
     this.model.change(mutator => {
       const resultRange = mutator.insertText(range, text);
+      const commonAncestor = resultRange.getCommonAncestor();
       resultRange.collapse();
       this.model.selectRange(resultRange);
+      return commonAncestor;
     });
   }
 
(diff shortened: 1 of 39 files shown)


addon/editor/input-handlers/text-input-handler.ts
```diff
@@ -91,171 +67,43 @@ export default class TextInputHandler implements InputHandler {
...
-        console.warn("no selected range, not doing anything!");
-      }
-    }
-
-    else {
-      throw "unsupport manipulation";
+  handleNativeManipulation(manipulation: TextHandlerManipulation) {
+    console.log("handling native manip");
+    if (manipulation.type === "insertTextIntoRange") {
+      console.log(manipulation);
```
nvdk, line 91: stray console.log statements

addon/editor/input-handlers/text-input-handler.ts
```diff
@@ -91,171 +67,43 @@ export default class TextInputHandler implements InputHandler {
...
+  getNextManipulation(event: KeyboardEvent): TextHandlerManipulation {
+    const range = this.rawEditor.model.selection.lastRange;
+    if (!range) {
+      throw new MisbehavedSelectionError();
     }
-    else if (selection && this.rawEditor.rootNode.contains(selection.anchorNode)) {
-      return { type: "replaceSelectionWithText", selection: selection, node: (selection.anchorNode as Node), text: event.key };
-    }
-    throw "selection is required for text input";
+    const text = event.key === " " ? NON_BREAKING_SPACE : event.key;
```
nvdk, line 103: I assume this is an attempt at simplifying the logic wrt to spaces, but this will result in all spaces being non breaking spaces. Which means our lines will not break if you type a long sentence in the editor. Unless this is addressed in the writer or the mutator I don't think this is a good solution

nvdk (changes requested):
handling of spaces seems to be insufficient, other than that looks fine.
Screenshot_2021-05-07 Dummy

abeforgit:
> handling of spaces seems to be insufficient, other than that looks fine.
> Screenshot_2021-05-07 Dummy
> 

Ughhh I knew it was too easy :p

addon/editor/input-handlers/text-input-handler.ts
abeforgit: should be handled in https://github.com/lblod/ember-rdfa-editor/pull/102/commits/6a51855654eb2709c0c100a67254990a628db05f

addon/model/model-position.ts
```diff
@@ -306,6 +306,61 @@ export default class ModelPosition {
...
+   * the parentOffset the sum of this position's parentOffset and amount (this
+   * means negative amounts will shift towards the left), but bounded
+   * by 0 and the parent's maxOffset.
+   *
+   * You can think of it as moving the cursor by amount characters,
+   * while never going outside of this positions parent, and counting elements
+   * as 1 character.
+   * @param amount
+   */
+  shiftedBy(amount: number): ModelPosition {
```
nvdk, line 352: this is awesome for plugins as well

addon/commands/insert-text-command.ts
```diff
@@ -15,6 +16,13 @@ export default class InsertTextCommand extends Command {
       throw new MisbehavedSelectionError();
     }
     this.model.change(mutator => {
+      if (text.charAt(0) === " ") {
```
nvdk, line 19: pedantic: You created the space constant but are not using it here

nvdk:
ah I did notice it behaves a bit weirdly if you do funky stuff like create a full line of spaces, but I think that was also the case with the previous implementaiton. so not really blocking imo

abeforgit:
Probably best to handle that when it becomes a problem I'd say

PR #103, opened 2021-05-10
initial support of rdfa attributes in vdom

nvdk:


addon/model/model-element.ts
```diff
@@ -336,6 +344,53 @@ export default class ModelElement extends ModelNode implements Cloneable<ModelEl
...
+      this.updateRdfaPrefixes();
+    }
+  }
+
+  /*
+   * returns a parsed representation of the rdfa attributes
+   * prefixes are expanded and multivalue attributes are split on space and returned as an array
+   */
+  getRdfaAttributes(): RdfaAttributes {
+    return new RdfaAttributes(this, Object.fromEntries(this.getRdfaPrefixes()));
```
nvdk, line 391: open to suggestions on how to best address the linting error here

abeforgit: it's simply a version mismatch
our tsconfig is targetting es2017, while this method was only introduced in es2019
if we can afford to target that version, we can just update the tsconfig

abeforgit:
note target branch is master, just making sure its intentional

addon/model/model-element.ts
```diff
@@ -336,6 +344,53 @@ export default class ModelElement extends ModelNode implements Cloneable<ModelEl
     }
   }
 
+  getVocab(): string | undefined {
```
abeforgit, line 347: I tend to prefer null for absent values
something like
```typescript
  getVocab(): string | null {
    const result = this.getRdfaPrefixes().get("");
    return result ?? null;
  }
```

addon/model/model-element.ts
nvdk: I don't think es2019 is fully supported by our target browsers yet, but this method is.

types/lblod/marawa/rdfa-attributes.d.ts
```diff
@@ -1,10 +1,19 @@
 declare module '@lblod/marawa/rdfa-attributes' {
   export default class RdfaAttributes {
-    constructor(domNode: Node, knownPrefixes?: unknown, options? :Record<string, unknown>);
+    constructor(node: { getAttribute(key: string): string | undefined}, knownPrefixes?: unknown, options? :Record<string, unknown>);
     get vocab(): string;
     get content(): string;
-    get property(): string;
+    get properties(): string[];
```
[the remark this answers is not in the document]
abeforgit: but they do exist right?

addon/model/readers/html-reader.ts
```diff
@@ -35,7 +48,8 @@ export default class HtmlReader implements Reader<Node, ModelNode[], void> {
 
   read(from: Node): ModelNode[] {
     from.normalize();
-    const context = new HtmlReaderContext(this.model);
+    const prefixes = calculateRdfaPrefixes(from);
+    const context = new HtmlReaderContext(this.model, prefixes);
```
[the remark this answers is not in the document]
abeforgit: so what this does is parse the prefixes that may be defined above the element we are reading right?
including going outside of the editor root

I think it's fine when we read the rootnode, but im a bit skeptical about the situation where we read a node that's not the rootnode
I think this currently only occurs in the insert-html command

On first thought I think it will be fine, but it's something to watch out for in any case

abeforgit (changes requested):
Looks good overall I think, just have a look at that typescript target version thing

If after that you still have the linter complain about using an `any` value WHEN you run the `npm run lint:error` command from the console (editors can give false positives on that rule)
just cast the resulting object to `unknown` or something.

From googling I know it's possible to type that method correctly, but I don't think it's worth it for now, we don't even have proper typing on what the object is supposed to look like in the first place

addon/model/model-element.ts
abeforgit: man writing javascript really is like trying to drive a car while the wheel is still being invented

Ill see what I can do with the tsconfig

types/lblod/marawa/rdfa-attributes.d.ts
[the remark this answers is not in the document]
nvdk: they exist indeed!

addon/model/model-element.ts
abeforgit: adding this line to the tsconfig while leaving target unchanged seems to do the trick:
`    "lib": ["ES2017", "ES2019.Object", "DOM", "DOM.Iterable"],`

addon/model/readers/html-reader.ts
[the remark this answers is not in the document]
nvdk: it should be fine indeed, though perhaps we should hava a param to enable it and default to false? I think it's only necessary when reading the rootNode (it doesn't hurt in other cases afaik, but I don't see much use).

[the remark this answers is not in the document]
abeforgit: id rather just leave it and see if we run into issues
ive had a period where I used a lot of default params but it can get confusing quickly

addon/model/model-element.ts
nvdk: I've found a better table, and according to this one all es2019 are supported atm: https://kangax.github.io/compat-table/es2016plus/ . So I'll go for targetting es2019. The other solution was still giving me issues, even when casting.

PR #104, opened 2021-05-10
Make model initialization more robust

abeforgit:
Createsnapshot is called on every model write,
including the initial write which happens in the rootNode setter, which
is triggered before the model initialization.

This also gets rid of the non-null assertion marker for the model property of the editor (since it can clearly be undefined sometimes) but for convenience also adds a getter which throws when the model is undefined. This will give us clearer errors when we run into the rare case of an undefined model, without needing a million nullchecks throughout the code


abeforgit:
@Asergey91 can you confirm if this fixes your issue?

Asergey91:
linked it, the error doesn't show up anymore @abeforgit good job.

Asergey91 (approved):

PR #105, opened 2021-05-11
hide add row in header row

lagartoverde:
To be honest I'm not a fan of having a property on the selection just for this, I was thinking that maybe we could expand the isInTable property to include the cell position of the cursor, I think is something to discuss but for now this works :)

### DIFF
--- a/tests/unit/model/readers/html-reader-test.ts
+++ b/tests/unit/model/readers/html-reader-test.ts
@@ -176,7 +176,6 @@ module("Unit | model | readers | html-reader", hooks => {
 
     assert.true(actual[0].sameAs(expected));
     assert.strictEqual((((actual[0].root as ModelTable)
-      // the header row is also counted
       .getCell(0, 0) as ModelElement)
       .firstChild as ModelText)
       .content, "cell00");


nvdk (changes requested):
The conclusion of our discussion yesterday was to remove the idea of a title/heading row completely.

lagartoverde:
Sorry I didn't pick up that part, then, I will remove it :)

abeforgit (changes requested):
one of the tests is failing (fairly sure its the test that broke because previous assumptions don't hold anymore)
otherwise seems to work well
I have some thoughts about the formatting but until we settle on a redpencil style I'll refrain from being too annoying about these things :stuck_out_tongue:

abeforgit:
**Warning**: the base branch is master, is that on purpose

nvdk:
> **Warning**: the base branch is master, is that on purpose?

Yes we want to push these fixes to the current QA release so they can be included in the sandbox release next week.

nvdk:
> I'm working on disappearing cursor issues for all table commands
> unsure if you need those before you release? I think that might be broken in the current version as well

yes that one should also go to master if possible (so also based from master :))

nvdk (approved):
Looks good to me, nitpicking: there is a comment in the test about the header that can go now.

abeforgit (approved):

PR #106, opened 2021-05-12
Fix selection focus issues when using the dropdown

abeforgit:
rewires the logic of the dropdown a bit
adds a focus listener to the rootNode which writes the modelSelection (this is the key improvement)

base branch master cause this needs to go to embeddable

fixes: https://binnenland.atlassian.net/browse/GN-2625

### DIFF
--- a/addon/components/rdfa/toolbar-dropdown.js
+++ b/addon/components/rdfa/toolbar-dropdown.js
@@ -1,7 +1,7 @@
 import Component from "@glimmer/component";
-import { action } from "@ember/object";
-import { guidFor } from '@ember/object/internals';
-import { tracked } from "@glimmer/tracking";
+import {action} from "@ember/object";
+import {guidFor} from '@ember/object/internals';
+import {tracked} from "@glimmer/tracking";
 
 export default class AuDropdown extends Component {
   // Create a dropdown ID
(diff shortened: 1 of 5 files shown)


addon/components/rdfa/toolbar-dropdown.js
```diff
@@ -1,23 +1,41 @@
...
+  }
+
+  @action
+  deactivateFocusTrap() {
+    this.focusTrapActive = false;
+  }
 
   @action
   deactivateDropdown() {
-    if (this.dropdownOpen)
```
nvdk, line 15: this was here for a reason (to avoid changing it twice in a runloop), is that solved if not reinstate it please :). things like these are typically there for a reason

addon/utils/ce/raw-editor.ts
```diff
@@ -75,15 +75,20 @@ class RawEditor extends EmberObject {
...
   initialize(rootNode: HTMLElement) {
-    if(this.modelSelectionTracker) {
+    if (this.modelSelectionTracker) {
       this.modelSelectionTracker.stopTracking();
     }
     this.registeredCommands = new Map<string, Command>();
     this.model = new Model(rootNode);
     this.modelSelectionTracker = new ModelSelectionTracker(this.model);
     this.modelSelectionTracker.startTracking();
+    this.rootNode.addEventListener("focus", () => {
```
nvdk, line 85: it's yet another listener we don't remove... though I guess it's removed when the element is removed :thinking:

addon/components/rdfa/toolbar-dropdown.js
abeforgit: it actually errorred when I left it in :)
Note the way the dropdown is closed is different now: only the the deactivation of the focusTrap triggers this method
before there was an onclick on the div which deactivated the dropdown AND also deactivated the focusTrap, which again deactivated the dropdown
you can see how that can cause issues

nvdk: in that case I said nothing!

nvdk (approved):
seems to work well!

PR #107, opened 2021-05-12
Use inserthtml for pasting

abeforgit:
- move placeholder extension logic into a util class so we use the same code in the text input plugin as in the paste handler
- replace slightly dodgy custom html insertion logic with inserthtml command
- made sure selection is collapsed to end after insertion

seems to be working nicely

fixes: https://binnenland.atlassian.net/browse/GN-2564

### DIFF
--- a/addon/components/ce/content-editable.js
+++ b/addon/components/ce/content-editable.js
@@ -22,6 +22,7 @@ import { A } from '@ember/array';
 import { createElementsFromHTML } from "@lblod/ember-rdfa-editor/utils/dom-helpers";
 import { PropertyState } from "@lblod/ember-rdfa-editor/model/util/types";
 import LegacyRawEditor from "@lblod/ember-rdfa-editor/utils/ce/legacy-raw-editor";
+import ModelRangeUtils from "@lblod/ember-rdfa-editor/model/util/model-range-utils";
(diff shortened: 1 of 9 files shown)


lagartoverde (approved):
Looks sensible to me and it works when I tested it so keep going :)

addon/components/ce/content-editable.js
```diff
@@ -273,20 +275,22 @@ export default class ContentEditable extends Component {
...
     }
 
     else {
       const text = this.getClipboardContentAsText(clipboardData);
-      inserHtml(text);
+      this.rawEditor.executeCommand("insert-html", text, pasteRange);
     }
+    this.rawEditor.selection.lastRange.collapse();
+    this.rawEditor.model.writeSelection();
     this.rawEditor.updateRichNode();
```
nvdk, line 294: I think updateRichNode can be removed here, since it will already be called when executing the command

nvdk (changes requested):
not sure if related to this change, but I could not remove a pasted table with backspace. I assume this is actually a bug in the backspace handler, that did not show up before because of that empty row?

Perhaps we can merge this, but we should improve the behaviour of lumpnodes in the editor before it can be released.

PR #113, opened 2021-06-04
bump marawa

nvdk:
bumps marawa, so possible sideeffects are in our richnode structure and the rdfa-blocks for plugins. Seems to be backwards compatible, but probably needs some testing


abeforgit (approved):
Seems to work. I think we're hitting the point where not having release notes is really hurting us, very hard to tell what actually changed in marawa

PR #110, opened 2021-05-18
Ignore empty link tags when pasting

lagartoverde:

PR #114, opened 2021-06-10
Bugfix/pasting text

nvdk:
fixes the bug discussed in #110


abeforgit (commented):
Works, however when copying an <a> tag with a valid href, this now also inserts as text. Is this something we want?

nvdk:
> Works, however when copying an tag with a valid href, this now also inserts as text. Is this something we want?

if the editor is configured to paste text only, if you're pasting in a table cell, or if you copy from something that doesn't provide the mimetype text/html we should indeed paste that as text and not as html.

PR #117, opened 2021-06-16
Make it possible to click outside of a table

abeforgit:
Listening for rootnode focus was not a good way to solve the focus issues with the dropdown

On the dummy editor, this solution also gives focus back correctly when using the table dropdown, but on GN it does not.
This is not a regression however since this is already the case on the dev env


nvdk:
well done!

PR #118, opened 2021-06-22
Bump ember-cli-typescript to latest

abeforgit:
After going through the changelogs and upgrade guides I couldn't find any reason not to update. Keeping at least typescript fresh has some benefits since the landscape for backwards compatible types in libraries is very spotty, falling too far behind breaks type support for newer libraries. This is all in type-space so never really blocking but still, it should be relatively painless to keep ts updated for that same reason.

Ran tests, linter, tried it in GN, no issues.


package.json
```diff
@@ -29,7 +29,9 @@
     "postpublish": "ember ts:clean",
     "icons": "svg-symbols ./public/icons > ./public/say-symbolset.svg",
     "prepare": "svg-symbols ./public/icons > ./public/say-symbolset.svg",
-    "release": "release-it"
+    "release": "release-it",
+    "prepack": "ember ts:precompile",
```
nvdk, line 33: yeay more hooks :roll_eyes:

nvdk (approved):
I see no reason not to go for this. The dummy app still runs and the ci seems happy so :+1:

PR #119, opened 2021-06-28
Feature/logging

abeforgit:
Adds [diary](https://github.com/maraisr/diary) as a logging framework. Its api is almost identical to `console` (because it just wraps it), however unlike `console`, a string message as a first argument is required. This means you should not do

```ts
const thing = { foo: "bar" };
diary.log(thing);
```

but instead
```ts
const thing = { foo: "bar" };
diary.log("Useful context: ", thing);
```

see also https://github.com/maraisr/diary/issues/14 for discussion.

- wraps the diary methods to provide better defaults. Prefer using the exported functions from `logging-utils` over importing from `diary` directly
- adds some handy decorators for logging method and command calls
- adds localstorage based loglevel filtering
- adds a convenience `debug` export which is simply a diary with the name 'debug'. prefer using it instead of `console`.


nvdk:
on what grounds did you pick diary? I'm a bit a wary, given that it looks like a very new package, with some risk of being abandoned given it's sole contributor...

abeforgit:
based on the fact that all other libs are worse :)
logging is an absolute wasteland in js

No but mainly 
- it is tiny and very simple, and the code is clean. So in case it gets abandoned we could easily fork it.
- it comes with the "scoping" api from `debug` which allows for easy and powerful filtering, which I quite like
- it works in node and the browser ootb (you'd think that'd be standard, and boy would you be wrong in thinking so)
- it defers to the builtin loggers, so we keep the crazy useful object/xml tree printing
- it's typescript

abeforgit:
oh and the dev has been real responsive and open to collaboration so far

nvdk:
fair enough, seems sufficiently simple to replace if needs be as well. I'm quite a fan of the debug filtering btw and no extra dependencies is also nice :).

maraisr:
Always happy to help and see you guys succeed! And thank you again so much for supporting me in raising issues, and PR's. Such a courage boosting thingy that im doing something right 🥰

PR #120, opened 2021-06-29
Feature/vendor environment

abeforgit:
This has been a lot less useful than I had hoped. But I'm releasing it anyway since I do like the improvements to the dummy app, and it at least lets us test in an iframe.

- Implement new route for vendor environment
  Some rearranging of the routes was necessary due to the unique
  requirement of the vendor route to display the editor in an iframe.
- Converted js files to ts along the way where applicable
- Modernize routes in dummy app


package.json
```diff
@@ -59,6 +58,7 @@
     "ember-cli-typescript": "^4.2.1"
   },
   "devDependencies": {
+    "@ember/render-modifiers": "^1.0.2",
```
nvdk, line 61: quite sure this needs to be a dependency? we use this on the contenteditable class

abeforgit: oh I didnt realize we already had it so I just `ember install`ed it.
Bit annoying it then overwrites your package.json for no reason

nvdk:
I'd suggest adding the change detection logic they have to the vendor route, might be related to that still?

PR #122, opened 2021-06-30
Feature/custom dummy data

abeforgit:
While bughunting I got annoyed with always having to call `__executeCommand("insert-xml")` from the console to reproduce model trees.

I may have gotten slightly carried away, but now we have a really nice workflow for reproducing editor state. 

- Changes to the editor document are persisted to localstorage now, which means they survive a refresh (like from a hot-reload)
- Custom html and custom xml buttons added, which show the respective representations of the document, and allow editing them directly. Upon saving, these states will be written to the document.
- uses [codemirror 6](https://codemirror.net/6/) as an editor. It's the only one that wasn't horrible to integrate with ember.


nvdk:
feature request: auto formatting on the html/xml code :). in all seriousness looks awesome :+1:

abeforgit:
the xml code is already autoformatted tho :)
html you cant format cause whitespace has a meaning there

PR #123, opened 2021-07-01
Bugfix/space eats chars

abeforgit:
Fixes https://binnenland.atlassian.net/browse/GN-2749
and 
https://binnenland.atlassian.net/browse/GN-2798


nvdk:
seems to solve the issue! noticed one thing (may or may not be a regression): after pressing space in a placeholder it show the cursor in the wrong position. I'll create a separate ticket for it.

PR #124, opened 2021-07-01
move the dispatcher service inside the editor addon

nvdk:
it seems farfetched that the current code will evolve into an addon of its own,
it's more likely the dispatcher as it exists now will dissapear relatively soon.

this commit also make the loading of plugins a bit more defensive and
informative, if a plugin is configured in a profile but could not be loaded a
warning is printed


abeforgit (commented):
Why did you move the dispatcher to the app folder instead of to the addon folder with a re-export?

nvdk:
This is an old pain (which still exists and I didn't want to fix as part of this PR). The way we load the config atm only works from within the app folder.

abeforgit:
fair enough :)

PR #126, opened 2021-07-06
Commands testing + implementation xml table reader

RobbeDP:
Added tests for:
- Insert table command
- Insert table row above/below commands
- Insert xml command

Implemented xml table reader class


abeforgit (approved):
Looks great, nice job! Good selection of testcases as well

PR #127, opened 2021-07-06
Changes to getFromSelection methods in ModelTable

RobbeDP:
Change `getCellFromSelection` and `getTableFromSelection` in the `ModelTable` class to no longer make use of deprecated methods. Also improved the predicate used in `getTableFromSelection`.


abeforgit (approved):
This fixes more bugs than intended... :smile: 
Uncollapsed selections in tablecells used to break the insert-row/column commands, not anymore

PR #129, opened 2021-07-09
Remove table when last row or column gets removed

RobbeDP:
Addition to #128, so merge that one first.
Removes table node when last row or column gets removed.
Added small tests for this as well.


abeforgit (approved):
looks good!

PR #130, opened 2021-07-09
Feature/event bus

abeforgit:
Implements a basic event bus to solve https://binnenland.atlassian.net/browse/GN-2808

This is mostly done so we can easily swap out the implementation later, and serves as a example on how we could implement a strongly typed event system. Currently listeners are notified in registration order.

@nvdk would like some input on where to best emit the contentChanged event to make sure we capture all changes, which is why this is a draft pr for now

### DIFF
--- a/addon/model/model.ts
+++ b/addon/model/model.ts
@@ -110,7 +110,7 @@ export default class Model {
     if (writeSelection) {
       this.writeSelection();
     }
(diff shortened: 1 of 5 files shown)


nvdk:
@abeforgit I guess right now emiting in either updateRichNode or (perhaps better, because it debounces) in generateDiffEvents would be best for now. those should both be called after any action that might change content.

PR #131, opened 2021-07-13
Refactor of table commands

RobbeDP:
Some small changes to commands that work on tables.
Combine insert row above and below + insert column before and after.


addon/commands/insert-table-row-command.ts
```diff
@@ -1,49 +1,39 @@
...
 
-  canExecute(): boolean {
-    return true;
-  }
-
-  @logExecute
-  execute(): void {
-
-    const selection= this.model.selection;
+  execute(above: boolean, selection: ModelSelection = this.model.selection) {
```
nvdk, line 14: using a boolean to determine behavior is a bit of a code smell IMHO. I find that usually it's better to have two differently named methods in such cases. In this case I think that translates most to having two separate commands. Giving the specific behavior its own name makes it easier to look for the specific usage later on, it's usually quite hard to search for a method with a certain param value (though some IDE's can do that for you)

RobbeDP: That's a good point. I wasn't too sure about this either. I gave the insert column before and after + row above and below their own classes again, that make use of the abstraction made earlier.

abeforgit (approved):
Looks good!

PR #132, opened 2021-07-16
Cut command

RobbeDP:
This pull request introduces the cut command.
It relies on a delete selection command, which deletes all nodes in a given range and returns them.
These nodes are then used to set the clipboard data.


addon/components/ce/content-editable.js
```diff
@@ -318,19 +322,51 @@ export default class ContentEditable extends Component {
     }
   }
 
-  /**
-   * cut isn't allowed at the moment
-   */
   @action
   cut(event) {
     event.preventDefault();
+
```
nvdk, line 328: I'd consider moving this to it's own inputhandler similar to other handlers (though admittedly paste does not have a corresponding handler either atm)

nvdk (approved):
seems to work well in the basic scenarios I've tested. Perhaps consider also adding the xml representation to the clipboard and move the cut event handling to an inputhandler.

addon/commands/delete-selection-command.ts
```diff
@@ -0,0 +1,81 @@
...
+      if (ModelNode.isModelElement(firstModelNode) && firstModelNode.type === "td") {
+        contentRange = range;
+        treeWalker = new ModelTreeWalker({
+          range: contentRange,
+          descend: false
+        });
+      }
+
+      modelNodes = [...treeWalker];
+      console.log(contentRange.toString());
```
nvdk, line 58: think you missed a debug statement here

addon/commands/delete-selection-command.ts
```diff
@@ -0,0 +1,81 @@
...
+      let contentRange = mutator.splitRangeUntilElements(range, commonAncestor, commonAncestor);
+      let treeWalker = new ModelTreeWalker({
+        range: contentRange,
+        descend: false
+      });
+
+      // Check if selection is inside table cell. If this is the case, cut children of said cell.
+      // Assumption: if table cell is selected, no other nodes at the same level can be selected.
+      const firstModelNode = treeWalker.currentNode;
+      if (ModelNode.isModelElement(firstModelNode) && firstModelNode.type === "td") {
```
nvdk, line 49: for pasted tables, consider they might have  th cells for headings

addon/components/ce/content-editable.js
```diff
@@ -318,19 +322,51 @@ export default class ContentEditable extends Component {
...
+    clipboardData.setData("text/plain", htmlTextString);
+    clipboardData.setData("text/html", htmlString);
   }
 
   /**
    * copy is relegated to the browser for now
    */
   @action
-  copy( /* event */) {
+  copy(/* event */) {
```
nvdk, line 369: I realized earlier this week that we should probably provide a copy handler as well and that the behaviour would be very similar to the cut command

RobbeDP:
I found another bug where if you select a list element from the start but not till the end, the wrong selection gets copied. I will also try to fix this,

addon/components/ce/content-editable.js
RobbeDP: I think it will indeed be very similar to the cut command and we can reuse most of the code from it (except for the part that removes the selected nodes). Also, the split that gets performed when doing a cut should not be performed on the actual HTML. But since we are using a virtual DOM, this shouldn't be a big problem.

abeforgit (commented):
Super impressive, works really well and the classic "this will surely break it" cases work flawlessly

as discussed, some todos still:
- [ ] handle merge conflicts
- [ ] hide behind a feature flag (on by default)

abeforgit (approved):

PR #133, opened 2021-07-16
Make cursor move to correct position after deleting table

abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2625
also adds cursor position checks to all table-delete tests and adds a new test specifically for the buggy behavior

PR #134, opened 2021-07-16
Fix cursor behavior when using table dropdown menu

abeforgit:
I seem to have finally landed on something that works both in the dummy app and in GN.

Yes, that's an async function to close a dropdown. Yes, it waits for the next paintCycle. It's the only thing I could come up with that worked. There is just too much going on with focus changes and events and whatnot to get any sort of sensible behavior out of it (at least I cant). Maybe Sam has some more ideas, but this is the best I can do.

base branch master, together with #133 is a fix for https://binnenland.atlassian.net/browse/GN-2625

Also refactors the menu a bit with everyone's favourite pattern: contextual components!

### DIFF
--- a/addon/components/rdfa/dropdown-item.hbs
+++ b/addon/components/rdfa/dropdown-item.hbs
@@ -1,0 +1,3 @@
+<button role="menuitem" type="button" {{on "click" this.activateMenuItem}}>
+  {{yield}}
+</button>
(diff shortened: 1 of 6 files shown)


nvdk (approved):
Feels like I've approved the fix for this 10 ten times already :). don't like the paint cycle hack, but don't know a better alternative (though perhaps using embers [scheduleOnce](https://api.emberjs.com/ember/release/classes/@ember%2Frunloop/methods/scheduleOnce?anchor=scheduleOnce) might be a bit cleaner). 

It seems to be working though, so I'll merge!

abeforgit:
It feels like trying to pick the lock on an open door
_it should really not be hard_
but somehow it is

PR #135, opened 2021-07-22
convert sass division using migrator

nvdk:


Dietr (changes requested):
I would not use the sass:math function yet as it seems to break certain projects using a old or different sass version.

nvdk:
Closing since this became obsolete with the move to regular css

PR #136, opened 2021-07-23
various cleanup chores in the editor

nvdk:
- [x] convert ContentEditable component to glimmer
- [x] convert ContentEditable component to typescript
- [x] upgrade ember-concurrency
- [x] convert RdfaEditor component to glimmer
- [x] convert RdfaEditor component to typescript


nvdk:
somewhat coincidentally this also fixes https://binnenland.atlassian.net/browse/GN-2318

addon/components/rdfa/rdfa-editor.ts
```diff
@@ -99,48 +114,43 @@ export default class RdfaEditor extends Component {
...
-   * @private
-   */
-  handlers = null;
 
   /**
    * editor controller
-   *
    */
-  @tracked editor;
+  @tracked editor?: PernetRawEditor;
```
RobbeDP, line 140: since this is now undefined, there need to be some extra undefined checks when editor is accessed

addon/components/ce/content-editable.ts
```diff
@@ -248,8 +209,8 @@ export default class ContentEditable extends Component {
    * the internal state to be inline with reality
    */
   @action
-  compositionEnd(event) {
-    const preventDefault = this.passEventToHandlers( event );
+  compositionEnd(event: KeyboardEvent) {
```
nvdk, line 212: wrong type of event, should be CompositionEvent

addon/components/ce/content-editable.ts
```diff
@@ -237,8 +198,8 @@ export default class ContentEditable extends Component {
    * this method is only called for input that hasn't been handled in earlier events (like keydown)
    */
   @action
-  handleInput(event) {
-    const preventDefault = this.passEventToHandlers( event );
+  handleInput(event: KeyboardEvent) {
```
nvdk, line 201: wrong type of event, should be InputEvent

addon/components/rdfa/rdfa-editor.ts
abeforgit: one option is a getter that throws when editor is null (nothing should try to run before editor is initialized, this may also alert us to some initialization bugs we were having)
if that generates too many problems, simply asserting this is non-null (`@tracked editor!: PernetRawEditor`) is also an option, though we should avoid that if possible

addon/components/rdfa/rdfa-editor.ts
```diff
@@ -99,48 +114,43 @@ export default class RdfaEditor extends Component {
...
+    if (this.args.profile) {
+      this.profile = this.args.profile;
+    }
   }
 
-  didUpdateAttrs() {
-    if (this.profile != this.eventProcessor.profile) {
+  @action
+  updateProfile() {
+    if (this.eventProcessor && this.profile != this.eventProcessor.profile) {
```
abeforgit, line 153: might be worth going through and making sure we use triple equals everywhere (kinda surprised and slightly alarmed this doesn't trigger the linter)

abeforgit (changes requested):
Love this, typescript ahoi!

abeforgit:
@nvdk WARNING: base branch master

addon/components/rdfa/rdfa-editor.ts
nvdk: that would be a lot of linter errors:
```
nielsv@smallbox ~/Code/say-editor/ember-rdfa-editor $ grep -r ' == ' addon/ | wc -l
164
nielsv@smallbox ~/Code/say-editor/ember-rdfa-editor $ grep -r ' != ' addon/ | wc -l
27
```

addon/components/ce/content-editable.ts
```diff
@@ -66,22 +67,27 @@ export default class ContentEditable extends Component {
...
 
   /**
    *
    * @property rawEditor
    * @type RawEditor
    */
-  @tracked rawEditor = null;
+  get rawEditor(): PernetRawEditor {
+    if (!this._rawEditor) {
+      throw new IllegalAccessToRawEditor;
```
RobbeDP, line 79: i think you missed the parentheses here

addon/components/ce/content-editable.ts
```diff
@@ -91,97 +97,60 @@ export default class ContentEditable extends Component {
...
-    this.set('capturedEvents', A());
-
-    this.set('pasteHandler', new PasteHandler({rawEditor}));
-    this.set('cutHandler', new CutHandler({rawEditor}));
-
-    if(!this.externalHandlers) {
-      this.set('externalHandlers', []);
-    }
+    this._rawEditor = rawEditor;
+      this.defaultHandlers = [
```
RobbeDP, line 118: this line is indented, but it shouldn't be

addon/components/ce/content-editable.ts
```diff
@@ -332,7 +300,8 @@ export default class ContentEditable extends Component {
           break;
         }
       }
-      this.rawEditor.generateDiffEvents.perform();
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      void taskFor(this.rawEditor.generateDiffEvents).perform();
       this.rawEditor.model.read();
```
RobbeDP, line 305: just a question: why do we need to read here?

addon/components/rdfa/rdfa-editor.ts
RobbeDP: such a getter should also be added in `rdfa-editor.ts`, since we have the same problem here as in `content-editable.ts`

addon/components/ce/content-editable.ts
nvdk: not all handlers work on the vdom, so we have to read to update the model with the changes in the vdom. once all handlers work on the vdom this should no longer be needed.

addon/components/rdfa/rdfa-editor.ts
nvdk: that won't work there though, since the editor is passed in the template before init.

RobbeDP (approved):

nvdk:
@abeforgit shall I merge?

PR #137, opened 2021-07-23
Move paste handler to its own input handler

RobbeDP:
Move paste handler from `content-editable.js` to `paste-handler.ts` in `addon/editor/input-handlers`.
Also convert `html-input-parser` from javascript to typescript.
Contains some `TODO`'s.


addon/utils/html-input-parser.ts
```diff
@@ -0,0 +1,155 @@
...
+    let cleanedNode = node.cloneNode();
+    if (node.nodeType === Node.ELEMENT_NODE) {
+      const tag = tagName(node);
+      const tagMapping = this.tagMap.get(tag);
+
+      // If we have to replace the tag name we create another node with the new
+      // tag name and copy all the attribute of the original node.
+      if (tagMapping) {
+        cleanedNode = document.createElement(tagMapping);
+        this.copyAllAttributes((node as HTMLElement), (cleanedNode as HTMLElement));
```
abeforgit, line 95: the reason you need to cast cleanedNode (even though you are sure it's an element here) is that the cleanedNode variable was defined as having the Node type. Probably best to just create a new var here and only overwrite cleanedNode at the end

addon/utils/html-input-parser.ts
```diff
@@ -0,0 +1,155 @@
...
+            cleanedNode.appendChild(cleanedChild);
+
+            if (this.lumpTags.includes(tag)) {
+              // Make sure we can place the cursor after the non editable element.
+              cleanedNode.appendChild(document.createTextNode(""));
+            }
+          }
+        }
+      }
+    } else if (node.nodeType === Node.TEXT_NODE) {
```
abeforgit, line 129: there should be a similar helper to isElement for textnodes. If there isn't, make one.

addon/utils/html-input-parser.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+   * Preprocess all nodes replacing the tag if it appears on the tagMap variable
+   * and adds the lumpNode property if needed.
+   * TODO: check if we can get rid of casts
+   *
+   * @method preprocessNode
+   */
+  preprocessNode(node: Node) {
```
abeforgit, line 85: This method  could use an explicit return type (`Node | null` in this case). In general it's always good to add explicit return types, but the more complex the method, the more it's needed.

addon/editor/input-handlers/paste-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor({rawEditor}: {rawEditor: PernetRawEditor}) {
+    super(rawEditor);
+  }
+
+  // see https://www.w3.org/TR/clipboard-apis/#paste-action for more info
+  handleEvent(event: ClipboardEvent, pasteHTML: boolean, pasteExtendedHTML: boolean): HandlerResponse {
+    const clipboardData = event.clipboardData;
+
+    if (!clipboardData) {
+      //TODO: if no clipboardData found, do we want an error?
```
abeforgit, line 20: If you're not sure, maybe add a warning?
best to use a logger, imported from logging-utils
```ts
const logger = createLogger("PasteHandler");  // you can also make this a static property
logger.warn("No clipboardData object found, ignoring paste");
````

abeforgit (changes requested):
Really good work, some typescript tips and tricks in comments

abeforgit (approved):

PR #138, opened 2021-07-27
Convert event handlers to typescript

RobbeDP:
Converted following handlers to typescript:
- Arrow handler
- Disable delete handler
- Fallback input handler
- Ignore modifiers handler
- Undo handler

Didn't change the logic of these handlers.

Converted some helper functions to typescript as well to get better support.


abeforgit (approved):
:ok_hand: good work

PR #139, opened 2021-07-28
Refactor commands

RobbeDP:
The point of this PR was to convert all of the commands, so that they would only mutate the model by means of a mutator. This seemed hard in some cases and needs some more thinking (especially commands concerning tables).

What I've done:
- Converted `InsertNewLiCommand` to make use of mutator and refactored most of the code..
- Refactored most of the code of `IndentListCommand` (doesn't use mutator yet, but no use of deprecated code anymore).
- Refactored most of the code of `UnindentListCommand` (doesn't use mutator yet, but no use of deprecated code anymore).

BUGFIX: Should also fix the bug that if we have a list with two items, of which the second item is indented, this second item couldn't be unindented anymore.


abeforgit:
fixes https://binnenland.atlassian.net/browse/GN-2806

abeforgit (approved):
looks good, works well

addon/commands/insert-newLi-command.ts
```diff
@@ -18,301 +14,53 @@ export default class InsertNewLiCommand extends Command {
...
-      });
-
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // @ts-ignore
-      this.insertNewLi(newRange.start);
-    }
-    //multiple lis selected case
-    else {
-      this.deleteRange(range);
+      throw new Error("Couldn't locate parent lis");
```
abeforgit, line 35: Try to use custom errors as defined in errors.ts instead for more semantic errors
(also allows for error handling based on error type, which is the main reason for it.)

abeforgit (changes requested):
Just a small nitpick, if you could change all Errors to an appropriate CustomError as defined in errors.ts, that'd be great

PR #140, opened 2021-07-30
Copy command

RobbeDP:
As discussed in #132, most of the code written for the cut command could be reused in order to write the copy command. This is what I did here. 

First of all, I created a new command `ReadSelectionCommand`, which is almost completely identical to the `DeleteSelectionCommand`, except for the fact that the nodes in the given range are not deleted and that the changes made on the VDOM are not written back to the DOM. In order to implement this second requirement, I created a way to store the model and restore it later on.

This way of storing and restoring the model introduced for a new way to implement the undo command and this is what I did here as well. A new snapshot gets stored in the history before every command is executed (except if the command says otherwise), since normally the commands will be the only thing that can manipulate the VDOM.

I provided a seperate command for the undo and a seperate input handler for the copy command.


abeforgit (changes requested):
Works well on the surface, but we can't merge this yet because non-command document edits such as backspace don't create a snapshot, and are skipped over when undoing

RobbeDP:
This PR now only contains the changes that introduce the copy command. The model undo has been moved to #143.

abeforgit (approved):

abeforgit:
this is already behind a feature flag, so I think we can merge this

PR #145, opened 2021-08-09
Arrow handler

RobbeDP:
This PR introduces a new handler for the arrow key down event. Currently only the left and right arrow are handled. For each of these arrow keys, a separate command is created that handles that key press.

I started on the backspace handler first, but while implementing this I realized that the arrow handler could easily be implemented using the methods I already created for backspace handler. That's why I worked on this first. 

This PR contains already some files of the backspace handler, but these are not yet in working in this PR.
Merge this before #142.

`move-left-command-old.ts` and `move-right-command-old.ts` are old implementations of the commands, but I didn't remove these files yet so I can go back to these and see my thinking process if later on bugs are found in the new implementations.


lagartoverde (approved):
Works good, remember to remove the old files before merging :)

abeforgit:
regression: shift+arrow keys no longer makes a selection
@nvdk do you think that's a blocking issue?

nvdk:
I think users complained about that in the past, but I'm not sure this was possible with the previous handler. This big report seems to suggest not: https://github.com/lblod/ember-rdfa-editor/issues/28

abeforgit:
> I think users complained about that in the past, but I'm not sure this was possible with the previous handler. This big report seems to suggest not: #28

it works on dev tho?

abeforgit:
Ive changed the base branch of these PRs to preserve the work done in them. We are going through too many structural changes to merge this, but the ideas and work done here will make their way into the codebase one way or another

PR #146, opened 2021-08-10
Backspace handler

RobbeDP:
This PR aims at reimplementing the backspace handler by means of commands.
Currently, this is still a draft, but I already opened this PR, so I can leave some notes and thoughts here.

At this moment, I think I already have a sensible basic behaviour for the backspace handler. Edge cases such as lump nodes are not yet considered.

I started testing the newly created commands, but I ran against something I don't understand (see `delete-li-backwards-commands-test`).


lagartoverde (changes requested):
The cursor jumps weirdly when backspacing
To reproduce click on the list template, start at the end of the fourth line and press backspace, when the line ends the cursor jumps to the next line and goes back, also there's like 4 presses where the button doesn't remove anything (I tried to upload a video but GitHub doesn't let me don't know why, if you want write me a message on rocket and I will send it to you)

abeforgit:
yep can reproduce here as well, should be fixed before merging

abeforgit:
Ive changed the base branch of these PRs to preserve the work done in them. We are going through too many structural changes to merge this, but the ideas and work done here will make their way into the codebase one way or another

PR #147, opened 2021-08-10
Convert list helpers

RobbeDP:
Came across `list-helpers.ts` and thought it might be useful to also convert it to typescript, so here it is.


lagartoverde (approved):
Good job :)

abeforgit (approved):

PR #148, opened 2021-08-11
Delete handler

RobbeDP:
A first attempt to implement a new delete handler.
Branched from #146.


lagartoverde:
https://user-images.githubusercontent.com/17006446/131498167-97264619-b06a-4699-8b59-f6b94a3aa3bd.mov

Sometimes the delete gets stuck, otherwise works great, I tested with the lists template

lagartoverde (changes requested):
Delete gets stuck as indicated above

Asergey91:
> Delete gets stuck as indicated above

I mean you could argue that its intended behaviour.... that being said we really do have vastly different behaviours in lists/tables/text do we have some abstraction for that?

lagartoverde:
> > Delete gets stuck as indicated above
> 
> I mean you could argue that its intended behaviour....

The problem is that sometimes it works sometimes it doesn't with the same blocks, I'm okay with either approach but we want to be consistent :)

PR #149, opened 2021-08-11
Small conversion

RobbeDP:
This PR contains a small conversion from js to ts. The `rich-node-tree-modification.ts` contained some initializations with missing arguments, but I didn't know what they should be. So maybe we can't merge this.


abeforgit:
while useful, I'll close without merging here, since its not a critical change

PR #152, opened 2021-08-25
moved set property command logic to the operation

lagartoverde:


abeforgit (approved):
Works well.
Ideally the attributeoperation would become generic to handle any attribute, not just text attributes. But for me this is ok for now.

PR #155, opened 2021-09-01
Feature/model undo

abeforgit:
Since it was not immediately trivial to hide this behind a feature flag, I decided to make a new draft pr to track progress here

see https://github.com/lblod/ember-rdfa-editor/pull/155


nvdk:
closing this, since I think we have undo on vdom now

PR #157, opened 2021-09-22
Fix sass syntax error

abeforgit:
With a later version of dart sass, this line generated a syntax error


abeforgit:
part of the fix for https://binnenland.atlassian.net/browse/GN-2924

abeforgit:
@Dietr mostly to check if the change could cause issues
on first visual inspection it looks fine but I'm not familiar enough with sass to be confident about this

Dietr (approved):

PR #158, opened 2021-09-22
[HOTFIX] Fix sass syntax error

abeforgit:
hotfix pr for #157


nvdk:
looks like there's an issue with babel?

abeforgit:
It's cause CI targets include ie 11, and that doesn't work without including the babel regenerator runtime
can reproduce locally by adding ie 11 to build target

PR #159, opened 2021-09-30
faster and cleaner builds

nvdk:
Bumps ember auto import for faster (hey ed Faulkner promised!) and cleaner builds. Does require the host app to follow the upgrade guide as well, as such it's a breaking change: https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md


abeforgit (approved):
Doesn't really seem to decrease build time tho

PR #161, opened 2021-10-12
Feature/convert commands to mutators

lagartoverde:


ember-cli-build.js
```diff
@@ -14,6 +14,9 @@ module.exports = function(defaults) {
       enabled: true,
       cascade: true,
       sourcemap: true
+    },
+    'ember-cli-babel': {
+      includePolyfill: true
```
nvdk, line 19: small remark here, typically this can be avoided by including    ` "ember-maybe-import-regenerator": "^1.0.0",` instead which has a smaller footprint

PR #163, opened 2021-10-13
Document core classes

abeforgit:
Expand documentation of core architecture and move some wiki information into the code.

The idea is to eventually follow [tsdoc](https://tsdoc.org/) as a standard and use something like typedoc as a doc generator, since it's at least **some** kind of standard, but the whole doc-comment landscape is a pretty sad affair at the moment so it's not an immediate priority, since it all basically looks like jsdoc anyway and it's the text content that really matters the most.


abeforgit:
@madnificent not gonna pretend this is gonna clear up all your questions, but I hope it will help a bit at least. 
More info also here https://github.com/lblod/ember-rdfa-editor/wiki/Plugins
but probably the part you're most interested in is still missing, I'll do that tomorrow.

PR #166, opened 2021-10-18
handle tab and enter in lists

lagartoverde:


lib/lists-plugin/addon/handlers/handle-enter-in-list.ts
```diff
@@ -0,0 +1,11 @@
+import ModelTable from "@lblod/ember-rdfa-editor/core/model/model-table";
+import ModelElement from "@lblod/ember-rdfa-editor/core/model/model-element";
+import EditorController from "@lblod/ember-rdfa-editor/core/editor-controller";
+import { PropertyState } from "@lblod/ember-rdfa-editor/util/types";
+ 
+export default function HandleTabInList(controller: EditorController) {
```
abeforgit, line 6: please use camelCase for functions

lagartoverde: yep they were a class before :P

lagartoverde:
This is currently not working as they need to cancel the tab event, I will do that now

PR #171, opened 2021-10-21
Add ember-appuniversum

Dietr:
Add ember-appuniversum as dependency so we can remove the duplicate components in plugins over time.
- simplified the theming: bundled all variables in one _s-theme file
- seperated the custom components in a file so we can only load these in Gelinkt Notuleren
- added some ember-appuniversum variables so we maintain the say style (purple primary color)

Next steps:
- update Gelinkt Notuleren. See also: https://github.com/lblod/frontend-gelinkt-notuleren/pull/212
- refactor all plugins to use ember-appuniversum so we can remove the deprecated components
- convert appuniversum sass variables to css variables to remove the sass dependency for vendors

Open questions:
- how do we best handle different versions of ember-appuniversum in GN and ember-rdfa-editor / plugins?


abeforgit:
> On the open questions:
> 
>     * plugins should specify app universum as a peerdependency and be as lenient as possible on version requirements.

Shouldn't the editor itself then also specify appuniversum as a peerdep?

PR #173, opened 2021-10-25
Internal/navigation improvement

lagartoverde:
Not 100% happy with the result to be honest but I feel like I could spend a lifetime working on this, it handles all the cases I encountered, but for sure there will be some that will not, I will switch to a different thing and come back later to improve, as I feel like there's more important things than getting the perfect keyboard navigation. It's good enough imho


abeforgit:
superceded by #268

PR #176, opened 2021-11-02
[RFC] ember-rdfa-editor stage 1

abeforgit:
This RFC serves as a vision document of how the architecture of the editor should evolve. 

The point of this PR is to discuss the design and close any hole we can identify.  Everything is up for discussion.

[rendered](https://github.com/lblod/ember-rdfa-editor/blob/rfc/structural-rework/rfcs/rdfa-editor-stage-1.md)

### DIFF
--- a/rfcs/rdfa-editor-stage-1.md
+++ b/rfcs/rdfa-editor-stage-1.md
@@ -173,6 +173,8 @@ interface Controller {
   name: string;
   executeCommand<A, R>(commandName: string, args: A): R;
   registerCommand<A, R>(command: new(): Command<A,R>): void;
+  registerWidget(spec: WidgetSpec): void;
+  registerNodeType(spec: NodeSpec): void;
   mutator: Mutator;
   selection: ModelSelection;
   emitEvent(event: EditorEvent);
@@ -239,7 +241,7 @@ a good amount of context so listeners can filter for content changes they care a
 
 This represents a single node in the document tree. It is in essence on its own an entirely
 valid document. 
-The node class and its descendants should provide methods to manipulate nodes in an immutable way.
+The node class should provide methods to manipulate nodes in an immutable way.
 
 *interface*
 
@@ -248,6 +250,7 @@ interface ModelNode {
   type: string;
   children: ModelNode[];
   parent?: ModelNode;
+  
 }
 ```
 
@@ -517,23 +520,65 @@ interface Command<A> {
 }
 ```
 
+### Plugins
+
+With extensibility as one of the core pillars, plugin support should be considered at every level. The above
+architecture gives plugins the following capabillities:
+
+- defining new node types
+- defining new commands
+- executing commands
+- updating the vdom through a mutator (either as part of a command, or directly)
+- listening to events
+- defining and emitting new events
+- defining new UI elements and registering them for rendering
+
+```typescript
+interface EditorPlugin {
+  name: string;
+  async initialize(controller: Controller): void;
+}
+```
+#### Widgets
+
+A widget is simply a ui component which can be registered in the editor. This can be a simple menu, a single button, or even a full-blown ember application.
+At this stage, only ember-based plugins are supported, and widgets have to be ember components. There are limited locations where a widget can be registered, 
+which are specified below.
+A widget is always rendered, and as such has full control over when it is visible. 
+
+```typescript
+type WidgetLocation = "toolbar" | "sidebar" | "cursor" | "block";
+
+interface WidgetSpec {
+  componentName: string,
+  desiredLocation: WidgetLocation
+}
+```
+
+#### Adding a node type
+
+Adding a new node type is very similar to adding a command or a widget. Simply call the appropriate method in the controller, providing it with a Spec object defined below.
+The spec object demands a reader and writer to respectively deserialize and serialize the node from and to the actual DOM.
+
+```typescript
+
+interface NodeSpec {
+  name: type;
+  implementation: new () => ModelNode;
+  htmlReader: Reader<Node, ModelNode>;
+  htmlWriter: Writer<ModelNode, Node>;
+}
+```
 
 
 ## Drawbacks
 
-> Why should we *not* do this? Please consider the impact on teaching Ember,
-on the integration of this feature with other existing and planned features,
-on the impact of the API churn on existing apps, etc.
-
-> There are tradeoffs to choosing any path, please attempt to identify them here.
-
 ## Alternatives
-
-> What other designs have been considered? What is the impact of not doing this?
-
-> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
 
 ## Unresolved questions
 
-> Optional, but suggested for first drafts. What parts of the design are still
-TBD?
+It is as of yet unclear how to best extend node types in plugins. The proposed method works well for "core" plugins which are developed as 
+in-repo-addons, but for external plugins would require a hard dependency on the editor itself 
+to be able to extend the base node class. Other options such as mixins
+or delegating custom node methods to a NodeType property on the base class (maybe even with use of Proxies)
+needs to be explored.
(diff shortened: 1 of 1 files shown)


abeforgit:
Likely certain sections don't make much sense or contain errors, don't hesitate to point those out

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+The main argument against this usually involved control over state changes. How
+do we implement a `table.insertRow` method while also making sure we can control and track any changes
+to the document? The current implementation of mutators and operations provides a solution, but
+we run into weird situations where we have methods on nodes which all take in a mutator, which they need to get from somewhere.
+It also puts a lot of strain on the mutator's api surface, as different node types can have different abstractions and might
+want to interact with their content in different ways.
+
+These problems dissapear if we apply the immutability principle here. If we demand that any node-owned method
+does not mutate the node itself (rather returns a new node with desired values), we can reduce the mutator api surface
+to convenient ways to replace existing nodes (or ranges) with their newly minted content. Prosemirror shows us that this
```
lagartoverde, line 110: I'm not really convinced by this, are you sure that prosemirror doesn't calculate the differences between the old node and the new node so it doesn't rerender it? I know React does this with the shadow dom so that may be the case with prosemirror also

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+approach can be implemented in a sufficiently performant way. 
+
+This demand is consciously not enforced, as javascript is a notably _very_ mutable language, and _actually_ making things immutable
+comes with a cost. Plugin authors which implement their own node types are operating on the lowest possible level, and 
+are expected to read documentation and adhere to conventions, just as someone assembling their own computer from parts can be expected
+to handle components with care and follow instructions.
+
+I also consider allowing plugins to extend the available node-types a very powerful and expressive tool. Rather than concerning
+oneself with node-structures, one could implement a TodoItem nodetype, let's say, with natural methods such as `setStatus()` and `schedule()`,
```
lagartoverde, line 119: Also interesting for templates and inputs where they can be nodes with special (on top of the model) components like you show on the demo

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+- does a range contain/touch a triple with a certain shape?
+- given a triple, which node defines its subject, which its predicate and which its object? (this also needs to support duplicate triples)
+- give me all nodes which are the subject of a given triple
+- give me all nodes which are the object of a given triple
+
+If this is starting to sound a lot like the questions you would ask a triplestore, that makes a lot of sense.
+In essence, the document _is_ a triplestore, just a very clumsy one, and one with added structure (the html markup).
+
+The idea here is to provide a single interface for all these questions. Whether the knowledge required for the questions is
+calculated ad-hoc or kept up-to-date throughout document edits is not specified here. Likely the former
```
lagartoverde, line 321: What about the initial load? I think keeping the store up to date could be expensive but not really impactful when changing the document (if you are just writing or deleting normally) but what I'm afraid is how much will take for long documents to load or when deleting/pasting large amounts of texts.
Anyway I think the datastore is a good idea, better than the context we had before

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+This does mean that we lose the ability to express "just move one character, regardless of tree structure". 
+
+Another way to represent positions is to embrace the tree structure and represent positions as paths of offsets from
+the root node. This is how the current ModelPositions work.
+The benefit of this representation is that it encodes way more information about the tree structure it operates in. It is
+also easier to calculate.
+
+As this second representation is currently implemented, this will be what we will use. However,
+the main reason for the small detour about character offsets above serves to illustrate that 
+both representations have merit, and a conversion between the two could be desirable.
```
lagartoverde, line 430: It would be very simple for the arrow handlers if we could just do position.getDocumentOffset() and ModelPosition.fromDocumentOffset() or something like that

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+```
+
+### Plugins
+
+With extendability as one of the core pillars, plugin support should be considered at every level. The above
+architecture gives plugins the following capabillities:
+
+- defining new node types
+- defining new commands
+- executing commands
```
lagartoverde, line 530: For executing commands we will have to talk about namespacing, can a plugin call another plugin command? if so how will it be uniquely identified?

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+#### Adding a node type
+
+Adding a new node type is very similar to adding a command or a widget. Simply call the appropriate method in the controller, providing it with a Spec object defined below.
+The spec object demands a reader and writer to respectively deserialize and serialize the node from and to the actual DOM.
+
+```typescript
+
+interface NodeSpec {
+  name: type;
+  implementation: new () => ModelNode;
```
lagartoverde, line 567: What should be inside the implementation? Not 100% clear

lagartoverde (commented):
Good job, left some questions and comments

rfcs/rdfa-editor-stage-1.md
abeforgit: it does, that's what I mean that it can be implemented efficiently

rfcs/rdfa-editor-stage-1.md
abeforgit: yeah I don't know how the implementation will be, thats why this is just an interface.
I think initially we calculate things on the fly and see where we can gain performance by precalculating/caching

rfcs/rdfa-editor-stage-1.md
abeforgit: that's typescript syntax for "a constructor for a ModelNode"
aka you pass in the class itself

rfcs/rdfa-editor-stage-1.md
abeforgit: yes, plugins can call other plugin commands
namespacing is indeed an issue with the current design.
I'd like to address that in a stage 2 rfc, I think for now we can get by with naming conventions

rfcs/rdfa-editor-stage-1.md
abeforgit: I think for arrow handlers I'd still like to lean on browser fallbacks
but yeah I think both systems can be useful. I don't think its a priority though

rfcs/rdfa-editor-stage-1.md
abeforgit: yeah custom nodes come with their own reader/writers
current design doesn't allow for inline components that way, I'd consider that for stage 2 as well

rfcs/rdfa-editor-stage-1.md
lagartoverde: isn't that kinda hard to implement?

rfcs/rdfa-editor-stage-1.md
```diff
@@ -0,0 +1,587 @@
...
+```typescript
+interface Controller {
+  name: string;
+  executeCommand<A, R>(commandName: string, args: A): R;
+  registerCommand<A, R>(command: new(): Command<A,R>): void;
+  registerWidget(spec: WidgetSpec): void;
+  registerNodeType(spec: NodeSpec): void;
+  mutator: Mutator;
+  selection: ModelSelection;
+  emitEvent(event: EditorEvent);
```
nvdk, line 180: no emitDebounced here?

abeforgit: the idea was to set the high-level goals, not define all the interfaces precisely. interface definitions are minimal and can be extended.

abeforgit:
Would like some input about marks. They really only make sense on leaf (aka text or <br>) nodes.
The distinction between leaf and non-leaf can be implemented in 2 ways:
- different classes (similar to ModelText and ModelElement)
- properties (using tagged union types for smart typing)

leaning towards the second way mostly because OOP sort of doesnt match very well with the current interaction pattern between plugins and core (through the controller and not via ES module imports)
also we constantly have to typecheck dynamically anyway (just look at how many times we use ModelNode.isModelElement now) and tagged unions may be a bit more ergonomic there

PR #179, opened 2021-11-04
Extend and improve eventbus

abeforgit:
Adds priorities and stoppropagation logic


lagartoverde (approved):
Looks good, just one question it being a EditorEvent means that we can preventDefault even if the function is not there?

nvdk:
> Looks good, just one question it being a EditorEvent means that we can preventDefault even if the function is not there?

I think you mean stopPropagation and every event should have that, whether or not stopPropagation also includes event.preventDefault on the dom event depends on the EditorEvent implementation I think

PR #180, opened 2021-11-04
Add controllers

abeforgit:
Add controller interface and raweditor-compatible implementaiton more or less as specified in the RFC (with some compromises)


addon/model/controller.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  get selection(): ModelSelection {
+    return this._rawEditor.selection;
+  }
+
+  executeCommand<A extends unknown[], R>(commandName: string, ...args: A): R | void {
+    return this._rawEditor.executeCommand(commandName, ...args);
+  }
+
+  offEvent<E extends AnyEventName>(eventName: E, callback: EditorEventListener<E>, config?: ListenerConfig): void {
```
lagartoverde, line 52: What's the meaning of this function? Not sure I understand

abeforgit: to remove an event listener

PR #181, opened 2021-11-08
Add a new way to initialize plugins

abeforgit:
The system exists separately from the existing way of registering plugins and is fully backwards compatible.
Makes sure a bare-minimum toolkit is provided to plugins in the form of the RawEditorController, which currently only implements a few methods of the RFC

### DIFF
--- a/addon/components/rdfa/rdfa-editor.ts
+++ b/addon/components/rdfa/rdfa-editor.ts
@@ -1,9 +1,9 @@
-import { debug, warn } from '@ember/debug';
-import { action } from "@ember/object";
-import { inject as service } from "@ember/service";
+import {debug, warn} from '@ember/debug';
+import {action} from "@ember/object";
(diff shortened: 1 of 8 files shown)


addon/components/rdfa/rdfa-editor.ts
```diff
@@ -177,8 +197,12 @@ export default class RdfaEditor extends Component<RdfaEditorArgs> {
    * @private
    */
   @action
-  handleRawEditorInit(editor: PernetRawEditor) {
+  async handleRawEditorInit(editor: PernetRawEditor) {
+    console.log("yeet")
```
lagartoverde, line 201: You forgot about this console.log :)

lagartoverde (approved):
Looks good overall 👍 , pretty hard to test as this is very internal though

nvdk (approved):
looks good to me!

PR #182, opened 2021-11-15
Add a more consistent and flexible treewalker

abeforgit:
Adds a new treewalker which in time is meant to replace the current one, but I'm keeping them side by side for the moment.
Walking logic is still the same, but now actually working properly (I hope).

### New features:
- onEnterNode and onLeaveNode handlers (kinda the reason I started doing this). Needed for rdfa-parsing and to enable use of the walker for all DFS needs. Currently ignores the filter function (aka skipped nodes still trigger this). Not sure if thats a good thing or not.
- generic interface, designed to be compatible with DOMnodes as well as modelNodes. Makes as little assumptions about the nodes as possible. 
- walk from root: aka "regular" preorder DFS through a subtree
- start/endnode walking
- more consistent range walking, following the metaphor:
    "select range in xml representation, start reading in normal order, when you encounter an opening tag, visit, dont visit on closing tag" (this was always the intended way but wasn't handled right)
- more sensible behavior with collapsed ranges
- reverse: all walking styles can be reversed. 
  - when using ranges, range start and end are automatically reversed, so in the above metaphor, do the same thing but read backwards and visit on closing tags.
  - when using root-walking, equivalent to [reverse-prefix DFS](https://en.wikipedia.org/wiki/Depth-first_search#Vertex_orderings) (_not_ postfix)
  - when using start/end nodes, caller is responsible to make sure endnode is before startnode.
- iterator now implemented using generator, simplifying things a bit
- ability to reset the generator state
  
### Missing features:
- visitParentUpwards and descend toggles (NotImplementedErrors are thrown when used)
- all other walking methods besides nextNode (tho tbh, who was using them anyway)
- FilterResult.REJECT is simply treated the same as FilterResult.SKIP. This is temporary. Handling the end-node while ignoring descending into children is not trivial.


abeforgit:
Note: just noticed filtering is untested, feel free to add if you have a minute, if not ill try to add asap

lagartoverde:
Kinda confused, does this mean that we are going to implement our own rdfa parser?

nvdk:
> Kinda confused, does this mean that we are going to implement our own rdfa parser?

well that was always a given, we already had that in marawa but will need one that understands the vdom and integrates more tightly with our use case.

lagartoverde:
yeah, but yesterday I was talking with arne about modifying https://github.com/rubensworks/rdfa-streaming-parser.js just a little bit in order to adapt to our use case, does this mean we are not doing this anymore?

lagartoverde (approved):
Anyway it looks good :)

abeforgit:
> yeah, but yesterday I was talking with arne about modifying https://github.com/rubensworks/rdfa-streaming-parser.js just a little bit in order to adapt to our use case, does this mean we are not doing this anymore?

Nono we still do that, this is just the DFS walking logic. The streaming parser uses onOpenTag and onCloseTag methods, which are represented here by onEnterNode and onLeaveNode, that's the main reason I made the new walker (combined with all the bugs in the old one)
Im just sick of reimplementing DFS walking all over the place

PR #183, opened 2021-11-18
Bugfix: we should not have contenteditable tables exported outside the editor

Asergey91:
fixes a bug where after saving tables would still have a cursor inside them


Asergey91:
Ok will try it and test it, but doesn't this get called when you update the document as opposed to only when it gets saved?

abeforgit:
This writer gets called whenever you ask for the htmlContent as a consuming app

nvdk:
but that's the same for the export writer right, that just uses the unpolluted-html-element-writer. looks good to me, merging

PR #184, opened 2021-11-19
Use css variables

Dietr:
Use css variables for ember appuniversum:
- update ember-appuniversum
- convert sass variables to css variables

**Todo**
- [x] update documentation
- [x] test styling implementation in GN and plugins

**Breaking changes**
- all scss variables are replaced by css variables. Vendors or applications using these should convert these.
- all scss toggles (paper, toolbar position, sidebar) have been replaced by component config settings. Vendors or applications using these will have to check if they still have the correct settings.

**Questions/help needed**
- How do I correctly use the peerDepedencies for appuniversum? When I don't have appuniversum in devDepedencies it breaks the test.
- We should find a way to generate the plain css on release to completely remove the sass dependency. Ideally also provide an option to not load this css when using sass. (See https://ember-power-select.com/docs/styles for an example approach)


abeforgit:
for dependencies:
we will always need appuniversum in devdeps, since devdeps are whats used for running the ember dummy app
when using the editor as an addon the devdeps wont be included anyway. There we can use peerdeps to give a signal to the host app that it needs to have appuniversum loaded

PR #185, opened 2021-11-26
Add the datastore api

abeforgit:
Adds a new `datastore` object owned by the raw-editor instance. It is available on all controllers.

Its api is kept as simple as possible. It's a fluent interface with the main idea being that you progressively limit the dataset you are interested in through a combination of `match`, `limitToRange` and `transformDataset` calls, all of which return a **new** datastore instance (none of the methods mutate the original datastore!), and then consume the left-over triples in whatever format you want, as a generator (to allow for lazy evaluation).

example usages:
```typescript
// find out if cursor is inside a node which has a certain type
datastore.limitToRange(selection.lastRange, "rangeIsInside").match(null, "a", "besluit:Agendapunt").size > 0;

//find all nodes in document which have a "firstName" property
const rslt = [...datastore.match(null, "foaf:firstName").asSubjectNodes()]
// you can see what I mean with this being a bit clunky
rslt.map(({subject, nodes}) => {console.log([...nodes])})

//do some crazy set logic on the underlying dataset
datastore
    .transformDataset(
        // the underlying dataset's match method doesnt support concise terms, so we use the provided converter
        (dataset, converter) => dataset.difference(otherDataset).match(null, converter("a"), converter("schema:Person"))
     )
    .limitTorange(range, "rangeContains")
    .asQuads()
```

Currently the only available consumers are `asSubjectNodes` and `asQuads`.
There are 2 more planned: `asObjectNodes` and `asPredicateNodes`. They currently throw a NotImplementedError, cause I want people to assume they exist and think as if they do, giving us a chance to evaluate usecases where they come in handy. 
They shouldn't be too difficult to implement.

Currently the format of items you get back from `asSubjectNodes` is structured as follows:
```typescript
interface SubjectNodesResponse  {
  subject: RDF.Quad_Subject,
  nodes: Set<ModelNode>
}
```
aka the nodes are grouped by subject, but are provided in no particular order. I can already see this causing issues, and it's not the most intuitive format, but it was the easiest.

### Concise terms

Nobody likes writing triples as RDFjs interfaces, so with inspiration from graphy I implemented a simplified version of their [concise triples](https://graphy.link/concise#string/c1) syntax. The idea is to identify node types with a single prefix character to make parsing very simple (aka no closing delimiters, as in turtle)
It is loosely inspired by turtle syntax. Heres a quick overview (also refer to the test file for usage examples)

absolute iri: `">http://schema.org/Person"`
prefixed iri: `"schema:Person"`
languaged literal: `'@nl"blabla'`
datatyped literal: `'^>http://w3.org/types/thing"blabla'` (note the datatype can also be a prefixed iri following the above syntax rules)
explicit literal string: `'"this:will:not:be:>interpreted:as:prefixes'`
rdf "type" predicate alias: `"a"`
blanknode with specific label: `"_:label"`

basic JS datatypes `number` and `boolean` are also supported and will be automatically converted to the right datatyped literal

When no prefix is found, it is assumed to be a literal string.

Concise terms are accepted by the `match` method of the datastore. (Note, it is not accepted by the underlying dataset, but a converter is provided in the `transformDataset` escape hatch, so you can still use it)

### performance

currently the datastore is rebuilt on every change event (this is already debounced at 1sec). I believe minimal updates are possible and should be just as hard to implement as minimal dom updates (likely implementation of either will also imply the other)

The parser seems fast enough for now though (rebuilding only takes linear time)
There are perfomance problems on chrome but they seem unrelated to this logic


nvdk:
> Nobody likes writing triples as RDFjs interfaces, so with inspiration from graphy I implemented a simplified version of their concise triples syntax. 

<3

addon/model/util/constants.ts
```diff
@@ -13,5 +13,9 @@ export const HIGHLIGHT_ATTRIBUTE = 'data-editor-highlight';
 export const PLACEHOLDER_CLASS = 'mark-highlight-manual';
 export const CORE_OWNER = 'rdfa-editor';
 
+export const RDF_TYPE = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type';
+export const USES_VOCAB_PREDICATE = 'http://www.w3.org/ns/rdfa#usesVocabulary';
```
nvdk, line 17: cool didn't know about this one :)

addon/utils/dom-helpers.ts
```diff
@@ -494,3 +496,18 @@ export function getWindowSelection(): Selection {
 
   return selection;
 }
+
+export function getPathFromRoot(to: Node, inclusive: boolean): Node[] {
```
nvdk, line 500: is dom-helpers the right place for this method?

nvdk:
do we still need / use the rdfaAttributes on an element? there seems to be some overlap with the parsing we now do and it might make sense to drop that?

nvdk (approved):
as with anything keeps references to dom elements, there's a risk for memory leakage and I didn't look into that. other than that seems to look and work fine.

lagartoverde (approved):
Looks pretty cool to be honest :), can't wait to use it, maybe you should save the PR description as it explains really well how the rdfa works

nvdk:
> Looks pretty cool to be honest :), can't wait to use it, maybe you should save the PR description as it explains really well how the rdfa works

let's copy it to a wiki article!

addon/utils/dom-helpers.ts
abeforgit: > as with anything keeps references to dom elements, there's a risk for memory leakage and I didn't look into that. other than that seems to look and work fine.

I think its fine cause everything is implemented in immutable style, datastore gets rebuilt on every parse so old ones should be garbage collected

abeforgit:
> as with anything keeps references to dom elements, there's a risk for memory leakage and I didn't look into that. other than that seems to look and work fine.

I think its fine cause everything is implemented in immutable style, so the datastore gets rebuilt on every parse and thus should be GCd

abeforgit:
> > Looks pretty cool to be honest :), can't wait to use it, maybe you should save the PR description as it explains really well how the rdfa works
> 
> let's copy it to a wiki article!

https://github.com/lblod/ember-rdfa-editor/wiki/Datastore

PR #187, opened 2021-11-30
Add package lock

abeforgit:
I've had enough.

No illusions that this is gonna magically make everything better, but it very much can't really get worse so what the hell.

Stick to npm v6.10 or above and you should be fine (aka **not** npm v7 or v8)
recommended reading: https://docs.npmjs.com/cli/v6/configuring-npm/package-locks


abeforgit:
@nvdk maybe CI should be updated to use the package-lock file? https://docs.npmjs.com/cli/v6/commands/npm-ci

PR #195, opened 2021-12-05
Update eslint and various non-ember plugins to latest

abeforgit:
Leaving the ember ones cause they might be tied to ember version, it seems ember-cli-upgrade bumps those


abeforgit:
seems to break the build _sigh_

abeforgit:
fixed by reverting eslint-prettier
see https://github.com/prettier/eslint-plugin-prettier/issues/434

nvdk:
I think at least eslint etc are also managed by ember-cli, so might not be the best idea to upgrade them separately?

nvdk:
Since I squash merged some PRs I've rebased this one on dev.

abeforgit:
wont squash merging break the lerna-changelogs?

abeforgit:
> I think at least eslint etc are also managed by ember-cli, so might not be the best idea to upgrade them separately?

hmm what does "managed" mean in this case? just that it installs and upgrades it?
if so I dont really see the harm tbh

PR #196, opened 2021-12-05
Update release-it to 14.11.8 and changelog to 4.0.1

abeforgit:
Couldnt spot any significant breaking changes
also updated the release gitbook guide to include the local npm release script route
it seems neater


nvdk:
must have messed something up, because this was merged but not picked up by github...

PR #200, opened 2021-12-06
bump docker ember image

nvdk:
the docker ember image we were using for our builds was quite old, i don't think
we actually use the build version of the rdfa editor anywhere... but just in
case :).


abeforgit:
did you test it?

nvdk:
I tested that it builds, does that count?

PR #203, opened 2022-01-18
Expose model range on controller

lagartoverde:


abeforgit:
not needed, as discussed. 
for reference: use `controller.rangeFactory`

PR #205, opened 2022-01-19
Fix broken datastore in prod

abeforgit:
Terser removes unused code by default, which got rid of the consumption of the walker, which meant the side-effects didn't trigger, which meant the rdfa-parser wasn't actually parsing very much.

_sigh_

So ignoring the "unused variables" lint has some consequences, watch out.


nvdk:
some nice sleuthing, seems to work with npm link in GN!

PR #206, opened 2022-01-19
Feature/gn 3152 create a debug component for the rdfa editor

benjay10:
You should be able to use a debug component for the editor by using `<Rdfa::RdfaEditorWithDebug .../>` instead of the normal `<Rdfa::RdfaEditor .../>` component with the same arguments. For example:
```handlebars
<Rdfa::RdfaEditorWithDebug
  @rdfaEditorInit={{this.rdfaEditorInit}}
  @editorOptions={{hash
    showToggleRdfaAnnotations="true"
    showInsertButton=null
    showRdfa="true"
    showRdfaHighlight="true"
    showRdfaHover="true"
    showPaper="true"
    showSidebar="true"
    showToolbarBottom=null
  }}
  @toolbarOptions={{hash
    showTextStyleButtons="true"
    showListButtons="true"
    showIndentButtons="true"
  }}
/>
```


tests/dummy/app/router.ts
```diff
@@ -8,5 +8,6 @@ export default class Router extends EmberRouter {
 }
 Router.map(function () {
     this.route("vendors");
+	this.route("debug");
```
abeforgit, line 11: I dont think this new route is necessary, just replace the index route with the component usage

package.json
```diff
@@ -58,14 +58,14 @@
     "ember-maybe-import-regenerator": "^1.0.0",
     "ember-truth-helpers": "^2.0.0",
     "mdn-polyfills": "^5.16.0",
-    "xml-formatter": "^2.4.0"
+    "xml-formatter": "^2.4.0",
```
abeforgit, line 61: @nvdk not sure how to deal with these
Ideally they'd be optional peer deps, but that's only a thing from node v7 and up
we could make them peerdeps and just ignore the warnings
looking at the syntax we may be able to already mark them as optional for when we do eventually switch to a new node (so we dont forget)

abeforgit (changes requested):
Looks good overall. Main thing is the branch, we tend to branch our feature branches off of development and make PRs against that.
I can resolve the conflicts for you if you're not comfortable doing that

abeforgit:
> Looks good overall. Main thing is the branch, we tend to branch our feature branches off of development and make PRs against that. I can resolve the conflicts for you if you're not comfortable doing that

about this, notice ive changed the base branch of the PR already. Here's how to proceed:
- fix the other review comments first
- rebase this branch on top of development (rebasing can be annoying, we can go through this together)
- `git push --force-with-lease` (always use `force-with-lease`, never `--force`, its basically a safer version of force preventing you from overriding other people's work)

tests/dummy/app/router.ts
benjay10: Okay, then the index template will also be replaced with the new component. Will do that in a next commit.

package.json
abeforgit: FYI went this route, dont see any immediate issues with it, to be monitored

benjay10:
I tried reopening the branch to see if I could add more fixes to this thing, but I then forgot to switch to that branch and pushed to `development` instead. I hope nothing breaks because of the `stream-browserify` dependency.

PR #207, opened 2022-01-20
moved hints logic to the editor component so it gets tracked

lagartoverde:
I tested and it seems to work, basically the suggested hints were in 2 places, it was rendering this.args.suggestedHints and it was clearing this.suggestedHints, I moved the closing logic to the parent component and now it works, but it's the editor so if someone who knows more than me about it (ehem ehem arne) can take a look it would be nice :)


nvdk (approved):
looks good to me!

PR #209, opened 2022-02-02
Feature: Marks and MarksRegistry

abeforgit:
Let me start by apologizing the mess of commits this pr dumps on the reviewer. Furthermore, documentation is as of yet non-existent. I'm hesitant to start writing it since a lot might still change soon. 

This PR can also be considered a refinement and implementation of the Mark concept from [the rfc](https://github.com/lblod/ember-rdfa-editor/blob/rfc/structural-rework/rfcs/rdfa-editor-stage-1.md#mark-immutable), which is of course heavily inspired by [prosemirror](https://prosemirror.net/docs/ref/#model.Mark).

### `mark.ts`

This file outlines the primary definitions of what it means to make a Mark and is a good place to start. 

The most important interface is the MarkSpec. It contains the definition of how to read and write a mark from and to HTML. Plugins can provide their own marks using `controller.registerMark(markSpec)`

The mark class itself is not much more than a factory to create instances of marks based on a spec.

You'll note here a definition of a RenderSpec, which is in essence a mini-DSL for specifying HTML trees. Think of it as baby-jsx or something. The HOLE is where the block content gets rendered.

### `marks-registry`

This is what all the fuss is about. It maintains a live registry of which marks are registered and which marks are active where, and by which plugins they were set. It makes use of events sent out by operations, so it can operate completely independently.
The implementation is quite basic and greedy. It reacts to change-events and recalculates any marks in the changed range.

### `plugins/basic-styles`

Here you can find the mark-based implementation of bold, italic, underline, and strikethrough.  I've moved these out into their own plugin to demonstrate how I'd like the codebase to evolve.
This plugin is always loaded, which you can see happen in the `getPlugins` method of rdfa-editor.ts` Eventually this could even move into its own package with a monorepo kind of setup, but that's unimportant for now.

### Commands

4 new commands were added: add and remove commands for marks based on a range, and based on a specific mark.  In implementing a highlighting-style plugin I've found both to be needed.

### tests/dummy/app/testplugin

This is an example on how to make plugins specific to the dummy app for easy testing of ideas.

## Core changes

In order to accomodate the new infrastructure, significant internal changes were needed.

### Gen-tree-walker bugfix

Due to some really boneheaded code (by me) the walker would get stuck in an infinite loop under certain conditions. A test was also added to make sure this doesn't happen again.

### ContentChanged event

The contentchanged event went from a void event to carrying quite a useful payload. It's designed to give listeners an idea of _what_ changed, however, due to the current architecture, this concept is harder to pinpoint than it sounds.
This interface will probably remain janky and experimental until we land immutable nodes and proper change management, but it serves its purpose for now.

### `mutator.insertText`

Due to operations now emitting events, and the marks-registry listening to them, the  implementation of basic text insertion became problematic. It used a strategy of "insert, then loop and fix loose ends" which used many more operations than actually needed, leading to many more event fires than needed.

Therefore the logic has been rewritten from the ground up to properly handle textnode merging in one go. The code itself is long-winded and can probably be made shorter, but I liked (aka needed) the clarity of having 1 case per if-statement.

### `insert-text-command`

Here's the dirty bit of this PR. Due to the changes in the insertText logic, I had to remove the bit that provided whatever basic partial updating we had. So a keystroke now rerenders the entire page. Ouch. 
TBD if we monkey-patch this once again or shoot for actually reasonable writing logic.

### readers and writers

Understandably, the readers and writers have been changed to support Marks. Thanks to the mark design, the added logic is minimal since it's contained in the markSpec.

### `model/util/hash-set.ts`

Those who've known me for a bit know that I like to lament about the sorry state of the builtin Javascript Map and Set collections. I've finally snapped and implemented an expanded version of a Set. It's backed by a Map, and takes a hashing function which is used to implement the normal Set methods (aka objects are compared by their hashes, not by reference equality).


addon/components/rdfa/rdfa-editor-with-debug.hbs
```diff
@@ -10,7 +10,7 @@
     </p>
     <div>
       <label for="pasteSelect">
-        paste behaviour:
+        paste behaviour:&
```
nvdk, line 13:
```suggestion
        paste behaviour:
```
woops

PR #210, opened 2022-02-03
[RFC] whitespace handling

nvdk:
whitespace handling in the editor is a nuissance and we've had more than our share of bug reports on it. Based on earlier input from @abeforgit I wrote out a possible solution. This needs further refinement, thus the RFC

[rendered](https://github.com/lblod/ember-rdfa-editor/blob/rfc/whitespace-handling/rfcs/whitespace-handling.md)

NOTE: this explicitely doens't cover cursor positioning in empty nodes, which is considered a separate issue. (for more on that see https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_view_filler.html#function-MARKED_NBSP_FILLER

### DIFF
--- a/rfcs/whitespace-handling.md
+++ b/rfcs/whitespace-handling.md
@@ -1,0 +1,51 @@
+---
+Stage: Draft
+Start Date: 01-02-2022
+Release Date: unreleased
+---
+
+# whitespace handling in rdfa-editor
+
+## Summary
+Whitespacing handling in contenteditable divs is a nuisance, for one its default behaviour is unconsistent accross browsers. Previously we've resorted to a hack by replacing a space with non breaking spaces when a user types a space. This nbsp would then be replaced with a regular space again if the following character inputted was not a space. The latest implementation of this hack is somewhat lacking and not all nbsp's are replaced properly, however even in the past spurious nbsp's would remain. 
+
+To further complicate things browsers themselves sometimes convert regular spaces to non breaking ones, which is why we should not rely on default browser behaviour for inserting spaces. See [this wordpress bug](https://core.trac.wordpress.org/ticket/31157) for example.
+
+This RFC proposes a different approach using CSS styles to ensure all whitespace is shown. This seems a more elegant solution and moves the difficulty of whitespace handling to loading into - and exporting content from the editor.
+
+## Motivation
+non breaking spaces make it hard to render text properly as it's unclear where the browser can break up sentences for overflowing text. If we can use css to make spaces visible without the `nbsp` hack we can forego complex/ugly logic to clean up spurious `nbsp`'s.
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


rfcs/whitespace-handling.md
```diff
@@ -0,0 +1,69 @@
...
+
+### loading content
+Load content through the insert-html command (and others) becomes more complex because we will want to filter out any unwanted whitespace and linebreaks.
+
+### exporting content
+Exporting content becomes more complex because we will want to translate whitespace that would normally fold (consecutive spaces, linebreaks, spaces at the end of a node, etc.) to something visible.
+
+## Caveats
+### are we the only ones?
+It's unclear whether other editors use this css property and that's something we should investigate. If they don't, perhaps there's a good reason not to do this.
```
nvdk, line 60: https://github.com/ProseMirror/prosemirror/issues/651 seems to confirm prosemirror is also using `white-space: pre-wrap`

nvdk:
one thing to take into account is that this will mean we can't implement a "text-align" feature later on as text-align and whitespace don't work together nicely.

abeforgit:
Nobody wants this more than I do 
I'm excited to investigate conversion strategies, though I'm unsure where to place this in the list of fundamental changes we still need to tackle

abeforgit:
also consider the alignment issue mentioned in your link is from 2018, with any luck this has improved by now

nvdk:
> Nobody wants this more than I do I'm excited to investigate conversion strategies, though I'm unsure where to place this in the list of fundamental changes we still need to tackle

Indeed, I think we will have to improve/bugfix the nbsp handling short term, this is more a long term solution.

rfcs/whitespace-handling.md
```diff
@@ -0,0 +1,78 @@
...
+| prosemirror | use pre-wrap by default, plugins exist to use nbsp behaviour |
+| ckeditor 4 & 5 | use nbsp, seems this is mostly handled in the view |
+| tinymce | use pre-wrap |
+
+
+### old content
+Older documents, pasted content and our templates contain a lot of spurious whitespace which will likely be shown when we use this css property. We will need some kind of strategy to deal with that.
+
+### inconsistencies between layout in the editor and when rendered outside
+Both `pre-wrap` and `break-spaces` also break lines on newline characters, which is likely unwanted behaviour as these will not show up in exported documents unless they apply the same whitespace styling. So extra care needs to be taken to replace these newlines with proper breaks on export.
```
abeforgit, line 75: Likely we will go for inserting <br> elements as a linebreak strategy, so I think this is a non-issue?

abeforgit (commented):
from my limited experimentation: 
converting seems doable, but as you discovered we seem to lose some browser-default behavior
which hints at a better handling of dom-updates as a prerequisite for this

PR #211, opened 2022-02-08
Improve datastore interface

abeforgit:
Adds new node query methods which return a more workable response
Changes:
- as___NodeMapping methods return a TermMapping object
- the asQuadsResultSet method returns a ResultSet

Both of these remain iterable (with for...of), but both offer some convenienve methods over normal sets.
- `single()`: takes the only element and returns it. Errors if there are more elements in the iterator. Useful in situations where you know there can only be one result.
- `map(mappingFunc)`: normal "apply this func to each item in the iterator" stuff

TermMapping also has a `get(termSpec)` method, for when you've matched multiple terms, but you need the nodes for a specific term. Takes the same kind of arguments as `datastore.match` (so concise term syntax, or RDFjs term objects)

Another improvement is that the nodes per term are now an array instead of a set, and they are guaranteed to be in document order, and contain no duplicates. The ordering was a piece of information we already had, but was destroyed by using sets to ensure no duplicates.

Also adds a minor improvement to the RangeFactory, setting default offset values for the `fromInNode` method.

Lastly, splits up the datastore file into several files, it was starting to become a little too long since it contained 4 class definitions and a bunch of types.


abeforgit:
Made draft until its made non-breaking

lagartoverde (approved):
Looks good, I think maybe we will have to change some of the range strategies in the future but we will see that when using them what makes sense now it may not make sense in a few months. Good job :)

PR #213, opened 2022-02-10
Also recalculate datastore on model-read

abeforgit:
Datastore was only being updated on contentChange, but modelRead also changes the vdom state so needs another recalc


lagartoverde (approved):
thanks :)

PR #214, opened 2022-02-10
Expose query utility on markset

abeforgit:
Provides a mehod on a markset to search by mark name, now that the hashing function is name + origin.


lagartoverde (approved):
The code looks good, I will try to finish the plugin using this branch and tell you if I find any bugs

PR #216, opened 2022-02-11
Don't export the debug component

abeforgit:
Dont export the debug component so that apps that don't need it don't have to depend on heavy debug dependencies.

To use the debug component in a plugin, do the following:

- create an rdfa-editor-with-debug.js file in `/tests/dummy/app/components` with the contents:

```
export {default} from "@lblod/ember-rdfa-editor/components/rdfa/rdfa-editor-with-debug";
```

- include the following dependencies as devDependencies:
    - @codemirror/basic-setup
    - @codemirror/lang-html
    - @codemirror/lang-xml
    - xml-formatter

- use the debug component in your dummy templates as `<RdfaEditorWithDebug />`

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,10 +177,9 @@ This addon uses CSS variables to customise the styling. You can override these v
 
 
 ## Compatibility
-* Ember.js v3.12 or above
-* Ember CLI v2.13 or above
+* Ember.js v3.20 or above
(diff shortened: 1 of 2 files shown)


nvdk:
probably nice to include those instructions in the readme

nvdk (approved):

PR #217, opened 2022-02-11
update instructions on the debug component

nvdk:


README.md
```diff
@@ -220,7 +220,22 @@ export default {
...
+
+include the following dependencies as devDependencies:
+```
+        @codemirror/basic-setup
+        @codemirror/lang-html
+        @codemirror/lang-xml
+        xml-formatter
+```
+
+The debug component can then be added with `<Rdfa::RdfaEditorWithDebug>` to your dummy app templates. It should support the same variables as the `<Rdfa::RdfaEditor>`. The block content of the debug component is yielded at the top of the component, right before the debug features. E.g.:
```
abeforgit, line 238: this is wrong, it will be just `<RdfaEditorWithDebug />` since its not in the rdfa namespace anymore

abeforgit (changes requested):

benjay10 (approved):

PR #219, opened 2022-02-11
Implement text-matching command

abeforgit:
Provides a way to regex-search the text inside a range.
Handles blocknodes (treated as newlines for searching purposes) and invisible spaces (ignored).

This is a heavily modified port of the match-text command from the now-defunct plugin-rework branch, for those that remember that.

### DIFF
--- a/addon/model/model-range.ts
+++ b/addon/model/model-range.ts
@@ -354,7 +354,7 @@ export default class ModelRange {
     for (const node of walker.nodes()) {
       if (ModelNode.isModelText(node)) {
         // keep a sparse mapping of character indices in the resulting string to paths
-        const sanitizedContent = node.content;
+        const sanitizedContent = node.content.replace(/\u200B/g, '');
         if (calculateMapping) {
           const path = ModelPosition.fromBeforeNode(node).path;
           mapping.push([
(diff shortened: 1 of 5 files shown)


abeforgit:
closes https://binnenland.atlassian.net/browse/GN-3214

addon/model/model-range.ts
```diff
@@ -311,6 +311,138 @@ export default class ModelRange {
...
+    const startOffset = this.start.isInsideText()
+      ? this.start.parentOffset - this.start.nodeAfter()!.getOffset()
+      : 0;
+
+    let currentIndex = 0;
+    // build the resultstring
+    for (const node of walker.nodes()) {
+      if (ModelNode.isModelText(node)) {
+        // keep a sparse mapping of character indices in the resulting string to paths
+        const sanitizedContent = node.content.replace(/\u200B/g, '');
```
nvdk, line 357: I'd prefer the use of constants for things like the invisible space, makes it more clear what is is

abeforgit: yeah but you cant do that in a regex

nvdk:
```
const myRegex = new Regex(`${MYCONSTANT}`,'g');
node.content.replace(myRegex, '');
```
should work, no?

abeforgit: yes indeed, good idea!

PR #221, opened 2022-02-16
Explicit import of the debug component

benjay10:
Due to the export of the debug component in the editor, the consuming app also imports modules only related to the debug component for no reason. The solution was to not export the debug component. The plugins that want to use the debug component need to explicitly import it first. This PR does that. Explicit importing is needed for ember-rdfa-editor v0.50.0-beta.8 and onward, but the explicit import won't break on older editor versions.


abeforgit:
Sorry, I pulled a sneaky, this is already done
https://github.com/lblod/ember-rdfa-editor/pull/220/files

PR #223, opened 2022-02-22
Enhancement/whitespace handling

nvdk:
Very basic implementation of pre-wrap whitespace handling in the editor:

- assumes anything inserted via insert-html command needs to be converted (e.g. collapsing whitespace, removing `\n` etc).
- Keeps existing non breaking spaces for now
- exports convert in the other direction, so whitespace that would collapses is replaced with `nbsp`.

Created as a draft because:
- needs more testing
- copy should probably also do the conversion from pre to normal collapsing
- some assumptions need to be validated: 
    - currently assumes we use <br> for newlines and that the editor will not use `\n`
    - empty textnodes can be dropped without consequence


addon/commands/insert-html-command.ts
```diff
@@ -24,7 +24,7 @@ export default class InsertHtmlCommand extends Command {
     const parser = new DOMParser();
     const html = parser.parseFromString(htmlString, 'text/html');
     const bodyContent = html.body.childNodes;
-    const reader = new HtmlReader(this.model);
+    const reader = new HtmlReader(this.model, true);
```
nvdk, line 27: think I should create a `WhitespaceCollapsingHtmlReader that extends `HtmlReader` instead

lagartoverde (approved):
Code looks good and it works as far as I can tell in the dummy app, some minor comments about commented code

nvdk:
looks to be working ok now, did not spot any immediate bugs. Needs some testing in an app with a plugin I think

abeforgit:
Targeting this for 0.51.0

... that feels nice to say, should we maybe incorporate that into the planning? jira has a concept of planning for releases :thinking:

nvdk:
> Targeting this for 0.51.0
> 
> ... that feels nice to say, should we maybe incorporate that into the planning? jira has a concept of planning for releases thinking

lets  keep the amount of administration minimal :pray:

PR #231, opened 2022-03-02
debounce text input slightly

nvdk:
By debouncing the text input we give chrome a bit more time to do it's rendering, it feels a lot better in chrome and the impact is minimal in firefox.


abeforgit:
nice trick, goes from unusable to frustrating but usable

PR #232, opened 2022-03-03
Breaking/remove old plugin wiring

nvdk:
This drops all old plugin API's and rendering, but keeps the code related to setting cursor positions etc for handlers. Also updates the readme to reflect the new API's.

Essentially the following is removed:
 - the event processor
 - the hintstregistry and all calls related to it
 - the plugin dispatcher
 - plugin api's such as the pernet API's
 - the text representation of the document and the generateDiffEvents calculation using it


abeforgit:
Putting this on hold until citation plugin is fixed

abeforgit:
Actually having a look at this again
any reason for keeping things like the fallback input handler and the pernet-raw-editor class?

nvdk:
Indeed older input handlers require pernet for setting the caret at least, I've kept only the functionality that is still required. Wrt to the fallback plugin handler, I didn't check if that could be dropped. I think it's likely we can drop it, but this needs to be verified.

nvdk:
I think this can be merged now, since all plugins have moved over to the new api's.

PR #233, opened 2022-03-03
Fix bug with predicate node generation

abeforgit:
Predicate node querying was broken because of an incorrect way to filter out subjects


addon/model/util/datastore/datastore.ts
```diff
@@ -470,6 +470,16 @@ export class EditorStore implements Datastore {
     return subjects;
   }
 
+  private getSubjectForNode(node: ModelNode) {
```
nvdk, line 473: a comment block detailing what this is supposed to do would be nice.

lagartoverde (approved):
Looks good, I'm also in favor of the comment :)

PR #234, opened 2022-03-03
Bug/mark attribute rendering

abeforgit:
fix mark rendering crashing when an attribute had `undefined` as a value

### DIFF
--- a/tests/unit/commands/match-text-command-test.ts
+++ b/tests/unit/commands/match-text-command-test.ts
@@ -184,4 +184,42 @@ module('Unit | commands | match-text-command-text', (hooks) => {
 
     assert.true(results[0].range.sameAs(expectedRange));
   });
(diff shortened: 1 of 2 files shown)


nvdk (approved):
don't know when this occurs, but the code looks sane.

tests/unit/commands/match-text-command-test.ts
```diff
@@ -184,4 +184,40 @@ module('Unit | commands | match-text-command-text', (hooks) => {
 
     assert.true(results[0].range.sameAs(expectedRange));
   });
+  test('only match inside of searchRange', (assert) => {
```
nvdk, line 187: this test doens't really belong in this PR I think

lagartoverde (approved):
Looks good :)

PR #235, opened 2022-03-04
Fix list indentation

Dietr:
In some cases (GN: intro & outro text) the list indentation was not displayed correctly.

### DIFF
--- a/app/styles/ember-rdfa-editor/_c-editor.scss
+++ b/app/styles/ember-rdfa-editor/_c-editor.scss
@@ -21,16 +21,12 @@ $say-annotation-width: $au-unit-huge * 1.5 !default;
 
 .say-editor {
   background-color: $say-paper-background;
-  padding: $say-editor-padding $say-editor-padding * .5;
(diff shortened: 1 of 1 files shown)


benjay10:
My apologies for reviewing, even when I'm not assigned.

The bug sure is fixed, bullet points follow logical indentation now, but I still have some remarks and not sure if this the correct time to solve them.

The bullet points seem to be negatively indented, they appear to much to the left (unless that is intentional). It looks a bit wonky, especially in overview components.
afbeelding
afbeelding

There is also no outlining of a virtual sheet of paper, in contrast to other parts of the application that use an editor. This is no big deal, but before these changes the text appeared all the way on the left of the screen, while now, with corrected padding, the text seems to float more in the center of the screen, which might be a bit confusing. It is as if there are irremovable spaces or tabs on the left.

Before this PR:
afbeelding

After this PR:
afbeelding

nvdk:
I think it's best to log these as separate bugs on the backlog, so we can already merge this fix.

PR #236, opened 2022-03-13
Implement incremental dom writing

abeforgit:
PLEASE TEST A LOT

This is a complete rewrite of the reading/writing logic, both of the content _and_ the selection.

### How it works:

Every node now keeps track of its own 'dirty' state. This is a set of strings with 3 possible values:
- "node": something in the node itself has changed (in practice, mainly attribute changes, or when it's a new node)
- "content": for elements, a change has been made to it's children array (aka children added or removed, changes in children are handled by a "node" dirtyness on the child). For textnodes, its textcontent has changed
- "mark": only relevant for textnodes, a change in its marks (mainly future proofing, has the same effect as a "node" dirtiness at the moment, but that's a coincidence). 

(Recall, it's a set, so nodes can have any combination of these active at the same time)

#### assumptions:
changes to nodes are done using their public methods. This is not 100% enforced, for example the `children` array is publicly readable, and because this is JS it is therefore also mutable. Though, directly messing with these kinds of things has been undefined behavior for a while already, and usage of the node API is only permitted within Operation logic, so unlikely to have any impact.

The html-writer then smartly updates the dom based on these dirtiness values. 
**This means that the writer is no longer a pure functional interface!**
Having the writer make instant changes to the dom avoids the need to traverse twice.

#### NodeViews

While working on the writing logic, the way that Marks were handled became a gigantic nuisance to debug. After some fruitless prodding, I decided to rework it. This has led to the introduction of the concept of a `NodeView`. 
Previously, ModelNodes were tied to a single DOMNode, called the `boundNode`. This coupling was necessary for converting selections back and forth. 
With marks however, a new problem was introduced: a textnode could now be rendered as a small tree of elements, instead of a plain Text node. With incremental writing, it became necessary to couple the ModelText node both to the dom Text node, as well as the root of the subtree defined by its marks.

NodeViews are the generic form of this coupling. At the moment it is simply an object with the following interface:
```ts
interface NodeView {
  viewRoot: Node,
  contentRoot: Node
}
```

The viewRoot represents the root of the DOM subtree defined by this modelNode. For a textnode, this is the root subtree created by its marks. This is what was previously called the `boundNode`
The contentRoot is then where the _content_ of the node (literally `content` for textnodes, and `children` for elements) will be rendered.

Currently, viewRoot and contentRoot are identical for ModelElements and only differ for ModelText nodes _if_ they have Marks.
But it's easy to see the value of this separation when we think about non-trivial rendering for special elements.

This change had consequences for all readers and writers, and especially for the selection reader/writer. Luckily, it was actually a beneficial change. Rather than the awkward heuristics of `findPositionForTextPropertyNode` (a relic from the pre-marks era) we now have a clear distinction between selections "in the content root" and "outside the content root". The former being a selection we can meaningfully transform, and the latter simply mapping to either before or after the modelnode, based on some heuristics.

Note, as usual when it appears I've done something smart, this is heavily inspired by ProseMirror.

Another detail: rather than storing the nodeView as a property on the node itself, I've decided to maintain a two-way WeakMap in the Model. Since at least a one-way WeakMap had to be maintained already, this was a much cleaner solution. 
It also makes the Nodes _almost_ a **true** model in the MVC sense. The only remnant of view-specific logic (and it's debatable whether it is view-specific) is the dirtiness flags. It's likely those will move away from nodes when immutable nodes are introduced, since at that point any node-update logic will live outside the node as well, leading to an even cleaner MVC structure. This is not a short-term goal though.

#### Other

- cloning of a markset was broken. (This will also be split off as a separate PR since it's an urgent fix)
- the debouncing hack in the TextInputHandler is no longer necessary since performance is vastly improved on chrome, so it is removed.

Closes https://binnenland.atlassian.net/browse/GN-3304

### DIFF
--- a/addon/commands/insert-html-command.ts
+++ b/addon/commands/insert-html-command.ts
@@ -35,6 +35,7 @@ export default class InsertHtmlCommand extends Command {
           modelNodes.push(...parsed);
         }
       });
+      modelNodes.forEach((node) => console.log(node.toXml()));
 
       const newRange = mutator.insertNodes(range, ...modelNodes);
       this.model.selectRange(newRange);
--- a/addon/commands/insert-html-command.ts
+++ b/addon/commands/insert-html-command.ts
@@ -35,7 +35,6 @@ export default class InsertHtmlCommand extends Command {
           modelNodes.push(...parsed);
         }
       });
-      modelNodes.forEach((node) => console.log(node.toXml()));
 
       const newRange = mutator.insertNodes(range, ...modelNodes);
       this.model.selectRange(newRange);
--- a/addon/editor/input-handlers/backspace-handler.ts
+++ b/addon/editor/input-handlers/backspace-handler.ts
@@ -388,6 +388,7 @@ export default class BackspaceHandler extends InputHandler {
 
     void this.backspace().then(() => {
       this.rawEditor.updateSelectionAfterComplexInput(); // make sure currentSelection of editor is up to date with actual cursor position
+      this.rawEditor.model.read();
       this.rawEditor.eventBus.emit(
         new ContentChangedEvent({
           owner: CORE_OWNER,
@@ -397,7 +398,6 @@ export default class BackspaceHandler extends InputHandler {
           },
         })
       );
-      this.rawEditor.model.read();
     });
 
     return { allowPropagation: false };
--- a/addon/editor/input-handlers/backspace-handler.ts
+++ b/addon/editor/input-handlers/backspace-handler.ts
@@ -397,6 +397,7 @@ export default class BackspaceHandler extends InputHandler {
           },
         })
       );
+      this.rawEditor.model.read();
     });
 
     return { allowPropagation: false };
--- a/addon/editor/input-handlers/disable-delete-handler.ts
+++ b/addon/editor/input-handlers/disable-delete-handler.ts
@@ -13,7 +13,6 @@ export default class DisableDeleteHandler extends InputHandler {
   }
 
   handleEvent(_: KeyboardEvent): HandlerResponse {
-    this.rawEditor.model.read();
     return { allowPropagation: false, allowBrowserDefault: false };
   }
 }
--- a/addon/editor/input-handlers/disable-delete-handler.ts
+++ b/addon/editor/input-handlers/disable-delete-handler.ts
@@ -13,6 +13,7 @@ export default class DisableDeleteHandler extends InputHandler {
   }
 
   handleEvent(_: KeyboardEvent): HandlerResponse {
+    this.rawEditor.model.read();
     return { allowPropagation: false, allowBrowserDefault: false };
   }
 }
--- a/addon/editor/input-handlers/text-input-handler.ts
+++ b/addon/editor/input-handlers/text-input-handler.ts
@@ -10,7 +10,6 @@ import {
 } from '@lblod/ember-rdfa-editor/utils/errors';
 import PernetRawEditor from '@lblod/ember-rdfa-editor/utils/ce/pernet-raw-editor';
 import { isKeyDownEvent } from '@lblod/ember-rdfa-editor/editor/input-handlers/event-helpers';
-import { restartableTask, TaskGenerator, timeout } from 'ember-concurrency';
 
 export type TextHandlerManipulation = InsertTextIntoRange;
 
@@ -37,7 +36,6 @@ export interface TextInputPlugin extends InputPlugin {
  */
 export default class TextInputHandler extends InputHandler {
   plugins: Array<TextInputPlugin>;
-  capturedText = '';
 
   constructor({ rawEditor }: { rawEditor: PernetRawEditor }) {
     super(rawEditor);
@@ -90,16 +88,8 @@ export default class TextInputHandler extends InputHandler {
     return { allowPropagation: false };
   }
 
-  @restartableTask
-  *inserText(): TaskGenerator<void> {
-    yield timeout(50);
-    this.rawEditor.executeCommand('insert-text', this.capturedText);
-    this.capturedText = '';
-  }
-
   handleNativeManipulation(manipulation: TextHandlerManipulation) {
     if (manipulation.type === 'insertTextIntoRange') {
-      // eslint-disable-next-line @typescript-eslint/unbound-method
       this.rawEditor.executeCommand('insert-text', manipulation.text);
     } else {
       throw new UnsupportedManipulationError(manipulation);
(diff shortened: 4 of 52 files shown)


abeforgit:
Note: this is _theoretically_ not a breaking change. 

But, you know...

abeforgit:
Also note when evaluating performance: A simple mouseclick in the document still causes a full model read, which completely rebuilds the entire document, you'll notice a significant lag when typing the first character. I expect this will be cleaned up in the near future.

addon/commands/insert-html-command.ts
```diff
@@ -35,6 +35,7 @@ export default class InsertHtmlCommand extends Command {
           modelNodes.push(...parsed);
         }
       });
+      modelNodes.forEach((node) => console.log(node.toXml()));
```
lagartoverde, line 38: You left this console.log here :)

addon/model/model.ts
```diff
@@ -217,15 +198,12 @@ export default class Model {
...
 
     if (writeBack) {
-      if (subTree) {
-        this.write(subTree);
-      } else {
-        this.write(this.rootModelNode);
-      }
+      this.write();
     }
+    console.log(this.toXml());
```
lagartoverde, line 206: Another console.log

addon/model/operations/operation-algorithms.ts
```diff
@@ -50,16 +79,16 @@ export default class OperationAlgorithms {
...
-      range.end.split(true);
-      const before = range.start.nodeBefore();
-      const after = range.end.nodeAfter();
-      if (before) {
-        _markCheckNodes.push(before);
-      }
-      if (after) {
-        _markCheckNodes.push(after);
-      }
+      // range.start.split();
```
lagartoverde, line 82: Try to not leave commented out code unless is necessary

addon/model/readers/html-node-reader.ts
```diff
@@ -62,6 +64,7 @@ export default class HtmlNodeReader
     } else {
       result = [];
     }
+    // result.forEach(node => node.clearDirty())
```
lagartoverde, line 67: Commented code :)

lagartoverde (commented):
Code looks good as always, leave some small comments about things to clean.
Will link my local to this branch of the editor while working on the plugins so I can test and detect bugs, I will approve in a few days if I don't find any. Good job 👍

nvdk:
:warning:  Looks like there's some situations where we're failing to write a selection. Looking into whether it's related to this PR, will update with more infore

addon/editor/input-handlers/backspace-handler.ts
```diff
@@ -397,6 +397,7 @@ export default class BackspaceHandler extends InputHandler {
           },
         })
       );
+      this.rawEditor.model.read();
```
nvdk, line 400: shouldn't this be before emitting the content changed event?

addon/editor/input-handlers/disable-delete-handler.ts
```diff
@@ -13,6 +13,7 @@ export default class DisableDeleteHandler extends InputHandler {
   }
 
   handleEvent(_: KeyboardEvent): HandlerResponse {
+    this.rawEditor.model.read();
```
nvdk, line 16: this shouldn't be necessary, nothing is happening here?

addon/editor/input-handlers/text-input-handler.ts
```diff
@@ -100,9 +99,8 @@ export default class TextInputHandler extends InputHandler {
 
   handleNativeManipulation(manipulation: TextHandlerManipulation) {
     if (manipulation.type === 'insertTextIntoRange') {
-      this.capturedText = `${this.capturedText}${manipulation.text}`;
```
nvdk, line 103: also remove the capturedText field on the class

addon/editor/input-handlers/text-input-handler.ts
```diff
@@ -100,9 +99,8 @@ export default class TextInputHandler extends InputHandler {
...
       // eslint-disable-next-line @typescript-eslint/unbound-method
-      void taskFor(this.inserText).perform();
```
nvdk, line 105: insertText task can be removed as well

addon/model/model-text.ts
```diff
@@ -25,6 +27,7 @@ export default class ModelText extends ModelNode {
 
   set content(value: string) {
     this._content = value;
+    this.addDirty('content');
```
nvdk, line 30: shouldn't a node be dirty on create as well?

abeforgit: look at the constructor of the superclass :)

addon/editor/input-handlers/undo-handler.ts
```diff
@@ -17,8 +17,7 @@ export default class UndoHandler extends InputHandler {
   }
 
   handleEvent(_: KeyboardEvent): HandlerResponse {
-    this.rawEditor.undo();
-
+    this.rawEditor.model.read();
```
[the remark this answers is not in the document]
abeforgit: simply a mistake, that line shouldnt have been deleted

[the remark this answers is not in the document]
nvdk: looking into it further, I think there's an unrelated bug here. The browser default should not be allowed here. this was introduced in https://github.com/lblod/ember-rdfa-editor/commit/aa7b1e24f060be6e1feacc5ec1a555506e4156f7#diff-e70efc7d6e1f4bda63edb7a4ef88b4fb4381db2b519f1f8161761ac3f31802f3R16

addon/editor/input-handlers/backspace-handler.ts
abeforgit: indeed it should

addon/editor/input-handlers/disable-delete-handler.ts
abeforgit: correct, wasn't paying enough attention

addon/model/readers/selection-reader.ts
```diff
@@ -84,72 +84,54 @@ export default class SelectionReader
...
+      return ModelPosition.fromInTextNode(modelNode, domOffset);
+    } else if (isElement(container) && ModelNode.isModelElement(modelNode)) {
+      if (modelNode.children.length) {
+        return ModelPosition.fromInElement(
+          modelNode,
+          modelNode.indexToOffset(domOffset)
+        );
+      } else {
+        if (domOffset === 0) {
+          return ModelPosition.fromInElement(modelNode, 0);
```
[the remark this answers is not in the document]
abeforgit: it seems pretty much impossible to arrive at this scenario (even forcing it through the browser console doesn't get you there)
So I think defaulting to setting the cursor in front of the element is probably a good heuristic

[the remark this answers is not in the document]
abeforgit: ill also add a check in case the element is root

nvdk (approved):
I think this can be merged

PR #237, opened 2022-03-21
Fixing unindenting

benjay10:
Unindenting is better now, the least amount of DOM operations (hopefully).
Before, unindenting a list item would make some previous list items break on multiple lines. This is because empty `<ul>`'s were inserted in the text of those previous list items.

Also looking at indenting as well, as that is also not always correct anymore. This PR remains a draft until that is also fixed or if no better solution could be found in reasonable time.


benjay10:
Something I'm uncertain about, is that sometimes the `mutator` is used to insert and delete nodes, and sometimes, the `element.addChild` and related methods are used. Being new to the editor, I hope these can be interchanged.

Using the mutator was actually part of the cause of the bug, as a position to insert is searched, then the DOM was changed, and then the insertion was done, but in the wrong place.

benjay10:
Okay, no idea what happened with this PR. After many rebase attempts and failures, all commit where lost all of a sudden, and the PR closed automatically. Was able to recover everything with some black magic and reopen this PR.

nvdk:
@benjay10 based on feedback from @abeforgit I understand the mutator must be used for making changes to our VDOM.

abeforgit (approved):
seems to work much better and code looks good, well done!

PR #239, opened 2022-03-23
execute undo on VDOM

nvdk:
while reviewing #236 I noticed the undo handler is not preventing default browser behaviour at the moment, this fixes that. 
edit: Turns out fixing that involved porting this to work on the VDOM, so updated the title to reflect that
NEEDS EXTENSIVE TESTING

### DIFF
--- a/addon/commands/insert-xml-command.ts
+++ b/addon/commands/insert-xml-command.ts
@@ -4,8 +4,7 @@ import { parseXmlSiblings } from '@lblod/ember-rdfa-editor/model/util/xml-utils'
 import Model from '@lblod/ember-rdfa-editor/model/model';
 import { logExecute } from '@lblod/ember-rdfa-editor/utils/logging-utils';
 import ModelRange from '@lblod/ember-rdfa-editor/model/model-range';
-import ModelNode from '../model/model-node';
-import ModelElement from '../model/model-element';
+import setNodeAndChildDirty from '@lblod/ember-rdfa-editor/model/util/set-node-and-child-dirty';
 
 /**
  * Allows you to insert model nodes from an xml string.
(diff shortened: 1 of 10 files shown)


nvdk:
looks like cursor handling is broken now

nvdk:
Turned out it was easier to get this working on the VDOM than getting the old version working. Still some (minor?) issues to work out, looks like I'm missing a few edits in history still at the moment.

addon/commands/insert-xml-command.ts
```diff
@@ -4,8 +4,7 @@ import { parseXmlSiblings } from '@lblod/ember-rdfa-editor/model/util/xml-utils'
 import Model from '@lblod/ember-rdfa-editor/model/model';
 import { logExecute } from '@lblod/ember-rdfa-editor/utils/logging-utils';
 import ModelRange from '@lblod/ember-rdfa-editor/model/model-range';
-import ModelNode from '../model/model-node';
-import ModelElement from '../model/model-element';
+import setNodeAndChildDirty from '@lblod/ember-rdfa-editor/model/util/set-node-and-child-dirty';
```
nvdk, line 7: moved to a helper, because it turns out restoring a snapshot is also a special case

addon/model/model.ts
```diff
@@ -248,11 +252,23 @@ export default class Model {
...
+          setNodeAndChildDirty(snapshot.rootModelNode);
+          mutator.insertNodes(range, ...snapshot.rootModelNode.children);
+        },
+        false,
+        false
+      );
       if (writeBack) {
-        this.write();
+        this.write(this.rootModelNode, false);
+        this._selection = snapshot.modelSelection.clone(this.rootModelNode);
```
nvdk, line 270: didn't seem to work without this clone, though I don't really get why...

addon/model/model.ts
```diff
@@ -195,11 +196,14 @@ export default class Model {
...
   change(
     callback: (mutator: ImmediateModelMutator) => ModelElement | void,
-    writeBack = true
+    writeBack = true,
+    saveSnapshot = true
   ) {
     const mutator = new ImmediateModelMutator(this._eventBus);
     callback(mutator);
-
+    if (saveSnapshot) {
```
nvdk, line 204: not sure if it's best before or after writing the tree (or if it even matters at all)

lagartoverde:
https://user-images.githubusercontent.com/17006446/159938257-4e88a897-58b6-483e-b9e9-d5efaf15d4bb.mov

So weird things happening with the undo, when I remove selecting a portion of the text I cannot undo it, and also when I'm undoing I can't undo the first character I deleted

lagartoverde:
The problem with undoing the first action also happens when inserting

nvdk:
Think I resolved that now @lagartoverde, I was creating a snapshot after a change was applied but I should have been creating it before as we do now

abeforgit (commented):
Works well, some issues with non-collapsed selections:
- select range of text
- backspace
- undo
resulting selection is somewhere unexpected
Not really blocking though

PR #240, opened 2022-03-23
add embroider test scenarios to ember try

nvdk:
According to the embroider readme:
> The best thing for all addons authors to do right now is to achieve the "Embroider Safe" support level. Follow the instructions in the @embroider/test-setup README to add the embroider-safe scenario to your ember-try config.


nvdk:
main things I've noticed:
- embroider doesn't seem to like the way ix does it's exports/imports
- some issues with how we set up crypt

PR #245, opened 2022-03-30
replace all special spaces when regular spaces when parsing html

nvdk:


nvdk:
created as a draft because I want to add some tests + I'm wondering if this shouldn't be part of the html reader instead

nvdk:
This behaviour I think can be generalised for both text and HTML inserts

abeforgit:
I'll leave that discussion for https://binnenland.atlassian.net/browse/GN-3369

It's unclear how harsh we should be in this conversion: e.g. if we do it in the operations, that means it becomes impossible to insert nbsps into the VDOM for any reason whatsoever

On the surface that sounds nice, but until we are fully vdom-based I'm not entirely sure of the consequences yet, so it deserves its own story still

PR #246, opened 2022-03-31
Implement self-updating regex-constrained sets of marks

abeforgit:
This PR introduces a new core feature: LiveMarkSet

## LiveMarkSet
Its a set of marks that trigger on a certain search pattern and automatically update themselves. You provide them with a list of specifications (annoyingly, they are not _quite_ the same as MarkSpecs, but are similar) for marks, along with a query for the datastore.
Upon every contentchange, the livemark will set the specified marks according to the query. It avoids writing to the real dom unnecessarily, and is therefore fast enough to not interrupt the typing experience.

For the query, you have the entire datastore api at your disposal, with the caveat that the query has to result in a TextMatch object. The intended way to get this object is to use the new `searchText` method on the datastore, which will search for text in the same way that the `match-text-command` does, but limited to the relevant ranges defined by the filtered datastore.

**plugins can create a LiveMarkSet with the `createLiveMarkSet` method available on the controller**

### LiveMarkSet events

LiveMarks have their own private eventBus, and will fire a LiveMarkSetUpdatedEvent whenever they update.
You can listen to it using the familiar `onEvent` and `offEvent` methods on the livemarkset instance

## Other new things

To support the LiveMarkSet, quite significant changes were made and new features added.

### RangeMapper

**WARNING** considered functional, but experimental. In particular, the `bias` setting is not clearly defined and needs some thinking

The idea here is to enable the ability to translate a range (or position) _before_ an operation to an equivalent one _after_ the operation, accounting for shifts in positions due to deletion or addition of content.
Every operation is required to return such a mapping.
Crucially, the mappers are combineable. Simply do `mapper.appendMapper(otherMapper)` to allow for valid translations across multiple operations in sequence.
The mutator makes use of this to collect all mappers it receives from operations, and provides a `mapRange` method of its own which runs through all the transformations in order.

Where possible, the return value of the mutator methods was kept the same, this is still the "default sensible range" from before.

### InsertTextOperation

After quite a few reworks, I decided to make text input a first class citizen and give it its own operation. This allowed for more efficient text insertion, since inside an operation all gloves are off and nodes can be directly edited, allowing us to do things like only changing the content instead of replacing entire nodes, and to be smart about merging textnodes if they are adjacent to other textnodes that have similar attributes and marks.

Note that this now creates a difference between `mutator.insertText("text")` and `mutator.insertNodes(new ModelText("text"))`
the latter being the "dumb" version that purely inserts the node without special handling. To be considered if this should be made smarter.

### RemoveMarksFromRanges command

New command to remove marks from a batch of ranges at once, avoiding costly write loops.

### ModelRange.contextNodes

The strategies were expanded. 'rangeTouches' now supports the same options as 'isInside', and every strategy also supports a filter function. No (intended) changes to existing api.

### GenTreeWalker

- now supports the `visitParentsUpwards` option, which "visits closing tags as well as opening tags". This can result in some elements being visited twice, this is by design. The option is off by default.

### Controller

- added the `createLiveMarkSet` method
- added the `getMutator` method to obtain a raw instance of a mutator. This is a lower-level access than `model.change`, in the sense that it doesn't automatically write back to the DOM. 
- added the `write` method to allow manual writing from the controller

### SetUtils & MapUtils
- added a `hasAny` method to both, allowing to check if a set or map has any of the provided elements.

### insert-new-li
- Fixed the selection issues by making sure the new li has some content so the selection can render inside of it (inserts an invisible space if needed)
- Noticed the logic was doing unnecessary splits, resulting in the new li sometimes ending up with an empty element it didn't need.

### insert-text
There were actually quite a few problems with the logic, hence the decision to break it out into its own operation.
Complete rewrite, and it's much simpler now.

### GenTreeWalker

- Fixed some issues with a range-constrained walker. The logic for searching the endNode was wrong and is now (hopefully) better. To be monitored, this is like the sixth time I'm doing this, so I don't trust myself anymore.

## Things noticed but not fixed:

- handling of inserting newlines into a position where they are the last child of their parent is iffy. The content and modelselection is correct but the browser renders it badly which will confuse users, but out of scope


addon/model/controller.ts
```diff
@@ -126,10 +139,22 @@ export class RawEditorController implements Controller {
     return this._rawEditor.rootModelNode;
   }
 
+  get marksRegistry(): MarksRegistry {
+    return this._rawEditor.model.marksRegistry;
+  }
+
+  getMutator(): ImmediateModelMutator {
```
nvdk, line 146: is this now preferred over `model.change( (mutator) => ... )` if you have a controller?

abeforgit: not yet

addon/model/live-mark-set.ts
```diff
@@ -0,0 +1,139 @@
...
+  get liveMarkSpecs(): LiveMarkSpec[] {
+    return this._liveMarkSpecs;
+  }
+
+  private update = () => {
+    const { matchesToAdd, matchesToRemove } = this.calculateRanges();
+    const mutator = this.controller.getMutator();
+    const marksRegistry = this.controller.marksRegistry;
+
+    console.log('TOADD', matchesToAdd);
```
nvdk, line 65: probably want to use a logger here instead of console.log, or remove them alltogether
```suggestion

```

addon/model/live-mark-set.ts
```diff
@@ -0,0 +1,139 @@
...
+    console.log('TOREMOVE', matchesToRemove);
```
nvdk, line 66:
```suggestion
```

addon/model/model-range.ts
```diff
@@ -433,9 +461,57 @@ export default class ModelRange {
   }
 
   sameAs(other: ModelRange): boolean {
+    console.log('THISROOT', this.root);
```
nvdk, line 464:
```suggestion

```

addon/model/model-range.ts
```diff
@@ -433,9 +461,57 @@ export default class ModelRange {
...
+    console.log('OTHERROOT', other.root);
```
nvdk, line 465:
```suggestion
```

addon/model/controller.ts
abeforgit: but I want to provide a controller instance to commands as well, and it will replace the model instance
the change method will probably move to the controller in that case, maybe renamed to something more clear

addon/model/model-range.ts
```diff
@@ -433,9 +461,57 @@ export default class ModelRange {
...
     return this.start.sameAs(other.start) && this.end.sameAs(other.end);
   }
 
+  compare(other: ModelRange): RangeComparison {
+    const startCompStart = this.start.compare(other.start);
```
nvdk, line 468: It took me a while to understand waht startCompStart meant, didn't immediately find a better name though

abeforgit: yeah the non-abbreviated version is really quite verbose

abeforgit: maybe startStartDiff?

tests/unit/model/model-range-test.ts
```diff
@@ -367,4 +369,34 @@ module('Unit | model | model-range', () => {
       });
     });
   });
+  module('Unit | model | model-range | compare', () => {
```
nvdk, line 372: more tests on the range comparison might be warranted

abeforgit: yeah, good point
I'll evaluate if its still needed, I don't think I ended up using it much so I might just get rid of it

addon/model/model-range.ts
```diff
@@ -433,9 +461,57 @@ export default class ModelRange {
   }
 
   sameAs(other: ModelRange): boolean {
+    console.log('THISROOT', this.root);
+    console.log('OTHERROOT', other.root);
     return this.start.sameAs(other.start) && this.end.sameAs(other.end);
   }
 
+  compare(other: ModelRange): RangeComparison {
```
nvdk, line 467: range comparisons are tough, but I couldn't spot an oversight. well done :)

abeforgit: Unfortunately I see now that the only usage is in its only test... :joy: 
I ended up not needing range comparisons at all

abeforgit: the problem is that there are just so many cases that a caller will probably implement its own logic anyway based on the case it needs

abeforgit: you either make assumptions about which cases are equivalent or you end up so many cases on the callsite that your logic is pretty much as complex as if you were to compare positions yourself

addon/model/operations/insert-text-operation.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    this.emit(
+      new ContentChangedEvent({
+        owner: CORE_OWNER,
+        payload: {
+          type: 'insert',
+          oldRange: this.range,
+          newRange: defaultRange,
+          insertedNodes: [newText],
```
nvdk, line 65: if I understand correctly, when the contents was merged, this should be overwrittenNodes

tests/unit/commands/insert-new-li-command-test.ts
```diff
@@ -113,6 +113,7 @@ module('Unit | commands | insert-new-li-command-test', (hooks) => {
             <text>abc</text>
           </li>
           <li>
+            <text>${INVISIBLE_SPACE}</text>
```
nvdk, line 116: all bow to the mighty invisible space. I think we might want to write an RFC on how to do this properly...

addon/model/controller.ts
nvdk: might want to write that down somewhere ;)

tests/unit/commands/insert-new-li-command-test.ts
abeforgit: yeah... its annoying
the easiest place to clean up is on write, but I really want to avoid making changes to the vdom while writing

addon/model/operations/insert-text-operation.ts
abeforgit: honeslty this whole event is a bit in shambles, but I'll have a look

addon/model/controller.ts
abeforgit: yeah im planning to do a writeup of where we stand tomorrow

tests/unit/commands/insert-new-li-command-test.ts
nvdk: we may not need to have the invisible space in the vdom at all is what I'm thinking... it's only relevant when writing a selection (I think). anyway not a discussion we should have on this PR :).

abeforgit:
This is breaking because of potential slight deviations in selection position after certain commands (which, unfortunately, always have the potential to blow up) and a slightly different (but more accurate) result of the `rangeTouches` contextNodes strategy

Existing apis were not changed (or at least not intentionally)

tests/unit/commands/insert-new-li-command-test.ts
abeforgit: ooh yeah good idea

nvdk:
thanks a lot for the thorough description that helps a lot in reviewing this :+1:

PR #247, opened 2022-03-31
convert newlines to br elements when inserting text

nvdk:
since text inserts can contain newlines and we don't expect those in the editor, we convert them to br elements in the insert-text command. this also fixes some issues I had with pasting content from PDFs and editing that.


abeforgit (approved):
Looks good
Some points, but not really for this pr:
- we should do the same logic in the insert-html command
- this logic should move to the operation(s)

Ill make a ticket for this
EDIT: https://binnenland.atlassian.net/browse/GN-3369

nvdk:
> we should do the same logic in the insert-html command
I dont think we want that, the expectations are different when pasting/inserting text vs HTML. In HTML the default expectation is that newlines are not visible.

nvdk:
So to conclude the expected behaviour when inserting HTML vs text is different and as such the code can not be generalised.

abeforgit:
Good point!

PR #248, opened 2022-04-05
Add documentation generator

abeforgit:
Uses [Typedoc](https://typedoc.org/) to generate a documentation site based on jsdoc comments and typescript types.
To update the docs, run `npm run docs`.
It's a bit of an experiment, and an incentive to be more consistent about adding doc comments. 

The separation into modules is a bit arbitrary atm. Also, you'll often see `unexported`, these are things that are not explicitly exported from anywhere, such as native DOM api classes, and non-exported classes and interfaces from our own code.

We should probably set up some sort of hook to automatically regenerate the docs, though it does take about 15s each time so we should be careful to make it not too annoying for development. Maybe a githook on push?


nvdk:
perhaps we can set it up in the build so that the hosted app has it available?

abeforgit:
postponed
I think we will be able to generate better output in TEDI

PR #249, opened 2022-04-05
Improved table insert, and column and row insert

benjay10:
This PR adds an way to insert a table with specific dimensions, and provides 2 buttons on the toolbar for adding columns and rows for much faster editing.

**All features in this PR**:

- Toolbargroups with separators between categories of functionality (text formatting, lists and indentation, and table buttons in separate groups). Not essential to table insertion, but maybe good idea for clarity when also table buttons are to be shown.
afbeelding
- Added inputs for the dimensions of the table to insert.
afbeelding
- Added "Add row" and "Add column" buttons on the toolbar for faster table editing.
- Made icons (POC) for all table actions. This was needed when I wanted to make toolbar buttons for these actions.
afbeelding
- All buttons now have tooltips for clarity (using the `title` HTML attribute), with proper translations added.
afbeelding

**NEED TO FIX**:

- [x] Width of the dropdown. Some text is cut off, and I couldn't find why. (Edit: fixed by setting the size of the icons based on height instead of max-width. This also fixes some other size problems of icons.)
- [x] Icons need a review. I quickly made these, although I really tried to make them look official and good quality. Some might not clear at small sizes. (E.g. the plus symbol.)
- [x] Icons need to be embedded in the HTML as proper SVG's? Or not? They certainly need to be optimised first (created using Inkscape, and that leaves a lot of metadata).
- [x] Do the icons need to be added to Appuniversum? The others as well (list indentation, ...)? Because now, the inline SVG's are used.


benjay10:
@Dietr, we will need your finishing touch on the lay-out of the dropdown, the icons and the approval of the design.

Dietr (approved):

abeforgit:
Nice work, I expect we'll get a request to move the other table icons out of the menu as well, atm it doesn't make much sense to only have 2 of the icons available on single click

PR #252, opened 2022-04-08
Fix space-eating issues

abeforgit:
The normalToPreWrapWhitespace function's purpose is to make the html you insert look the same as it looks outside the editor (aka in an environment with normal whitespace handling). This means deleting all whitespace that wouldn't be rendered elsewhere.
However this was done too often: when reading from our own document, (which is already in prewrap-mode) this would delete spaces that were really part of the content. 
Fixed by not converting when we are reading from our own document.


lagartoverde (approved):
Code looks good, and didn't went super deep into testing but seems to work

PR #253, opened 2022-04-11
More consice removing of RDFa type

benjay10:
This version of the remove-type command does not create extra spaces on removing a type from the attribute of a DOM element. This is linked to the problems we found in https://github.com/lblod/besluit-publicatie-publish-service/pull/8 where extra spaces in RDFa attributes (although legal) caused the RDFa parser to create triples with empty subjects.


nvdk (approved):
that's a lot more readable as well :+1:

lagartoverde (approved):
Good job

PR #254, opened 2022-04-12
Fixed weird cases where the unindent button appeared without being available

lagartoverde:
Fixed the most straightforward case and the one the [Jira issue](https://binnenland.atlassian.net/browse/GN-3350) referred to, still some weird interactions with certain cursor position when you have a list like
```
- a
- b
- c
  - d
```
For example it's very hard to put the cursor after c and you can't indent b and c at the same time. But that's another different bug and will create a Jira issue for it


benjay10:
This does work a lot better! I couldn't find any issue within about 10 minutes of trying to break it.

Also, I don't have a problem with moving (by clicking, or using the arrow keys) the cursor to just after 'c' from the example above.

benjay10 (approved):

abeforgit (changes requested):
This breaks previous behavior of the indenting command

repro:
```
- a
- b
- c
- d
```
- put cursor on b, press indent. b, c and d will all get indented. whereas before, only b would get indented

While the behavior can be considered sensible, I do think the previous behavior is better

abeforgit (approved):

PR #255, opened 2022-04-13
improve whitespace collapsing

nvdk:
should address most common cases. I tested with the following:
```
<p> space before me</p>
<p>space before me</p>
<p>space <em> <strong>before </strong>me</em></p>
<p>space<em><strong> before</strong> me</em></p>
<p>space<em> before me</em> and me</p>
<p>space<em> before me </em>and me</p>
```

### DIFF
--- a/addon/model/readers/html-text-reader.ts
+++ b/addon/model/readers/html-text-reader.ts
@@ -16,7 +16,7 @@ export default class HtmlTextReader
 
     let trimmed = from.textContent;
     if (context.shouldConvertWhitespace) {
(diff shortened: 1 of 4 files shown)


nvdk:
there's a test failing related to the match text command, but I don't see how it's related to these changes?

nvdk:
found it, some parts of the code assume a br should be a block element. It's not, but I've removed it from the inline content list for now to respect the old behaviour.

abeforgit (approved):
Great work! Solves the cases from which I discovered the issue, and couldn't break it either

PR #258, opened 2022-04-20
Fix bug that selection was wrong when creating en empty list

lagartoverde:


lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-3392

abeforgit (changes requested):
works well, but breaks a few tests, please have a look

abeforgit (approved):

PR #259, opened 2022-04-20
Fix collapsed selections not detecting marks correctly

abeforgit:
What was happening was that the range.getMarks logic was taking the mark information from both sides of a collapsed selection, which in most normal cases wouldnt be the same. It then does an intersection on the marksets for all found nodes, so in the case where you just clicked lets say the bold button, it would create the node, and intersect left (the new node) and right, coming up with an empty set.

This combined with the new text-insertion logic that actually takes the mark status of the selection into account (this was not the case before) it kept inserting non-marked characters when typing.

Solution was actually already implemented in the range.contextNodes method, where it was already possible to determine in which direction to look for adjacent textnodes, it was just set to both instead of left only (this replicates google docs behavior in how it treats the "markedness" of a collapsed selection.

### DIFF
--- a/addon/commands/add-mark-to-range-command.ts
+++ b/addon/commands/add-mark-to-range-command.ts
@@ -22,7 +22,8 @@ export default class AddMarkToRangeCommand extends Command<
     const spec = this.model.marksRegistry.lookupMark(markName);
     if (spec) {
       this.model.change((mutator) => {
-        mutator.addMark(range, spec, markAttributes);
+        const resultRange = mutator.addMark(range, spec, markAttributes);
+        this.model.selectRange(resultRange);
       });
     } else {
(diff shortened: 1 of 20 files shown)


nvdk:
While testing I found some issues still exist:
- disabling bold doesn't work
- keyboard shortcuts (ctrl+b, ctrl+i, ctrl+u) don't seem to work for collapsed selections

lagartoverde:
Sorry to be that guy, but removing the marks doesn't work on complex selections, adding seems fine 
https://user-images.githubusercontent.com/17006446/164453323-3752076a-75e0-4905-a62e-c6f239c3840e.mov

nvdk (approved):
tested in the citation plugin and didn't notice any issues in there :+1: 
coincidentally also fixes this bug https://binnenland.atlassian.net/browse/GN-2942

PR #260, opened 2022-04-22
Fixed bug with making list at the end of the document

lagartoverde:
Fixes https://binnenland.atlassian.net/browse/GN-3395

### DIFF
--- a/addon/model/util/gen-tree-walker.ts
+++ b/addon/model/util/gen-tree-walker.ts
@@ -294,7 +294,6 @@ export default class GenTreeWalker<T extends Walkable = Walkable> {
       // There is a currentNode, so we can simply walk from it
       next = this.walkNode(this._currentNode, this._reverse);
     }
-
     // whatever we found is now the currentNode
     this._currentNode = next;
     if (!next) {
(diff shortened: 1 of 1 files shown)


abeforgit (changes requested):
Breaks a lot of treewalker tests, please check

nvdk (approved):
addresses the bug and there are no longer any failing tests. I don't fully grasp why we would not return the starting node though, so I'd appreciate a bit more info :)

abeforgit:
The walkFromNode function really walks _(away) from_ the node. Aka there is an implicit assumption that the node we are walking from has already been yielded in the generator in a previous step. Therefore, returning the node we are walking from should actually always result in an infinite loop in the generator, and can never be the correct return value.

addon/model/util/gen-tree-walker.ts
```diff
@@ -394,6 +389,7 @@ export default class GenTreeWalker<T extends Walkable = Walkable> {
       }
       result = this.filterNode(node);
       if (result === FilterResult.FILTER_ACCEPT) {
+        if (node === fromNode) return null;
```
abeforgit, line 392: I would assume that this check can be moved 2 lines up, can you try that? 
I think the check is also correct here, but it hinges on the fact that if the result is NOT accept, we will also return null. 
Not that this is something I can see ever changing, but it may be more robust to not rely on that?
If you move it and it works, consider the pr approved and just merge.

abeforgit: ill just leave it as it is in the interest of time

abeforgit (approved):

PR #261, opened 2022-04-26
Fix textsearch on quads defined outside the root element

abeforgit:
It was simply creating impossible ranges :man_facepalming:


nvdk (approved):
confirmed to work in embeddable, thank you!

PR #262, opened 2022-04-26
feature/improved prefix handling

abeforgit:
Result of a rabbithole that ended nowhere while debugging #261 
- Collect seen prefixes on parse
- Allow users to specify extra prefixes ad-hoc

### DIFF
--- a/addon/model/util/datastore/datastore.ts
+++ b/addon/model/util/datastore/datastore.ts
@@ -230,6 +230,7 @@ export class EditorStore implements Datastore {
       nodeToPredicatesMapping,
       quadToNodesMapping,
     } = RdfaParser.parse(config);
+    console.log([...dataset]);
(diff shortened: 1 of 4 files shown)


addon/model/util/datastore/datastore.ts
```diff
@@ -229,8 +239,13 @@ export class EditorStore implements Datastore {
       predicateToNodesMapping,
       nodeToPredicatesMapping,
       quadToNodesMapping,
+      seenPrefixes,
     } = RdfaParser.parse(config);
+    console.log([...dataset]);
```
nvdk, line 244: log statement should go

nvdk:
> 



> Result of a rabbithole that ended nowhere while debugging #262

I think you might be linking to the wrong issue/PR here :)

nvdk (approved):
didn't test the code specifically but didn't encounter any issues and the code looks sound

PR #263, opened 2022-04-27
Remove dummy Say theming

benjay10:
Some theming (s-root) from the dummy app imported a file that would override colors and fonts used in the dummy test application and the plugins. This would interfere with the styling of the plugins, and caused uncertainty about the lay-out of the plugin (for me at least). With this file removed, the plugin would be rendered as if it were in the final editor, to be able to properly fine tune the lay-out.


benjay10:
I've realised something. We could keep this PR open for a bit longer and pile up more cleanup of the Say theming code, because we are working on porting all the plugins to (Ember-)AppUniversum.
Or we could make a new PR for more of that later.

Dietr:
We lost the global appuniversum variables with removing the custom variables so added these back by using the grouped appuniversum settings file. 

I do wonder if using the appuniversum styling is correct here. We used to approach the editor as a standalone product (by styling it similar to https://say-editor.com/) which was not related to the style of Vlaanderen. This PR removes this custom style. It sure is easier to maintain this way. I don't know who decides on this styling? Maybe @nvdk

nvdk:
As long as we document how the styles can be customized I think we can drop the custom style for now.

PR #264, opened 2022-04-29
Ember upgrade to v3.28 and others

benjay10:
Basically trying to upgrade packages that depend on an older version of embroider/macros which causes problems in frontend-gelinkt-notuleren. However, this turned into a complete Ember upgrade (using ember-cli-update).

Might be best to separate the embroider/macros fixes into their own separate branch.

There are still issues with this branch, which I could not figure out:

- Invalid .eslintrc.js config, even though the part that it complains about is just like any other project.
- Incorrect tests/dummy/app/router.js types.


benjay10:
There is one more error that I havn't been able to fix. This requires more TypeScript knowledge. Hopefully @abeforgit can take a look at least.
Upon building via `ember serve` the error presented is:
```
tests/dummy/app/router.ts:5:3 - error TS2416: Property 'location' in type 'Router' is not assignable to the same property in base type 'Router'.
  Type 'string' is not assignable to type '"none" | "auto" | "history" | "hash"'.

5   location = config.locationType;
    ~~~~~~~~
```
The project continues to be built, and runs correctly otherwise.

PR #268, opened 2022-05-12
Addition of the shiftedVisually method which determines a new position based on an existing position and a number of visual steps.

elpoelma:
- Addition of the shiftedVisually method which determines a new position based on a step count in a left or right direction.
- Addition of the charactersAfter method in order to determine characters occuring after a certain position.
- Replacement of the default left and right arrow handler by an implementation which uses shiftedVisually.


abeforgit (changes requested):
Overall, it would be nice if you could add a few comments to explain the big steps in the shiftedVisually function, as it's quite dense.

Few more edgecases I found:

```xml
<h1>
    <text>test</text>
</h1><br/><br/>
```
Cursor handling around the br going into and out of the h1 still skips an extra position, in both directions

```xml
<ul>
<li><text>test</text></li>
<li><text>{$ZERO_WIDTH_SPACE}</text>​</li>
</ul>
```
if you click to the right of the li with the invisible space, you need to press left twice to go to the previous li

elpoelma:
I have added some additional comments, the problem with br tags and block elements should now be resolved.

PR #269, opened 2022-05-12
implemented merging of marks on adjacent text nodes

elpoelma:
When a specific mark is added to a text node which has an adjacent text node with exactly the same marks. The marks are merged in order to limit the number of html tags.


lagartoverde (approved):
Code looks sane and works correctly, I don't know enough about the internals to know if this could have been done better I guess that's something arne has to say. But on my part good job :)

PR #270, opened 2022-05-13
Make sidebar min-height same as window height

Dietr:
Fix for https://github.com/lblod/ember-rdfa-editor-rdfa-date-plugin/pull/6#issuecomment-1126028771


benjay10:
Functionally, this is much better!

Some nitpicks, if I may: on the date-plugin, the datepickers opens a popup, and the extra space needed dictates the height of this div to the pixel. The padding on this div does not help, because the datepicker is positioned absolute.
In Firefox, it seems there is a 1-pixel scroll, even when the div is not even half filled. Other browsers don't have this.

PR #271, opened 2022-05-16
Feature/better remove algo

Asergey91:
this new deletion algorithm will basically delete all nodes that it discovers opening tags for in between the 2 range positions, and then migrate the siblings from the end of the range up to the begining of the range
(square brackets denote the range)
from this
```
      <modelRoot>
        <span>
          <text __id="text1">te[st</text>
          <text>goodbye</text>
        </span>
        <span>
          <text>im gonna dissapear</text>
        </span>
        <div>
          <div>
            <text __id="text2">te]st</text>
            <text>moving</text>
            <text>up</text>
          </div>
          <text>staying here</text>
        </div>
      </modelRoot>
```
to this
```
      <modelRoot>
        <span>
          <text __id="text1">te[]st</text>
          <text>moving</text>
          <text>up</text>
        </span>
        <text>staying here</text>
      </modelRoot>
```
currently if the result of the delition are 2 text nodes next to the cursor it will merge them, however this should be easily extendable to other cases.


nvdk:
just wondering @Asergey91 did you have a look at the 'delete-selection' command? is there a specific reason not to use that here?

Asergey91:
> just wondering @Asergey91 did you have a look at the 'delete-selection' command? is there a specific reason not to use that here? See https://github.com/lblod/ember-rdfa-editor/blob/development/addon/commands/selection-command.ts#L33

yes, however that uses insert/old remove algos under the hood and they dont really do node merges nicely which is needed for backspace/delete hendlers

addon/model/operations/remove-operation.ts
```diff
@@ -0,0 +1,35 @@
...
+  get nodes(): ModelNode[] {
+    return this._nodes;
+  }
+
+  set nodes(value: ModelNode[]) {
+    this._nodes = value;
+  }
+
+  execute(): void {
+
```
abeforgit, line 33: why is this empty?

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,103 @@ import RangeMapper, {
...
+      node.remove();
+    });
+
+    const before = range.start.nodeBefore();
+    if (before?.parent && before.index != null) {
+      nodesToMove.forEach((node) => node.remove());
+      before.parent.insertChildrenAtIndex(before.index + 1, ...nodesToMove);
+    }
+
+    //merge logic... needs more work
```
abeforgit, line 93: what work?

abeforgit, line 93: what work?

addon/model/model-range.ts
```diff
@@ -259,6 +262,63 @@ export default class ModelRange {
     return this.start.sameAs(this.end);
   }
 
+  normalize(): void {
```
abeforgit, line 265: could you add a doc comment saying what this actually does?

abeforgit, line 265: could you add a doc comment saying what this actually does?

addon/model/model-range.ts
```diff
@@ -259,6 +262,63 @@ export default class ModelRange {
...
+    let flip = false;
+    const length =
+      this.start.path.length > this.end.path.length
+        ? this.start.path.length
+        : this.end.path.length;
+
+    for (let i = 0; i < length; i++) {
+      if (this.start.path[i] == undefined || this.end.path[i] == undefined) {
```
abeforgit, line 273: please use triple equal signs, and explicit type coercions where needed

abeforgit, line 273: please use triple equal signs, and explicit type coercions where needed

addon/model/model-range.ts
abeforgit: Looking at it, it seems like you're making sure that start is before end. On a range, this should always be the case. Where did you encounter this not being so?



Also, modelPosition.compare exists, most of this function just reimplements the same logic

addon/model/model-range.ts
```diff
@@ -259,6 +262,63 @@ export default class ModelRange {
...
+    }
+
+    if (flip) {
+      const buffer = this.start;
+      this.start = this.end;
+      this.end = buffer;
+    }
+  }
+  /**
+   * janky debug function
```
abeforgit, line 291: This seems useful, please un-jank this

abeforgit, line 291: This seems useful, please un-jank this

addon/model/model-range.ts
```diff
@@ -259,6 +262,63 @@ export default class ModelRange {
...
+    }
+    root = root.clone();
+    const range = ModelRange.fromPaths(root, this.start.path, this.end.path);
+
+    const startRange = new ModelRange(range.start, range.start);
+    const endRange = new ModelRange(range.end, range.end);
+    const startText = new ModelText('[===START===]');
+    const endText = new ModelText('[===END===]');
+
+    OperationAlgorithms.insert(endRange, endText);
```
abeforgit, line 306: This breaks a big assumption of operations being the smallest unit of state changes, please use at least an operation. You can pass undefined for the event-bus parameter. (you're working on a copy, so technically it's fine to use the algorithm directly, but it sets a bad precedent.

abeforgit, line 306: This breaks a big assumption of operations being the smallest unit of state changes, please use at least an operation. You can pass undefined for the event-bus parameter. (you're working on a copy, so technically it's fine to use the algorithm directly, but it sets a bad precedent.)

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,103 @@ import RangeMapper, {
...
+    const after = range.start.nodeAfter();
+    if (before && after) {
+      if (before.modelNodeType === 'TEXT' && after.modelNodeType === 'TEXT') {
+        (<ModelText>before).content += (<ModelText>after).content;
```
abeforgit, line 97: please use `thing as Type` syntax for typecasts

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,103 @@ import RangeMapper, {
...
+    const before = range.start.nodeBefore();
+    if (before?.parent && before.index != null) {
+      nodesToMove.forEach((node) => node.remove());
+      before.parent.insertChildrenAtIndex(before.index + 1, ...nodesToMove);
+    }
+
+    //merge logic... needs more work
+    const after = range.start.nodeAfter();
+    if (before && after) {
+      if (before.modelNodeType === 'TEXT' && after.modelNodeType === 'TEXT') {
```
abeforgit, line 96: If you use `ModelNode.isModelText()` you wouldn't need to cast on the next line

abeforgit, line 96: If you use `ModelNode.isModelText()` you wouldn't need to cast on the next line

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,103 @@ import RangeMapper, {
...
+    const openingTagNodes = allNodes.filter((node) => {
+      if (confinedNodes.includes(node)) {
+        return false;
+      } else {
+        return true;
+      }
+    });
+
+    openingTagNodes.forEach((opNode) => {
+      const nodesToUnindent = (<ModelElement>opNode).children;
```
abeforgit, line 75: please use `thing as Type` syntax for typecasts

abeforgit, line 75: please use `thing as Type` syntax for typecasts

abeforgit (changes requested):

addon/model/model-range.ts
Asergey91: yes I actaully wanted to ask you abou this, in tests this was the case, range end and range start were always based on the order of parameters passed to the constrcutor.

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,182 @@ import RangeMapper, {
...
+    const allNodes: ModelNode[] = [];
+    const walker = new ModelTreeWalker({ range: range });
+    for (const node of walker) {
+      allNodes.push(node);
+    }
+
+    //get all nodes that are fully contained in the range
+    //ie [<span><text>abc</text>]</span>
+    //would grab just the text node
+    const confinedNodes: ModelNode[] = [];
```
abeforgit, line 73: see prev comment about unpacking iterables

abeforgit (changes requested):

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,182 @@ import RangeMapper, {
...
+      if (!cantRemove) {
+        cantRemove =
+          ModelNode.isModelElement(opNode) &&
+          !opNode.getRdfaAttributes().isEmpty;
+      }
+
+      if (cantRemove) {
+        cantRemoveOpeningTagNodes.push(opNode);
+      } else {
+        const nodesToUnindent = (opNode as ModelElement).children;
```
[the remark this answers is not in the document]
Asergey91: opening tag node should always be an element... unless we can nest text nodes in each other?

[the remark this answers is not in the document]
abeforgit: theres nothing in the code that guarantees that, you're getting all nodes in a range, and then checking if they're also part of the confined nodes (which is a weird check on its own, not sure what that achieves)

addon/model/operations/operation-algorithms.ts
Asergey91: not sure that i can simplify this

addon/model/operations/operation-algorithms.ts
```diff
@@ -7,13 +7,182 @@ import RangeMapper, {
...
+
+    openingTagNodes.forEach((opNode) => {
+      //check if we can remove it
+      let cantRemove;
+
+      //check in the config const
+      cantRemove =
+        ModelNode.isModelElement(opNode) &&
+        cantRemoveOpeningTagNodeTypes.find((type) => type == opNode.type)
+          ? true
```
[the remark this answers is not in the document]
Asergey91: with .has this should no longer be ambiguous right?

Issue #144, opened 2021-08-09
Bug in ModelTreeWalker


RobbeDP:
The `ModelTreeWalker` returns wrong results in some cases when an empty range is given. This is however not always the case. Below are two examples, one that works as expected and one that doesn't work as expected. The paths at some point in the XML representation are indicated by an array of numbers.

### Works as expected:
```xml
<root>
  <div>
    [0, 0]<text>inside</text>[0, 6]
  <div>
  [1, 0]<text>outside</text>[1, 7]
</root>
```

When instantiating as follows `new ModelTreeWalker<ModelNode>({filter: toFilterSkipFalse(ModelNode.isModelText), range: range})` with `range` equals to [0] - [0, 0], we get an empty list as result.

### Doesn't work as expected:
```xml
<root>
  [0]<text>outside</text>[7]
  <div>
    [7, 0]<text>inside</text>[7,6]
  </div>
  [8]
</root>
```

When instantiating as follows `new ModelTreeWalker<ModelNode>({filter: toFilterSkipFalse(ModelNode.isModelText), range: range})` with `range` equals to [7, 6] - [8], we get the following list as result:
- `<text>outside</text>`
- `<text>inside</text>`

This should however be and empty list.

RobbeDP:
Should by fixed in #145.

nvdk:
@elpoelma could you check if this is fixed?

elpoelma:
I will look into it.

elpoelma:
Solved in #282

PR #282, opened 2022-06-09
Fix/tree walker

elpoelma:
This pull request solves issue #144. When supplying empty ranges to the ModelTreeWalker, the result is an empty list.


Asergey91 (approved):
tested with removal algorithm, works as expected, gj

abeforgit (approved):
We should move all ModelTreeWalker usage to GenTreeWalker
But for now the fix is useful and the extra tests will remain useful to make sure GenTreeWalker works as expected

abeforgit:
> We should move all ModelTreeWalker usage to GenTreeWalker But for now the fix is useful and the extra tests will remain useful to make sure GenTreeWalker works as expected

The reason this is not already the case is that GTW is missing a descend: false implementation

PR #285, opened 2022-06-21
Fix the default blueprint

Windvis:
When installing the addon I noticed the default blueprint was throwing errors due to the use of 'import' statements in a cjs module.

It seems that the addon blueprint was stored in the `blueprints/@lblod/ember-rdfa-editor` folder but ember-cli doesn't fully support nested folders. This is also the reason why `require` and `module.exports` were causing linting errors.

I now moved the `blueprints/@lblod/ember-rdfa-editor` folder to `blueprints/lblod-rdfa-editor` to bypass these issues.

Alternatively the linting config can be adjusted so the extra folder doesn't trigger linting errors, but I think sticking to the defaults is preferred?


package.json
```diff
@@ -177,7 +177,8 @@
     "edition": "octane"
   },
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "defaultBlueprint": "lblod-ember-rdfa-editor"
```
Windvis, line 181: https://cli.emberjs.com/release/writing-addons/addon-blueprints/#blueprintconfig

abeforgit (approved):
Was able to repro, and the pr fixes it! thanks!

PR #286, opened 2022-06-27
Vdom-based deletion

Asergey91:
```
intended backspace/delete behaviour

merging
  nodes that can not be merged 
    'a',
    'ul',
    'li',
    'td',
    'tr',
    'table',
    'thead',
    'tbody',
    'th',
    'all nodes with rdfa knowledge'

  nodes that you cannot merge into
    'a',
    'table',
    'tr',
    'th',
    'thead',
    'tbody',
    'ul',
    'all nodes with rdfa knowledge' 

  example
    <h1>tes[t</h1><p>]test<span>test</span></p>
  becomes
    <h1>tes[]test<span>test</span></h1>

  example
    <a>tes[t</a><p>]test<span>test</span></p>
  becomes
    <a>tes[]<a><p>test<span>test</span></p>
  opening tag gets removed and all siblings of the ending position get merged up into the parent of the starting position
  
confined selection
  if a node is completely confined in the selection it should always be removed unless it contains rdfa knowledge, in this case a user should be prompted in some way
  
visual selection and empty nodes
  at some point we will end up with empty nodes
  example
    <h1>[test</h1><p>test]</p>
  becomes
    <h1>[]</h1>

  in this case backspacing again will remove the node. However if you leave the empty node, you should still be able to position the cursor inside of it
  visual selection should ideally never select node boundaries if there is text inside of them that is selected

```
work that still needs to be done to achieve this

- [ ]   a user should be able to select text inside a node without selecting the node
- [x]   a user should be able to place cursor inside an empty node
- [ ]   there should be some sort of warning when deleting rdfa knowledge
- [x]   there is some inconsitency with model range after a selection is removed
- [x]   edge cases need to be covered in lists because at the moment they dont merge


nvdk:
I should note there's also a wealth of info in the old integration tests and the backspace sample data. Things I'm missing in the RFC: behavior at the start/end of a document is probably a special case worth covering. I don't really see why we exclude collapsed ranges, perhaps good to explain why we exclude it?

not sure I agree a 100% with what's written out here, but it's a good starting point for an in depth discussion.

nvdk:
Would be good to add that to the RFC imo

abeforgit:
I'll add it, but I'm more interested in your thoughts on rdfa handling

abeforgit:
Remaining: 
- [ ] look into cases with truly empty list items (technically not a state we allow, but still)
- [ ]  handle error in "nested divs" case on backspace sample data
- [ ] handle removal of lumpnodes
- [ ] repurpose as many integration tests into new unit tests as possible before losing sanity

nvdk:
### on removing 'empty' elements
Previously we've always been very careful about removing RDFa nodes (either inline or block elements), Mainly because not all RDFa can easily be reinserted. The same applies to elements, since at the moment we use tags in our templates that are not reinsertable. Without solving that I'm personally not a fan of removing empty elements and think that we should think about how we can make them accessible and findable.

For RDFa specifically I think extra care is required in any case, as some RDFa nodes might still have knowledge that doesn't have a visual representation (anymore). I think for those elements (but perhaps for RDFa nodes in general) you can't just remove the elements without making clear to a user that this is happening. In previous iterations we highlighted RDFa elements if you were close to clearing them out and had a similar process to lump nodes where you have to backspace twice to remove them. I'm not suggesting that's *the* solution, as it confused users but I do believe you need something like it.

### on merging (applying my logic above)
```
<rdfa>
  <text>ab[c</text>
</rdfa>
<span>
  <text>d]ef</text>
</span>
```
If a user wants to merge these two, I think it makes most sense to merge ef into the RDFa

```
<span>
  <text>ab[c</text>
</span>
<rdfa>
  <text>def]</text>
</rdfa>
```
To a user seems to mean, I want to merge these two things. What that means for the RDFa knowledge is ambiguous if the RDFa is not visible to the user.
1. we could move the rdfa to the first span
2. we can remove the rdfa element
3. we can keep an empty rdfa element

I think only 1 and 3 are an option to be honest.
-

nvdk:
Thinking about this more, I don't think it's a clear cut decision. But there's is definitely value in making a difference for the cases "I delete it and type something new" vs "I delete it and I do something else". I think specifically for those cases where the contents of an entire node is selected (but no more)

E.g. 
- if you're selecting an annotated date, delete it and start typing a new one I'd expect that to work (it won't :D)
- if I select the article number I would expect to able to type a new number and for it to be annotated correctly.

abeforgit:
Suggestion: differentiate between required and optional rdfa (default: everything required)
with optional: behave as described: delete-then-type stays in rdfa, delete-then-mode removes the empty node
with required: delete-then-move fills the node with placeholder text or adds some other css

abeforgit:
> if you're selecting an annotated date, delete it and start typing a new one I'd expect that to work (it won't :D)

did you try this? cause it does

nvdk:
Ah yes I did notice that worked, however I meant the expectation is that the actual content also reflects the new typed in date, which is not what will happen afaik.

elpoelma:
- [ ] In some cases, when trying to merge text with a previous line using backspace, a not-an-element error occurs.

elpoelma:
- [ ] In the backspace template, when pressing delete with the cursor positioned at the end of the "data" header, nothing happens.

elpoelma:
- [ ] When pressing backspace at the beginning of the first list-item of a list, the list-item is not correctly removed.

elpoelma:
Backspacing and deleting lump nodes now work, but only one backspace/delete is required instead of two.

abeforgit:
There are still some issues in tables, related to removing invisible spaces in the table cells

elpoelma:
> There are still some issues in tables, related to removing invisible spaces in the table cells

I added a modification so that delete and backspace in empty table cells do nothing.

PR #288, opened 2022-06-29
Feature/tab handler vdom

abeforgit:
solves https://binnenland.atlassian.net/browse/GN-1905

Tab behavior around lists is still not what it should be but at least it doesn't break things anymore.
Also all changes now happen on the vdom so no more unnecessary reads.
Calling it quits here cause to get proper list behavior i'd have to get way to deep into the weeds, will improve when on immutable engine.


nvdk (changes requested):
### Testing remarks
#### regression with this vdom handler tabbing out of a table
- in an empty document
- create a table
- tab until the end of the table
- tab again

Expectation: you end up after the table
Actual: cursor remains in the last cell

#### error thrown when tabbing directly after setting editor content
- insert the notulen template 
- press tab
Expectation:
Actual result: Uncaught ModelError: Domnode without corresponding modelNode

#### not jumping back into root
given the following html:
```<div>[]some content<br></div>
other content
<h1>more content</h1>
```
The cursor jumps to the start of the h1 when tabbing, I'd expect it to jump to just before other content. This use case can be found in the br sample data.

### lists
Behaviour in lists seems a bit broken, in several occasions I was jumping between a list item and it's child infinitely

elpoelma:
Traversal between text nodes seems to work mostly well, but when the document contains lists, the behaviour isn't alaways consistent (indentation or traversal). When enabling the rdfa mode, the tab handler doesnt always go over all rdfa blocks and skips some of them.

abeforgit:
disabled indenting behavior around tabs
even when working as intended it is a pretty jarring switch from the non-tab behavior
competing editors don't do particularly great in this space either
ck tabs out of text area _except_ when you're in a list-item which can be indented (aka not the first)
prosemirror doesn't handle tab at all
GDocs is the best, inserting a tab character except when at the start of any list-item, which then indents

I propose revisiting this at a later stage, current behavior is predictable and non-intrusive, I'm simply not good enough to do better it seems

abeforgit:
@Dietr this plugin adds back the `data-editor-position-level` attribute to the path of the cursor
also a bit of a showcase how we can visualize the rdfa nodes when "show annotations" is active

abeforgit:
sheesh I really need to learn to split up prs

nvdk:
shall we merge this as is?

PR #291, opened 2022-07-04
Pass options object to plugins

abeforgit:
Allows host-app to pass options to plugins.
solves https://binnenland.atlassian.net/browse/GN-3106
fully backwards-compat

# Usage

The `@plugins` argument to the rdfa-editor component now takes a list of either strings or objects.
If an object, the plugin name should be in the `name` field, and the `options` field (which can have any value) will be passed to the plugin's initialize method as a second argument. All other fields will be ignored.

e.g.:

template:
```hbs
<RdfaEditor @plugins={this.plugins} />
```
template controller:
```ts
this.plugins = [
   "plugin-without-options", 
   { 
      name: "plugin-with-options",
      options: { /* anything you want */ }
    }
];
```
plugin:
```ts
initialize(controller: Controller, options: MyOptions) {
   if (options.tralala) {
      this.doWololo();
      }
  }
```

### DIFF
--- a/addon/utils/ce/raw-editor.ts
+++ b/addon/utils/ce/raw-editor.ts
@@ -152,9 +152,8 @@ export default class RawEditor {
     const controller = new RawEditorController(config.instance.name, this);
     await config.instance.initialize(controller, config.options);
     this.logger(
-      `Initialized plugin ${config.instance.name} with options: ${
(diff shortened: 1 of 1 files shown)


nvdk (approved):
looks good to me!

PR #293, opened 2022-07-08
Null check on the parent of the range in live mark set

elpoelma:
Conservative check which determines if the ranges stored in the live mark set are still valid (still have a parent).


benjay10:
In the citaten-plugin, I can no longer reproduce the problem with the error message `this.parent is null` since editor version 0.60.x and the changes made in the citaten-plugin up to that point. Apart from that, I cannot approve or reject the changes from these commits, as I don't know enough about the editor.

abeforgit (approved):

PR #294, opened 2022-07-08
when storing the previous selection, clone the anchor nodes

elpoelma:
On each dom selection change, the ModelSelectionTracker stores the current selection in order to be able to compare it with the next selection on the next selection change. This ensures the selection is cloned in the tracker so comparisons in the next selection change are valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,22 @@
 
 
 
+## 0.60.1 (2022-07-07)
+
+#### :rocket: Enhancement
+* [#291](https://github.com/lblod/ember-rdfa-editor/pull/291) Pass options object to plugins ([@abeforgit](https://github.com/abeforgit))
+* [#287](https://github.com/lblod/ember-rdfa-editor/pull/287) Article plugin styling ([@Dietr](https://github.com/Dietr))
+
+#### :bug: Bug Fix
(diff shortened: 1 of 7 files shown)


benjay10:
Again, can't speak about the correctnes of the changes, but they do seem to have a positive effect on the citaten-plugin! It now triggers better while typing.

abeforgit (changes requested):
this is effectively the same as removing the isSameAsPrevious check, since it will now always return false. If that's what's needed, its better to simply remove the check and avoid the costly cloning.

I can see now why the comparison causes problems: a dom selection only consists of anchor and focusnodes, and offsets within them. The selection can be identical even though the nodes themselves have moved, whereas in the model we care about paths from root. 
So yeah, please remove the whole check altogether, it's incorrect.

abeforgit:
Note this check was put in place to prevent every write to also cause 2 selection changes
debugging a bit, this was actually not caused by what I thought (dom-write causing immediate selectionchange) but simply due to the way the selectionwriter updated the selection (in two steps!)
I have a fix ready

nvdk:
@elpoelma will you optimize the read after write in this PR or can this be merged?

elpoelma:
> @elpoelma will you optimize the read after write in this PR or can this be merged?

This can be merged, I will optimize it at a later time.

PR #295, opened 2022-07-08
Insert empty space when inserting an li above another one

elpoelma:


nvdk (approved):
issue is resolved and I couldn't immediately break it :+1:

PR #296, opened 2022-07-11
Remove erroneous check to avoid duplicate selectionchange events

abeforgit:
The check was never correct, with potential false positives.
Dom selection is only based on anchornode and focusnode, and their offsets.
This means the nodes themselves can move without this being visible to us.

The reason for this check was the fact that we triggered two selectionchanges every domwrite.

I used to think this was because changing the dom content would also trigger a selectionchange,
but that's actually not true. I was simply updating the selection in two steps instead of one.

Now, this does mean we still always do 1 potentially useless selection read every selection write.
After all, we just set the selection ourselves, so why should we read it again?
It's an optimization we could make, but it's not a bad safety net should the browser disagree
with our selection update, and it's not expensive to calculate.

Note that the old check also did not prevent this.

### DIFF
--- a/addon/model/writers/selection-writer.ts
+++ b/addon/model/writers/selection-writer.ts
@@ -26,11 +26,20 @@ export default class SelectionWriter implements Writer<ModelSelection, void> {
 
   write(modelSelection: ModelSelection, moveSelectionIntoView = false): void {
     const domSelection = getWindowSelection();
-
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Selection changes are now correctly and less triggered, it could indeed be useful to prevent reads just after writes in order to reduce the SelectionChange events.

PR #297, opened 2022-07-12
Allow plugins to send arguments to their components

abeforgit:
Introduce an extra option in the widgetSpec interface.
When registering a component, plugins can pass whatever information they like to
the components like so:

``` typescript
controller.registerWidget({
  componentName: "my-component",
  desiredLocation: "toolbar",
  widgetArgs: {
    hello: "world",
    // if you want the whole plugin instance to be available
    plugin: this,
  }
})
```

the component will recieve the widgetArgs object as one of its arguments.

### DIFF
--- a/addon/components/rdfa/rdfa-editor.hbs
+++ b/addon/components/rdfa/rdfa-editor.hbs
@@ -25,6 +25,7 @@
           widget.componentName
           controller=widget.controller
           plugin=widget.plugin
+          widgetArgs=widget.widgetArgs
         }}
       {{/each}}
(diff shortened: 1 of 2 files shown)


abeforgit:
@lagartoverde this is what you need for the article-structure plugin
turns out we _were_ passing the plugin, but expected plugins to pass themselves along in the register method. 
That's a bit arbitrary, so I expanded it to pass through an object with whatever you want

lagartoverde (approved):
Thanks 👍

PR #298, opened 2022-07-13
Use GentreeWalker in make-list-command

elpoelma:
This fix ensures that nodes which do not only contain text (but can also contain citations, links etc.) can be correctly converted to a list item.


abeforgit (approved):
looks good and works well

PR #301, opened 2022-07-18
Internal/tree diffing

elpoelma:
Adds a Tree differ which determines nodes are modified and which type of modification it is. This is then used in order to partially rewrite the DOM.


addon/model/util/tree-differ.ts
```diff
@@ -0,0 +1,52 @@
+import ModelElement from '../model-element';
+import ModelNode, { DirtyType } from '../model-node';
+import ModelNodeUtils from './model-node-utils';
+
+export type Difference = {
+  node: ModelNode;
+  changes?: Set<DirtyType>;
+};
+
+export default class TreeDiffer {
```
abeforgit, line 10: looks like this can be a pure function

abeforgit (changes requested):
Looks clean and seems to work well, I like that you implemented it on the nodes!
just maybe keep the differ as a pure function

PR #302, opened 2022-07-18
Improve selection handler

elpoelma:
This PR combines tree diffing and improved selection handling in order to improve typing and navigation performance.

- Tree diffing ensures that only the modified nodes are written to the DOM
- A check is added in the changedSelection-handler in order to determine whether a transaction modifying the selection is necessary.


abeforgit (approved):
Huge improvement!

PR #303, opened 2022-07-18
Reduce selectionChanged events

elpoelma:
- When reading the selection from the dom. only update the selection and emit an selectionChanged event when the selection has actually changed.

- Emit a selectionChanged event in the model in the selectRange method.


abeforgit (approved):
looks good!

PR #304, opened 2022-07-18
Fix/lump node plugin selection

elpoelma:
This modifies the set-property and remove-property commands to modify the modelnode and not replace it with a new one.
In the lump node plugin, we only remove the 'data-flagged-remove'-property if the modelnode has it.

### DIFF
--- a/addon/commands/node-properties/remove-property-command.ts
+++ b/addon/commands/node-properties/remove-property-command.ts
@@ -12,8 +12,8 @@ export default class RemovePropertyCommand extends Command {
 
   @logExecute
   execute(element: ModelElement, property: string) {
-    this.model.change((mutator) => {
-      mutator.removeProperty(element, property);
+    this.model.change(() => {
+      element.removeAttribute(property);
     });
   }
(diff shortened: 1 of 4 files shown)


elpoelma:
The old remove algorithm still uses the ModelTreeWalker which returns the wrong nodes, which causes an exception. The removeNew algorithm can not yet be used, as it merges text, which is not what we want in this situation (replacing a node between two text nodes). This PR circumvents this by not cloning and replacing the node when modifying an attribute, but simply by modifying the node itself.

elpoelma:
> The old remove algorithm still uses the ModelTreeWalker which returns the wrong nodes, which causes an exception. The removeNew algorithm can not yet be used, as it merges text, which is not what we want in this situation (replacing a node between two text nodes). This PR circumvents this by not cloning and replacing the node when modifying an attribute, but simply by modifying the node itself.

Aside from the element and text merging, removeNew also removes more elements (opening tags) than remove, so insert cannot easily replace remove by removeNew.

PR #306, opened 2022-07-24
Merge latest dev, Remove non-TEDI code, cleanup types&tests, rework commands

abeforgit:
Main api changes:
you can no longer call commands from the controller
you can now call commands from transactions with the following interface:
```
tr.commands.myCommand(args)
// and 
tr.commands.myCommand.canExecute(args)
```
Newly declared commands in typescript projects (this includes this codebase) need to be registered in the type system as follows:

```
declare module '@lblod/ember-rdfa-editor' {
   export interface Commands {
      myCommand: MyCommand
      }
}
```
This pattern is called [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). 

Controllers have a new convenience method which creates a new transaction, calls your callback on it and dispatches it, called `perform`
e.g.
```
controller.perform(tr => tr.commands.insertText({text, range}))
```
I've thought about adding a oneliner for the relatively common case where you only want to execute a single command, but I'm not convinced. I like exposing the transaction, the hope is that it will make it more visible when you are creating unnecessary transactions.

There is also `dryRun` which does the same as `perform` but does not dispatch the transaction. It is not shorter than creating a transaction and not dispatching it, but it may help express intent (creating and then not dispatching looks a bit like a mistake)

Various internal handlers now work on a controller instead of the editor instance, ideally I'd like to make sure that "anything a plugin can do, can be done via a controller", and things like handling input events are things I want plugins to be able to do.

The view and current state are now directly accessible from the controller. They should be considered read-only (but obviously are not, this is js after all).

**Known todos**: the tab handler seems broken in some cases. I'd planned to simplify it since it calculates its surrounding context in a way which is now a bit unnecesarily complex.

There are a few dummy plugins with commented out code in /addon/plugins
these are what should replace the inputplugins, and the commented is the inputplugin code added there for convenience. 
I'd hoped to convert these but I ran out of time.
This will require the capability for plugins to intercept and react to input events. This is different from reacting on transactions and is necessary alongside it for plugins to finally be able to do everything. This is also different from the old eventbus, since plugins can never create inputevents, thus avoiding any infinite loop scenarios.
Undecided if we should wrap the incoming dom events or directly pass them through.


elpoelma:
- [ ] Fix and simplify tab handler
- [ ] Fix cursor handling around tables
- [ ] Add the capability for plugins to react on input events.
- [ ] Convert input plugins

elpoelma:
The new way to call commands on the transaction and the dryrun, perform methods look sane.

elpoelma:
Conversion of input plugins:

- [ ] anchor
- [ ] list
- [ ] lump-node
- [ ] placeholder
- [ ] rdfa-confirmation?
- [ ] show-active-rdfa
- [ ] table

elpoelma:
Note: the targetRange provided with an InputEvent does not always seem to be correct, e.g. for deletion, it is possible that this range encapsulates too many nodes which results in too many nodes to be deleted.

PR #307, opened 2022-07-28
Feature/transactional api steps

elpoelma:
This PR introduces the concept of steps. The main idea is that each transaction consists of a series of steps. When the transaction is getting dispatched, transaction listeners are notified of the steps the transaction consists of and can repeatedly add steps to the transaction themselves.

This PR provides the abstract ```Step``` base class. It consists of an overridable ```execute``` method which in its most simple implementation takes a ```State``` object as input, applies some modifications to the state, and returns a ```State``` object as output.

Currently 3 implementations of ```Step``` exist:

- ```ConfigStep``` modifies the configuration of a state (given a key-value pair) and returns a new instance of a ```State``` object with the configuration modified.
- ```SelectionStep``` modifies the model-selection of a state (given a key-value pair) and returns a new instance of a ```State``` object with the model-selection modified.
- ```OperationStep``` applies an operation to the input state in place, and returns the same state object. These operations will mostly modify the content of the state document. This type of step is thus not yet immutable.

Both ```ConfigStep``` and ```SelectionStep``` work in an immutable way, they do not modify the state in place, but return a new instance of the state with their modifications.

Instead of subscribing to newly added operations, transaction-listeners now subscribe to newly added steps.

The ultimate goal is to make operations completely immutable: they ultimately should create a new version of the document. This document should not be a brute-force clone, but should only replace/clone the nodes which need to be replaced: the modified/removed/added nodes and the chain of parents.

TODO:

In order to implement fully immutable operations, model-nodes and model-positions require some structural changes:
- ```ModelNode``` instances should be parent-independent as well as sibling-independent. This means that a single model-node can be reused in different documents. Instances of ```ModelElement``` still keep track of their children. This means that when creating a new instance of a document with some modifications, most nodes can be reused.
- ```ModelPosition``` instances can be resolved based on the root node and a path of indices. A position keeps track of a list with nodes and indices.

Model-node parents and siblings can be resolved given a document-object/root-node. 

This concept also lays out the way to partial RDFa-updates as not all nodes in the document are copied.
RDFa-updates should only be necessary when the content of the document changes. Simple transactions with only selection/configuration changes should not have to recalculate the datastore based on the RDFa.

By not having to clone the whole document/update the whole datastore, the performance of transactions should increase significantly.


elpoelma:
Additionally, this PR also gets the show-active-rdfa plugin working again.

abeforgit (approved):
This feels very natural, I love it!

PR #308, opened 2022-08-08
Bump used node version and builder image

abeforgit:


abeforgit:
Fun fact: `npm ci` also runs `prepare` since npm 8

Dockerfile
```diff
@@ -4,6 +4,7 @@ LABEL maintainer="info@redpencil.io"
 
 WORKDIR /app
 COPY package.json package-lock.json ./
+COPY public ./public/
```
nvdk, line 7: why was this necessary?

abeforgit: because new npm now runs prepare scripts on `npm ci` :) which in our case includes the `svg-symbols` invocation, which needs the public folder to exist

PR #309, opened 2022-08-11
Remove the appuniversum SASS includePath

Windvis:
ember-appuniversum 1.3+ no longer uses the `@appuniversum/appuniversum` package so this config fails the build if the project updated to that version (and the host app doesn't install the package itself).

The config isn't needed on older ember-appuniversum versions either since those also add the includePath already.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,9 @@ module.exports = {
       app.options.sassOptions.includePaths || [];
 
     app.options.sassOptions.includePaths.push(
-      'node_modules/@appuniversum/appuniversum',
+      // This is needed for the editor plugins dummy app styles to work.
+      // They import the ember-rdfa-editor/a-dummy file, which imports the appuniversum styles
+      // with an absolute path which only works if this includePath is added.
       'node_modules/@appuniversum/ember-appuniversum/app/styles'
     );
   },


index.js
```diff
@@ -34,7 +34,9 @@ module.exports = {
       app.options.sassOptions.includePaths || [];
 
     app.options.sassOptions.includePaths.push(
-      'node_modules/@appuniversum/appuniversum',
+      // This is needed for the editor plugins dummy app styles to work.
+      // They import the ember-rdfa-editor/a-dummy file, which imports the appuniversum styles
+      // with an absolute path which only works if this includePath is added.
       'node_modules/@appuniversum/ember-appuniversum/app/styles'
```
Windvis, line 40: It seems this is needed because the dummy apps import the `ember-rdfa-editor/a-dummy` file which [imports from ember-appuniversum](https://github.com/lblod/ember-rdfa-editor/blob/29c948eff52a1d44c1391083ecd8f79d0eecba19/app/styles/ember-rdfa-editor/_a-dummy.scss#L6). In theory I think it should be possible to convert that to `@import "../ember-appuniversum/a-settings";` but it seems ember-cli-sass still can't find the file then even though it's there on the "file system".

I assume it was added here so it didn't need to be added to each plugin's dummy app config?

Windvis: I think we can in theory move `ember-rdfa-editor/a-dummy` up a level and import it like that, similar to how it works for the `ember-rdfa-editor.scss` file. Dummy apps would still need to be updated, but the config is no longer needed.

abeforgit: It was added because we were tired of broken builds and this seemed to fix it
I don't think anyone working on this project understands what is actually going on with sass imports

Windvis: Yea, ember-cli-sass(?) seems to have issues with relative paths that refer to scss files of other addons (that are supposedly merged with the `app/styles` folder of the host app).

So the `ember-rdfa-editor/_a-dummy.scss` file can't import `ember-appuniversum/something` because that file doesn't exist in the `ember-rdfa-editor` folder, but `../ember-appuniversum/something` doesn't work either for some reason. Moving that dummy file up a level makes the `ember-appuniversum/something` import work as well. I have no idea what the difference is. The includePath sidesteps the issue because ember-cli-sass also looks in the node_modules folder for the file. 

In any case, since it's only used for the dummy apps of the editor (and plugins) I wouldn't ship that includePath to consuming apps since we might run into similar issues in the future, but changing it would break all the editor plugin dummy apps that depend on this being set.

Windvis:
Thanks!

PR #310, opened 2022-08-20
Transactional Edits

abeforgit:
Implementation of https://github.com/lblod/ember-rdfa-editor/pull/311

While I don't want to make a habit of these monster PRs, I do want to mention that having a branch free from overhead and worry about compatibility was invaluable to get something like this off the ground. More than just a refactor, I feel like we have a much better understanding of where we're going now, and i am very grateful for the freedom we had with this (I know I forced it a bit, but I appreciate not being stopped :smile:  )

### wiring in place for letting the browser do its thing

Hence the name of the branch, this was the original goal like 3 months ago. The way we catch up after the browser has made the change is through the use of a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver). It's the only reliable way, we can't use what the browser _promises_ it will do cause it frequently lies.
Currently only text insertion that is not in a mark is handled this way.

Note the pause() and resume() methods in the input-handler, which are called before any dom changes are made to prevent infinite loops. 

### we use beforeinput events now

It's a nicer API, although not as nice as I initially hoped. Basically just better semantics, the `getTargetRanges()` method looked promising but its ultimately unusable due to the spec being really loose and browsers just giving you garbage from time to time.
see input-handler.ts for the logic

### keymaps are a thing

They're also dynamically adjustable from transactions, technically, although I haven't added a method for it yet.
See keymap.ts, yes, the key specifier string is typechecked, keep your pants on please.

### input-plugins are gone

The concept on "input-plugin" is no longer needed, since "real plugins" can provide the same event-based logic.
Previous input-plugins are converted into in-repo plugins, simply part of the core codebase so unnecessary messing with dependencies is avoided.

The API surface for intercepting DOM events can and will improve, especially the logic around reusing or extending the default handling.

### no more nodeviews or boundNodes

There is no longer an explicit binding between any given modelNode and a DOMnode. Instead, the translation happens purely
through positions and the tree structure. This is definitely the most fragile logic here, since it implicitly depends on the exact implementation of things like marks and inline components. It is concentrated in 2 methods in dom-helpers, modelPosToDomPos and domPosToModelPos. 

Why? positions stay meaningful across copies of the document tree, object references do not. We sacrifice a bit of speed - O(1) vs O(technically _n_ but let's be honest its gonna be more something like _log n_) - and simplicity, but we get the ability to work across document clones and we have to do a whole lot less bookkeeping.

### Controllers are still a thing

I didn't really mention them in the RFC, but controllers remain. They are now a wrapper around the view, providing slightly modified apis and allowing every actor to be distinguished by the controller they use. I like this design and I don't really see it going anywhere.

### node.addChild now removes the child from its parent

 see [this commit](https://github.com/lblod/ember-rdfa-editor/pull/310/commits/b5d989f9ce76ba5f56aa64597d6f9050bc5f9c7c)
This is the first change to this method in over a year, and for such a core method it's important to be aware.
Sort of baffling how long this went unnoticed. Very easy to create cycles by adding a node from one subtree to another. The childnode would get the correct parent ref, but both the old parent and the new one would point to the child. 
Solved more than it broke, and I don't think many plugins directly use this method (or at least not in situations that matter)

### Implementation of transactions 

Documents are lazily cloned into a "workingcopy", which is then mutated. This is to allow us to reuse the logic of the mutator.
Another RFC is in the works detailing how I want to improve this. (essentially, avoid cloning the entire tree and reuse unmodified subtrees)

### selection after list commands is incorrect (but consistent, and somewhat logical)

The reason this broke is that it heavily depended on node references staying valid throughout edits. The reason it's not fixed is because I want to handle this with smarter RangeMappers, which I want to implement using simple integer positions (aka global text offsets)

The idea here is that it is much easier to express the concept of "the same visual position" after an insert operation with a simple position. Using current modelpositions, it is probably also possible, but much more error prone, and simple positions have other benefits, see https://binnenland.atlassian.net/browse/GN-3620

The pr also adds failing tests for this (feel free to add more, I only did indent-list)

### eventbus

It's still floating around, though it's not really used anymore. It might make a comeback still, we might want to wrap dom events before we send em to plugins or something. 

### Fixing issues with behaviour of inserting new inline components.

Currently only the wrapping dom nodes of inline components are inserted but the editor ember component is not correctly notified of new inline components so it does not render them.

### general cleanup (optional)

- mutator can go
- readers and writers can become pure functions
- simplifiedmodel and modelhistory can go
- node-view can go
- would be good to consolidate all the utils we have into one util folder in top level
- content-editable component is starting to look a bit redundant now that the view sets up its own handlers

Probably best to squash on merge

Once the new plugin api has been put through its paces, I would consider this to be v1.0.0


elpoelma:
general cleanup (optional)

- [ ]    mutator can go
- [ ]     readers and writers can become pure functions
- [ ]     simplifiedmodel and modelhistory can go
- [ ]     node-view can go
- [ ]     would be good to consolidate all the utils we have into one util folder in top level
- [ ]     content-editable component is starting to look a bit redundant now that the view sets up its own handlers

abeforgit:
Think I came back around on CE being redundant, the separation is useful for keeping the hbs clean

PR #312, opened 2022-08-30
Replace selection arg by range arg in insert-component-command

elpoelma:
Previously, the insert-component command required a selection argument, this is replaced by a range argument in this PR.


abeforgit (approved):

abeforgit:
note this is breaking

PR #315, opened 2022-08-31
docs(rfc): Add datastore recalc rfc

abeforgit:


rfcs/datastore-performance.md
```diff
@@ -0,0 +1,77 @@
...
+  let current = root.clone();
+  let newRoot = current;
+  newParser.visit(current);
+  
+  for node in pathFromRootWithoutRoot(oldNode, root):
+    oldParser.visit(node);
+    clone = node.shallowClone()
+    current.replaceChild(node, clone)
+    newParser.visit(clone);
+
```
elpoelma, line 67: Current should probably be updated here with the value of clone?

elpoelma (changes requested):
Looks good!

PR #319, opened 2022-09-06
Fix/ember appuniversum

abeforgit:
This cleans up the sass config in the editor by removing deprecated styles and reordering the declaration of the dummy component.
This way we follow best practices for addons which expose their sass to apps.

For usage of the editor, the procedure is now identical to ember-appuniversum:
simply add `@import 'ember-rdfa-editor';` to your app.scss. If the app is also using appuniversum directly (which it likely is), the `@import 'ember-appuniversum';` should remain there (and likely precede the editor import, though I'm not sure it matters).

EDIT: (see conversation below)
We now depend on `ember-appuniversum` sass to be imported in the host app _before_ the import of the editor sass
appuniversum was a peerdep already, so to my dismay this actually kinda makes sense

Once released, I will create the prs for all relevant plugins.


app/styles/ember-rdfa-editor.scss
```diff
@@ -23,6 +22,10 @@
 // OBJECTS
 @import "ember-appuniversum/a-objects";
 
+@import "ember-appuniversum/a-generic";
```
Windvis, line 25: Hmm, assuming an app has these imports:

```scss
@import "ember-appuniversum":
@import "ember-rdfa-editor":
```

I _think_ the css will be duplicated if it's imported both here and in `ember-appuniversum.scss`. AFAIK `@import` does a "dumb" copy paste of the code which means some of these files will be present twice in the final output. It's only safe to do for things like variable definitions.

So I think `ember-rdfa-editor` needs to export only it's own custom styles so consuming apps can import it as a "plugin".

The dummy apps simply need to import the Appuniversum styles as well.

abeforgit: You're probably right, but for the record, I really, really hate this

abeforgit: but the editor references appuniversum styles in its own custom styles, such as variables and utilities
what do I do with those then?

abeforgit: I guess I can just assume they'll be there since we peerdep on appuniversum then?

abeforgit: Seems so

Windvis: Yea, it's not nice. I think things might improve if we switch to [`@use`](https://sass-lang.com/documentation/at-rules/use) instead of [`@import`](https://sass-lang.com/documentation/at-rules/import). The disclaimer on the `@import` page also seems to suggest this is the way to go. I haven't looked into the specifics myself yet though, but @Dietr might have some insights.

To be fair, the editor already lists ember-appuniversum as a peerDependency which assumes the consumer is using Appuniversum as well, so I don't necessarily think it's a bad thing to expect that the app loads the Appuniversum styles in a way that the editor can depend on that.

Long term (v2 addons) I think it might be better to switch to a "each component imports its own styles" system and discard the Sass setup altogether. I don't think apps configure the Sass variables (or _should_ even) and I assume most of it could be replaced with CSS variables if needed.

Windvis: > but the editor references appuniversum styles in its own custom styles, such as variables and utilities

importing a variables definition file should be safe to do since it doesn't output any css. I'm not sure what you mean with utilities but as long as it doesn't output css by simply importing the file it should be good.

PR #358, opened 2022-09-13
Introduction of a MarksManager

elpoelma:
This PR introduces a MarksManager class, seperate from the existing MarksRegistry. The functionality of the MarksManager includes:
- updating the different mark instances in a document 
- the possiblity to retrieve mark instances grouped by type/owner.

When retrieving mark instances given a mark type, an array containing VisualMarkGroups is returned. Each VisualMarkGroup contains a list of mark instances which are visually rendered as the same mark. (e.g. adjacent text with the same mark types).

For a specific transaction, the MarksManager can be retrieved using the `getMarksManager()` function.

This PR also removes the Mark management functionality from the MarksRegistry. The MarksRegistry will only be responsible for keeping track of the registered MarkSpecs.

### DIFF
--- a/addon/core/model/readers/html-reader.ts
+++ b/addon/core/model/readers/html-reader.ts
@@ -1,14 +1,11 @@
-import { MarkSpec } from '@lblod/ember-rdfa-editor/core/model/marks/mark';
 import MarksRegistry, {
   SpecAttributes,
 } from '@lblod/ember-rdfa-editor/core/model/marks/marks-registry';
 import ModelNode from '@lblod/ember-rdfa-editor/core/model/nodes/model-node';
-import ModelText from '@lblod/ember-rdfa-editor/core/model/nodes/model-text';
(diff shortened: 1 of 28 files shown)


elpoelma:
Additionally, this PR also includes changes to the toolbar widget-location. It is now split into toolbarMiddle and toolbarRight.

addon/core/model/readers/html-reader.ts
```diff
@@ -61,13 +58,13 @@ export class HtmlReaderContext {
...
   }
 
-  addMark<A extends AttributeSpec>(
-    node: ModelText,
-    spec: MarkSpec<A>,
-    attributes: A
-  ) {
-    return this.marksRegistry.addMark(node, spec, attributes);
-  }
+  // addMark<A extends AttributeSpec>(
```
abeforgit, line 61: remove commented code

abeforgit (changes requested):
Looks good, but could you add a test or two for the marks manager calculation?

PR #362, opened 2022-09-19
Implement htmlContent setter and getter on the RawEditorController

elpoelma:
This PR moves the htmlContent setter and getter implementations to the RawEditorController class.


lagartoverde (approved):
Looks good

PR #363, opened 2022-09-19
Remove tracked array from inline components registry

elpoelma:
This PR removes the tracked modifier from the `activeComponents` array in the inline components registry as it was causing issue with circular tracking.
In order to replace the purpose of the tracked array, `modelWritten` events were used. The `rdfa-editor` components updates its component instances in reaction to a `modelWritten` event.

### DIFF
--- a/addon/components/rdfa/rdfa-editor.ts
+++ b/addon/components/rdfa/rdfa-editor.ts
@@ -119,7 +119,6 @@ export default class RdfaEditor extends Component<RdfaEditorArgs> {
     if (this.args.rdfaEditorInit) {
       this.args.rdfaEditorInit(rdfaDocument);
     }
-    // this.initializeComponents();
     this.editorLoading = false;
   }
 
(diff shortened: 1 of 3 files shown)


addon/components/rdfa/rdfa-editor.ts
```diff
@@ -116,7 +119,7 @@ export default class RdfaEditor extends Component<RdfaEditorArgs> {
     if (this.args.rdfaEditorInit) {
       this.args.rdfaEditorInit(rdfaDocument);
     }
-    this.initializeComponents();
+    // this.initializeComponents();
```
lagartoverde, line 122: Are we keeping this commented for any reason

lagartoverde (approved):
Looks good

addon/components/rdfa/rdfa-editor.ts
elpoelma: No my mistake, I will remove it.

PR #365, opened 2022-09-20
Fix issue with inline components not being persisted correctly across reloads

elpoelma:
This PR ensures that the ember inline components are loaded on startup of the editor. Fixes https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4/backlog?view=detail&selectedIssue=GN-3675&issueLimit=100


addon/components/rdfa/inline-component-manager.ts
```diff
@@ -17,15 +17,20 @@ export default class InlineComponentManager extends Component<InlineComponentMan
     this.args.controller.addTransactionDispatchListener(
       this.updateInlineComponents.bind(this)
     );
+    this.refreshInlineComponents();
```
abeforgit, line 20: it might be better to run this on did-insert, I'm not sure

PR #384, opened 2022-09-30
Chore/cleanup feature flags

nvdk:
drop unused feature flags and be more consistent about the other ones. we should probably also consider:
- dropping the cut and copy feature flags
- moving extended html vs  html vs text paste to a config of the editor and not a feature flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,8 @@ module.exports = function(environment) {
 };
 ```
 
-* editor-html-paste: if enabled, support html paste input
-* editor-extended-html-paste: if enabled, support extended html paste input (old behavior)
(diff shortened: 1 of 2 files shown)


abeforgit:
In tedi, cut, copy and paste can be implemented as plugins, which can then be disabled by not adding them
we could then export 2 components, one "fully featured" which includes all the "default" and one barebones, where every plugin has to be specified explicitly

abeforgit:
while we're at it, could you elaborate a bit what the "extended" means in "extended-html-paste"? Even I don't know that :stuck_out_tongue:

abeforgit (approved):

PR #386, opened 2022-09-30
Fix: inline component selection issues

elpoelma:
This PR includes changes to:
- the `fromInNode` method, it now takes into account inline components
- the `MutationObserver` handler, it now ignores mutation targets which are not content editable. (In order to ensure inline components are not taken into account)

This PR solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3706.

### DIFF
--- a/addon/commands/insert-component-command.ts
+++ b/addon/commands/insert-component-command.ts
@@ -3,15 +3,12 @@ import {
   Properties,
   State,
 } from '../core/model/inline-components/model-inline-component';
-import ModelElement from '../core/model/nodes/model-element';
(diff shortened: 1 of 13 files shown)


abeforgit:
could you add a test inline component to the dummy plugin? and a button to insert/delete it?
That would make it easier to test

elpoelma:
> could you add a test inline component to the dummy plugin? and a button to insert/delete it? That would make it easier to test

Good idea, will get on it!

abeforgit (approved):
nice!

PR #389, opened 2022-10-03
Enable html pasting by passing a property to the editor component

elpoelma:
This PR enables HTML Pasting for the TEDI API. Users can pass a `pasteBehaviour` argument to the editor component.
This property can be one of three values:
- `textonly`
- `standard-html`
- `full-html`
The default value is `standard-html` (only keep html structures which are supported by the editor).


abeforgit:
first off love the renaming of the config values, much more clear

abeforgit (approved):

PR #396, opened 2022-10-05
Fix issue with computing differences on transaction dispatch 

elpoelma:
This PR introduces the following fixes and improvements:
- Text insertion is now again completely handled by the virtual dom instead of the browser. This solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-3729. It does unfortunately reduce typing performance.
- The `moveCursor` input handler now uses state only dispatch to prevent an extra write and difference calculation
- When a transaction is created, `rdfInvalid` and `marksInvalid` are initialized to `false` to prevent unnecessary mark and datastore calculations.
- Internally, the datastructure which keeps track of transaction listeners has been changed from a list to a set.
- The toolbar now uses a `TransactionDispatchListener` in order to update the status of its buttons.
- Transaction listeners have been mostly converted to arrow function in order to be able to omit the use of `this` binding.

### DIFF
--- a/addon/components/rdfa/editor-toolbar.ts
+++ b/addon/components/rdfa/editor-toolbar.ts
@@ -40,12 +40,12 @@ export default class EditorToolbar extends Component<Args> {
 
   @action
   didInsert() {
-    this.args.controller.addTransactionStepListener(this.update.bind(this));
+    this.args.controller.addTransactionStepListener(this.update);
   }
 
   @action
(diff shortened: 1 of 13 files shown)


abeforgit:
Note that changing to a Set is only OK because js Sets are ordered sets as defined in the spec, cause the ordering of listeners is important

PR #410, opened 2022-10-12
feature/simple positions

abeforgit:
implements the simple positions from the [modelnode](https://github.com/lblod/ember-rdfa-editor/blob/master/rfcs/model.md) rfc
closes https://binnenland.atlassian.net/browse/GN-3620
First I tried converting places we were using the modelposition to simple positions, but I ended up rolling back those changes and do them as part of the modelnode rework

### DIFF
--- a/addon/core/model/nodes/model-node.ts
+++ b/addon/core/model/nodes/model-node.ts
@@ -3,6 +3,7 @@ import ModelElement from '@lblod/ember-rdfa-editor/core/model/nodes/model-elemen
 import {
   ModelError,
   NoParentError,
+  NotImplementedError,
   OutsideRootError,
 } from '@lblod/ember-rdfa-editor/utils/errors';
 import XmlWriter from '@lblod/ember-rdfa-editor/core/model/writers/xml-writer';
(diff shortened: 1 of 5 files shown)


addon/core/model/simple-position.ts
```diff
@@ -0,0 +1,91 @@
...
+          cur = cur.firstChild;
+        } else {
+          return ModelPosition.fromInNode(cur, simplePos - count);
+        }
+      } else {
+        if (ModelNode.isModelText(cur)) {
+          return ModelPosition.fromInNode(cur, simplePos - count);
+        } else {
+          return ModelPosition.fromBeforeNode(cur);
+        }
```
elpoelma, line 41: In theory, this case shouldn't be able to happen right?

addon/core/model/nodes/model-node.ts
```diff
@@ -143,6 +144,20 @@ export default abstract class ModelNode implements Walkable {
...
+    } else if (this.isLeaf) {
+      return 1;
+    } else if (ModelNode.isModelInlineComponent(this)) {
+      return 1;
+    } else if (ModelNode.isModelElement(this)) {
+      return this.children.reduce((prev, node) => prev + node.size, 0) + 2;
+    } else {
+      throw new NotImplementedError();
+    }
+  }
```
elpoelma, line 159: If we introduce a size getter on modelnodes, we should probably do away with length (or rename it) as I can imagine it can lead to some confusion.

elpoelma (approved):
Code looks sane and nicely written.

addon/core/model/nodes/model-node.ts
abeforgit: its not the same thing though

addon/core/model/simple-position.ts
abeforgit: no, afaict not

PR #451, opened 2022-11-11
GN-3716 - Improve copy/paste from word

usrtim:
This is a solution that I came across during my experimentation. Now I can copy and paste a list from ms word(desktop version) to the editor. Because I don't know yet If I may break something with this change, I just created this PR for the team to check the solution. After that, I will do the refactoring part.


abeforgit (changes requested):
Looks like a decent library, good find!
some more remarks:
- we should still send the cleaned html through our own cleaner, to cleanup inline styles
- I would like some tests for the rtf -> html conversion flow. Best way to do that is to extract the logic into a simple input->output function and test it like that. The reason is that once we turn this on, it has to stay consistent, so we need at least some idea of what output we can expect. This also makes updating the library less dangerous.

And finally a question: did you try any other libs? if so, what made you choose this one? If not, why not?

usrtim:
> Looks like a decent library, good find! some more remarks:
> 
>     * we should still send the cleaned html through our own cleaner, to cleanup inline styles
> 
>     * I would like some tests for the rtf -> html conversion flow. Best way to do that is to extract the logic into a simple input->output function and test it like that. The reason is that once we turn this on, it has to stay consistent, so we need at least some idea of what output we can expect. This also makes updating the library less dangerous.
> 
> 
> And finally a question: did you try any other libs? if so, what made you choose this one? If not, why not?

@abeforgit yes, during my research I was investigating lots of approaches. I saw and tried other libraries as well, but this was the only one that I could find to help us treat our case.

tests/unit/utils/rtf-to-html-conversion-test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const htmlContent =
+      '<html xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:w="urn:schemas-microsoft-com:office:word" xmlns:m="http://schemas.microsoft.com/office/2004/12/omml" xmlns="http://www.w3.org/TR/REC-html40"> <head> <meta http-equiv=Content-Type content="text/html; charset=utf-8"> <meta name=ProgId content=Word.Document> <meta name=Generator content="Microsoft Word 15"> <meta name=Originator content="Microsoft Word 15"> </head> <body lang=en-AL> <!--StartFragment--><span style=\'font-size:14.0pt;font-family:"Calibri",sans-serif; mso-ascii-theme-font:minor-latin;mso-fareast-font-family:Calibri;mso-fareast-theme-font: minor-latin;mso-hansi-theme-font:minor-latin;mso-bidi-font-family:"Times New Roman"; mso-bidi-theme-font:minor-bidi;mso-ansi-language:#0C00;mso-fareast-language: EN-US;mso-bidi-language:AR-SA\'>of</span><!--EndFragment--> </body> </html>';
+
+    const content = pasteHandler(rtfContent, htmlContent);
+    // when handling the RTF content, and the process is done successfully the pasteHandler -> cleanDocx function attaches <!--StartFragment--> in the output
+    assert.true(content.includes('<!--StartFragment-->'));
```
abeforgit, line 13: that's a very light test, still not telling us what the lib is actually doing
I'm mainly interested in testing _how_ it converts, not whether it can handle correct input
if you look at most other tests, they compare some expected output to the actual output.
It's ok in this case to let the library parse the input, and then use that in the test. We're not looking for bugs in the library, rather we want to keep a record of how we expect it to behave, so that when things change we have a better understanding of what changed. 
Tip: We use https://www.npmjs.com/package/common-tags in this project, so you can use e.g. `oneLineTrim` to write html in the test but still format it nicely

addon/utils/ce/paste-handler-func.ts
```diff
@@ -0,0 +1,12 @@
+import { cleanDocx } from '@prezly/docx-cleaner';
+
+export default function pasteHandler(
```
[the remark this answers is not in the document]
abeforgit: that being said, I'd rather test the whole paste-handler logic than just the library. That allows us to change the implementation but be guarded against regressions.

the best way to do that is to extract the conversion logic into a function, use it in the pastehandler, and then test the function. So the pasteHandler coould look something like
```ts
      try {
        const inputParser = pasteExtendedHTML
          ? new HTMLInputParser({})
          : new HTMLInputParser({ safeTags: LIMITED_SAFE_TAGS });

        const htmlPaste = clipboardData.getData('text/html');
        const rtfPaste = clipboardData.getData('text/html');
        let cleanHtml;
        if(rtfPaste)  {
          cleanHtml = convertMsWordHtml(htmlPaste, rtfPaste);
          } else {
          cleanHtml = convertGenericHtml(htmlPaste);
          }

        this.rawEditor.executeCommand('insert-html', cleanHTML, pasteRange);
      } catch (error) {
        // Fall back to text pasting.
        console.warn(error); //eslint-disable-line no-console
        const text = this.getClipboardContentAsText(clipboardData);

```
and then we'd test the convertMsWordHtml function in isolation

addon/utils/ce/paste-handler-func.ts
```diff
@@ -0,0 +1,12 @@
...
+  rtfPaste: string,
+  htmlPaste: string
+): string {
+  if (!rtfPaste && !htmlPaste) return 'No content found!';
```
abeforgit, line 7: I should add a linting rule for this, but the convention in most of our projects is to always use braces for conditionals, even if they can be oneliners
it's just a style choice really, we feel like its easier to make mistakes with braceless conditionals

abeforgit (changes requested):

addon/utils/ce/paste-handler-func.ts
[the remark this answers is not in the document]
usrtim: Done

tests/unit/utils/rtf-to-html-conversion-test.ts
usrtim: @abeforgit  Please check if the new commit addressed also this point

tests/unit/utils/rtf-to-html-conversion-test.ts
```diff
@@ -1,32 +1,66 @@
...
 
-    const content = pasteHandler(rtfContent, htmlContent);
-    assert.strictEqual(content, htmlContent);
+    assert.strictEqual(expectedHtml, actualHtml);
   });
 
-  test('It should display "No content found!"', function (assert) {
-    const rtfContent = '';
-    const htmlContent = '';
+  test('sbx6 It should display clean HTML', function (assert) {
```
abeforgit, line 38: what is sbx6? just curious :)

usrtim: ahah it's like a "nickname" my friends used to call me :D

tests/unit/utils/rtf-to-html-conversion-test.ts
```diff
@@ -1,32 +1,66 @@
...
+  test('sbx6 It should handle rtf -> html correctly', function (assert) {
+    const rtfContent = String.raw`{\\\\rtf1\\\\ansi\\\\ansicpg1252\\\\cocoartf1265\\\\cocoasubrtf210\\n{\\\\fonttbl\\\\f0\\\\fswiss\\\\fcharset0 Helvetica;}\\n{\\\\colortbl;\\\\red255\\\\green255\\\\blue255;}\\n\\\\paperw11900\\\\paperh16840\\\\margl1440\\\\margr1440\\\\vieww10800\\\\viewh8400\\\\viewkind0\\n\\\\pard\\\\tx566\\\\tx1133\\\\tx1700\\\\tx2267\\\\tx2834\\\\tx3401\\\\tx3968\\\\tx4535\\\\tx5102\\\\tx5669\\\\tx6236\\\\tx6803\\\\pardirnatural\\n\\n\\\\f0\\\\fs24 \\\\cf0 Hello, World!\\\\\\n\\\\\\nThis is RTF. :-)`;
+    const expectedHtml = '<span>Lorem Ipsum</span>';
```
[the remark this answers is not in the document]
usrtim: ok, so I'm just pasting empty string for now as parameter for rtf

usrtim:
> starting to get there. But the tests are still very basic. I'd like to see (at least) 2 more tests added, one for the ms word flow and one for the generic flow, but with a document containing the following:
> 
>     * a list
> 
>     * a list with some items indented
> 
>     * a table
> 
>     * some text that's bold, italic, and/or underlined
> 
>     * a header
> 
> 
> and if possible:
> 
>     * a list in a table
> 
>     * a table in a list
> 
>     * whatever other weird scenario you can think of, except images (we explicitly don't support those yet so we don't have to worry about them)
> 
> 
> The reason people paste from word is _because_ these complex structures are easier to create there, so we definitely need to test them.
> 
> if you want you can spread these cases out over multiple tests, that would be ideal, but I've been annoying enough already so I'll accept one big test as well
> 
> about the linter errors: don't worry about the "unsafe assignment/call of an any value" **in this case** (do fix the one about prettier formatting though). It's because the lib you added doesn't have types, I'll fix that once this pr is good to go and then explain how it works.

@abeforgit please have a look now. The only thing that it's not supported now, are lists with some items indented. I also removed the usage of  **common-tags**, because when using that HTML provided by **common-tags** on my functions - **convertMsWordHtml** and **convertGenericHtml** tests fail because produce some weird HTML.

tests/unit/utils/rtf-to-html-conversion-test.ts
abeforgit: aha ok :)
you don't have to take it out, but it's also not necessary to add it in the future
we have `git blame` if we wanna see who wrote which piece of code

abeforgit:
> The only thing that it's not supported now, are lists with some items indented. 

what do you mean with "not supported"? what does it do? please add a test for it

abeforgit:
> > The only thing that it's not supported now, are lists with some items indented.
> 
> what do you mean with "not supported"? what does it do? please add a test for it

I tried it out, and it seems they are producing invalid html in this case
instead of 
```
<ul>
   <li>
      <ul>
          <li>
``` 
they do
```
<ul>
  <ul>
     <li>
 ```

**Comment by abeforgit** [comment]
we can't ship it like this, we would need to fork the library or just port over the logic and do it ourselves

**Comment by usrtim** [comment]
> 

nvm, I just during my copy-paste html added more space. Now I'm switched again using **common-tags**

**Comment by usrtim** [comment]
> > > The only thing that it's not supported now, are lists with some items indented.
> > 
> > 
> > what do you mean with "not supported"? what does it do? please add a test for it
> 
> I tried it out, and it seems they are producing invalid html in this case instead of
> 
> ```
> <ul>
>    <li>
>       <ul>
>           <li>
> ```
> 
> they do
> 
> ```
> <ul>
>   <ul>
>      <li>
> ```

Yes, the only thing that I found It's that nested lists are not supported. In all other cases(you can see also from my tests, eveything looks fine)

**Comment by usrtim** [comment]
> we can't ship it like this, we would need to fork the library or just port over the logic and do it ourselves

Ok sure, I will see what we can do about it ..

**Comment by abeforgit** [comment]
yeah the problem is the way it breaks
it would be not so bad if it would make a list thats unindented or something
but because its invalid html the content of the list items gets lost and thats not something we can live with

File: addon/utils/ce/paste-handler-helper-functions/cleanDocx.ts [hunk 6]
```diff
@@ -0,0 +1,18 @@
...
+} from '@lblod/ember-rdfa-editor/utils/ce/paste-handler-helper-functions';
+
+const parser = new DOMParser();
+
+export function cleanDocx(html: string): string {
+  const document = parser.parseFromString(preCleanHtml(html), 'text/html');
+  const { body } = document;
+
+  cleanEmptyElements(body);
+  cleanListElements(body);
```
abeforgit, line 15: why did you remove all the other cleaners here?

tests/unit/utils/convert-ms-word-html-test.ts
```diff
@@ -0,0 +1,1240 @@
...
+                  <o:p></o:p>
+               </span>
+            </p>
+            <!--EndFragment-->
+         </body>
+      </html>
+    `;
+
+    const actualHtml = convertMsWordHtml(htmlContent, inputParser);
+    assert.strictEqual(expectedHtml, actualHtml);
```
abeforgit, line 67: qunit expects the first argument to be actual and the second to be expected
it makes a difference in the test reporting when it fails, so it's fairly important to get it right

abeforgit (changes requested):

addon/utils/ce/paste-handler-helper-functions/buildList.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (nextListLevel < listLevel) {
+      // Lower level found. Current list is done.
+      break;
+    }
+
+    if (nextListLevel > listLevel) {
+      const nestedList = buildList(nextSibling);
+
+      if (nestedList.list) {
+        listHtml += nestedList.list.innerHTML;
```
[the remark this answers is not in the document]
usrtim: yes, and removing lot of cleaners that we are not using currently.

tests/unit/utils/convert-ms-word-html-test.ts
```diff
@@ -0,0 +1,1240 @@
...
+  });
+
+  test('It should display nested list correctly as HTML', function (assert) {
+    const inputParser = new HTMLInputParser({});
+    const expectedHtml = oneLineTrim`
+      <ol>
+         <li><span>Test1</span></li>
```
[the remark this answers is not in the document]
usrtim: ok sure, do you mean adding spaces when It's nested? If yes, I can also do that if you want.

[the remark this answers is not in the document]
abeforgit: no, I mean making an actually nested html list https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#nesting_lists

see also my pr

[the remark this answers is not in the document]
usrtim: ahaa ok, thanks

addon/utils/ce/paste-handler-helper-functions/cleanDocx.ts
usrtim: yep, because I did lot of tests, and didn't see any impact to use other cleaners. We can return these without a problem, but just I don't think we will benefit anything doing that. If you want we can add any more tests, but on my testing experience everything seems to work without other cleaners as well.

abeforgit: but like, did you test a document with footnotes? with links, multiple fonts, etc?

abeforgit:
Ok so after thorough testing, here's what we lose from the original cleanDocx:
cleanLists: removes <a> tags that point to the current document. this is nice cause we end up with link tags that cannot be removed, that dont point anywhere useful either
cleanQuotes: transforms text marked as a quote into the html <blockquote> element. We remove this element anyway in our own cleaner.
cleanTextNodes: removes carriage returns and normalizes whitespace a bit. I think we do this ourselves as well
cleanFootnotes: removes some markup around footnotes, but the resulting paste doesn't look great anyway
cleanSpans: does some processing on runs of the space character. However it doesn't seem to have any effect.
I'll merge I guess, losing the link processing is a shame but not that big of a deal in the end
@nvdk input appreciated

PR #455, opened 2022-11-15
Inline components: serializable properties

elpoelma:
This PR introduces the following changes:
- The `props` and `state` object of inline components are merged into one. 
    * Properties can be written and read using the `setProperty` and `getProperty` methods.
- When defining an inline component spec, you need to define which properties it uses. For each of the properties you can defined whether it should be serialized and what default value it has (if it has one). This makes the DOM attributes less cluttered if only the properties are included that should be serialized.
    * E.g. properties that can be derived from other properties should not be serialized.
- Inline component matchers can now match an inline component on a dataset attribute instead of a class.
    * Inline components now contain a `data-inline-component` attribute with as value the name of the inline component.
- Two sample inline components are included in the editor dummy app to make testing easier.

### DIFF
--- a/addon/commands/insert-component-command.ts
+++ b/addon/commands/insert-component-command.ts
@@ -1,7 +1,6 @@
 import {
   ModelInlineComponent,
   Properties,
-  State,
 } from '../model/inline-components/model-inline-component';
 import Model from '../model/model';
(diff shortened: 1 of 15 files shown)


abeforgit (approved):
looks good

PR #499, opened 2022-12-07
fix(initialization): emit a selectionchanged after init

abeforgit:
selectionchanged events only emit when a new, different selection is set on the model. In the initialization phase, this never happens, since it's the first time the selection is set, so the selection is not "different". But plugins do need to recieve this initial trigger.

note: left the logger statements in intentionally, I think it's useful logging output (and we have a logging framework for a reason).

fixes https://binnenland.atlassian.net/browse/GN-3856


nvdk (approved):
fixes the issue with the template plugin not showing options on an empty document in GN :ok_hand: .

PR #526, opened 2023-01-10
Back to treating rdfa as nodes

abeforgit:
I'm really glad we had time to try using marks! In the end I think we can conclude it creates more problems than it solves.

This pr reverts the inline_rdfa from a mark back to a node and removes associated datastore logic and nodespecs that served as workarounds for invisible rdfa. It also includes adjusted list commands to make them work as expected (aka the reason we tried marks in the first place).

Still feel like there's room for improvement but at the moment we don't even really understand what works and what doesn't, so let's continue like this.

Other things that seem interesting to play with:
- normalize rdfa on initial load, and work with a clear distinction of subject/predicate/object nodes in the PM schema
- try a custom datastructure for storing the triplestore, one that's more aware of the document tree. PM's decoration tree could serve as inspiration there. There's still a dream of a purely rdf way of inserting content and knowledge there

### DIFF
--- a/addon/commands/list-commands.ts
+++ b/addon/commands/list-commands.ts
@@ -60,7 +60,6 @@ export function wrapInList(
       doJoin = true;
     }
     const wrap = findWrapping(outerRange, listType, attrs, range);
-    console.log('Wrapping found:', wrap);
     if (!wrap) {
       return false;
     }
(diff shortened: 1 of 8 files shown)


elpoelma (commented):
Seems to work well overall :smile: .
- Backspace around rdfa-nodes is still a bit weird, sometimes it removes the node completely
- Pressing return inside rdfa-nodes does not always work.

abeforgit:
thanks, will look into those before merging

abeforgit:
in the end we won't do this yet, but let's keep the branch for future reference

PR #531, opened 2023-01-12
Support for subscript and superscript marks

elpoelma:
Addition of a plugin which allows to mark text as subscript or superscript.


elpoelma:
@brenner-company I converted your plugin to a prosemirror-based plugin, thanks for the work!

PR #532, opened 2023-01-13
Addition of a menu which allows to insert headings

elpoelma:
This PR includes a new plugin which adds a toolbar widgets which allows for insertion of headings of different levels.

### DIFF
--- a/addon/commands/set-block-type.ts
+++ b/addon/commands/set-block-type.ts
@@ -1,0 +1,124 @@
+/**
+ *
+ * Modified from https://github.com/ProseMirror/prosemirror-transform and https://github.com/ProseMirror/prosemirror-commands
+ *
+ * 
+ * Copyright (C) 2015-2017 by Marijn Haverbeke <marijnh@gmail.com> and others
+
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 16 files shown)


addon/commands/set-block-type.ts
```diff
@@ -0,0 +1,124 @@
...
+import { PNode } from '..';
+
+/// Returns a command that tries to set the selected textblocks to the
+/// given node type with the given attributes.
+export function setBlockType(
+  nodeType: NodeType,
+  attrs: Attrs | null = null,
+  keepAttrs = false
+): Command {
+  return function (state, dispatch) {
```
abeforgit, line 39: Could you add a comment where you changed something from the PM implementation?

addon/components/plugins/headings/headings-menu.ts
```diff
@@ -0,0 +1,67 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { NodeType, ProseController } from '@lblod/ember-rdfa-editor';
+import { setBlockType } from '@lblod/ember-rdfa-editor/commands/set-block-type';
+
+type Args = {
+  controller: ProseController;
+};
+export default class HeadingsMenu extends Component<Args> {
```
abeforgit, line 9: The menu is janky when used in contexts where headings are not allowed, such as lists. Better to remove the menu when it is not applicable to be consistent with the other menus

abeforgit (changes requested):

PR #555, opened 2023-01-23
Rework set-html-content function to ensure no extra paragraphs are added

elpoelma:


addon/core/prosemirror.ts
```diff
@@ -250,14 +250,15 @@ export class ProseController {
     this.doCommand(selectAll);
```
abeforgit, line 250: I think then we don't need this anymore right?

abeforgit (changes requested):

addon/core/prosemirror.ts
```diff
@@ -250,14 +250,15 @@ export class ProseController {
...
     const domParser = new DOMParser();
-    tr.deleteSelection().insert(
+    tr.replaceWith(
       0,
+      tr.doc.nodeSize - 2,
       ProseParser.fromSchema(this.schema).parse(
         domParser.parseFromString(content, 'text/html')
       )
     );
+    tr.setSelection(Selection.atStart(tr.doc));
```
abeforgit, line 260: the original code set the selection at the end, not a huge deal but I think it's more in line with how our other insertions work

abeforgit (changes requested):

PR #556, opened 2023-01-24
add recreate uuids plugin to prosemirror

lagartoverde:


elpoelma (commented):
Can you add the `recreateUri` and `uriAttributes` to the args of the ember-node utility functions?

elpoelma (commented):
Some suggestions:
- This algorithm can maybe be extended to marks?
- Maybe the nodes themselves can provide how to regenerate attributes? That way it is a bit more generic and can be easily extended to other type of attributes (not only uris).

abeforgit:
agree with elena's comments, good otherwise

PR #568, opened 2023-01-30
Upgrade ember-source to 4.8.4

elpoelma:


abeforgit:
make sure we still support 3.28

elpoelma:
> make sure we still support 3.28

So it seems ember try:each correctly detected features that are not yet supported in 3.28.

abeforgit (changes requested):
- the ember type packages probably need a major version bump, iirc they have linked major versions to the source
- we should update the CONTRIBUTING.md file
- let's drop node 14 & 16 support, and set the volta version to 18
- please add this to the release notes

PR #590, opened 2023-02-03
Addition of deprecations which will be removed in version 3.0.

elpoelma:
This PR adds deprecations for the following methods, classes and types:
- `ProseController.view`, use `ProseController.getView` instead.
- `ProseController.state`, use `ProseController.getState` instead.
- `ProseController.checkAndDoCommand`, use `ProseController.doCommand` instead.
- `NodeConfig`, `MarkConfig`, `PluginConfig`, `ResolvedPluginConfig` and `RdfaEditorPlugin`
- string-based version of `toggleMark`, use the `MarkType`-based version instead

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,7 +10,7 @@ module.exports = {
       legacyDecorators: true,
     },
   },
-  plugins: ['ember', '@typescript-eslint'],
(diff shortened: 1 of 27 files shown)


abeforgit:
before we merge this I want to doublecheck reactivity of the getter functions

PR #599, opened 2023-02-08
Major rework of editor widgets and components

elpoelma:
This PR includes a major rework on how widgets are defined and passed to the editor component:
- Widgets are no longer passed using the `widget` component property, they are included in the template as children of the editor-component.
- The editor-component has been renamed from `Rdfa::RdfaEditor` to simply `Editor`.
- The `Editor` component now yields two different named blocks: `top` and `aside`. In these blocks, it is possible to include toolbars or sidebars.
- Seperate `Toolbar` and `Sidebar` components have been added. `Toolbar` yields a `left` and `right` block, while `Sidebar` yields a `collapsible` and `main` block. 
- Seperate components representing the toolbar widgets have been added, these can be included in a `Toolbar` component.
- The `RdfaEditorWithDebug` has been removed in favor for the `DebugTools` component. You can add this component to your template above the `Editor` component. It expects a `controller` argument of type `ProseController`.
- Several unused `editorOptions` have been removed.

This PR is currently a draft, it may include modifications in the future.

This is a solution to https://binnenland.atlassian.net/browse/GN-4002?atlOrigin=eyJpIjoiYzI4MDJkN2U5ZTk2NDkwNDgzNzEwMGMxZDM3MTgwNzEiLCJwIjoiaiJ9.


addon/components/sidebar.hbs
```diff
@@ -0,0 +1,54 @@
...
+              />
+              <span class='au-u-hidden-visually'>
+                Sluit accordion
+              </span>
+            {{/if}}
+          </Group>
+        </AuToolbar>
+        <div class={{if this.expanded 'au-c-accordion__content' 'au-u-hidden'}}>
+          <AuList @divider={{true}}>
+            {{yield to="collapsible"}}
```
abeforgit, line 45: I'd yield this accordion list as a contextual component here, so you can have multiple collapsible sections

addon/components/sidebar.hbs
```diff
@@ -0,0 +1,54 @@
...
+          </AuList>
+        </div>
+      </div>
+    </div>
+  {{/if}}
+  <ul class='say-editor-hints__list'>
+    {{yield to="main"}}
```
abeforgit, line 52: might as well make this the default block

abeforgit (commented):
just some initial thoughts that popped up

addon/components/sidebar.hbs
elpoelma: Good idea, I'm wondering if we convert all yielded named blocks to contextual components to make it consistent.

abeforgit:
Changlog pls

elpoelma:
> Changelog pls

I'll extend it.

PR #600, opened 2023-02-08
Handle ctrl click on links

lagartoverde:
Needs testing on windows and linux, I added the meta key so it works as expected on mac

### DIFF
--- a/addon/plugins/links/index.ts
+++ b/addon/plugins/links/index.ts
@@ -11,8 +11,8 @@ export const linkHandler: ProsePlugin = new ProsePlugin({
       const schema = view.state.schema;
       // The handler only handles clicks on nodes, and the link is a mark of the textnode inside the node, so we have to get it
       const textNode = node.childAfter(pos);
-      if (!textNode || !textNode.node) { 
-        return; 
+      if (!textNode || !textNode.node) {
+        return;
       }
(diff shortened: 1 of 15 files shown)


elpoelma (approved):
Seems to work well!

addon/plugins/links/index.ts
```diff
@@ -0,0 +1,25 @@
...
+      view: RdfaEditorView,
+      pos: number,
+      node: PNode,
+      nodePos: number,
+      event: MouseEvent
+    ) {
+      const schema = view.state.schema;
+      // The handler only handles clicks on nodes, and the link is a mark of the textnode inside the node, so we have to get it
+      const textNode = node.childAfter(pos);
+      if (!textNode || !textNode.node) return;
```
abeforgit, line 14: I know I know it's pedantic, and if you disagree you can merge without applying in this case (cause its really _just_ a return)

```suggestion
      if (!textNode || !textNode.node) { 
        return; 
      }
```

abeforgit (changes requested):
Please add a changelog entry, then you can merge

addon/plugins/links/index.ts
lagartoverde: I prefer your syntax, I don't know what happened :), thanks for catching that

PR #607, opened 2023-02-10
Improved link support

elpoelma:
This PR adds improved link supported to the editor. It includes:
- a new `link` node and `linkView` nodeView: an interactive ember-node which you can use to edit the text and underlying url of a link.
   * The `link` node has two variants: an interactive and a non-interactive one. You can configure whether a link is interactive by passing the `interactive` option. When an interactive link node is selected, a little in-editor popup is show to edit the link.
- Additionally, if using the non-interactive link node, you can also use the `Sidebar::LinkEditor` component to edit a link when it is selected.
- A `Toolbar::InsertLink` component is added to ensure users can insert new links into the editor.

Note: this PR does not remove the previously used `link` mark.

This PR solves https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?modal=detail&selectedIssue=GN-1169.


abeforgit (changes requested):
- too much padding: 
image
- pressing the "insert link" button with text selected -> I expect the selected text to become the link text
- we need design feedback in general, the pill looks nice but we're starting to have a lot of similar looking pills. However, that's not blocking to merging this, we can discuss that when we start implementing it in GN

elpoelma:
> * too much padding:
>       image
> 
>     * pressing the "insert link" button with text selected -> I expect the selected text to become the link text
> 
>     * we need design feedback in general, the pill looks nice but we're starting to have a lot of similar looking pills. However, that's not blocking to merging this, we can discuss that when we start implementing it in GN

I reduced the minimum length of links and implemented the functionality so that selected text becomes the link text when inserting one. I agree with the design feedback.

abeforgit (changes requested):
looks good, just one more thing:
- after inserting a link with the toolbar, make sure the selection is inside the link so the link UI is active

elpoelma:
> looks good, just one more thing:
> 
>     * after inserting a link with the toolbar, make sure the selection is inside the link so the link UI is active

Should be implemented now.

elpoelma:
Updating this PR so it includes the changes of the base branch caused some issues with the focusing in the link nodes, I'll investigate further.

abeforgit (changes requested):
I would like to solve the issue with links (or any inline pills) at the start of a line preventing you from clicking to the left of it before merging

elpoelma:
> I would like to solve the issue with links (or any inline pills) at the start of a line preventing you from clicking to the left of it before merging

This issue does not occur in chrome, only in firefox atm. The issue is in the gecko engine and is logged here: https://bugzilla.mozilla.org/show_bug.cgi?id=1612076

abeforgit:
Unfortunately the "detach link" feature is now broken

elpoelma:
> Unfortunately the "detach link" feature is now broken

Fixed it, I was applying the removal of the link on the active editor-view.

PR #608, opened 2023-02-10
Modify the base keymaps to better support mac ones

lagartoverde:
Solves https://binnenland.atlassian.net/browse/GN-4013, please test if the behaviour is unaltered in Windows and Linux as this should only make a difference to Mac users


elpoelma (approved):
Can't speak for mac and windows of course, but on linux the shortcuts still work as expected :). Could you just update the changelog? Then you can merge :)

PR #611, opened 2023-02-13
Rework link handler a bit to prevent range errors

elpoelma:


elpoelma:
Not adding a changelog entry as the ctrl+click plugin has not been released yet.

PR #613, opened 2023-02-15
Increase left-padding of lists

elpoelma:
Sets the padding-left of lists to 1.25rem.

Solves https://binnenland.atlassian.net/browse/GN-4051?atlOrigin=eyJpIjoiMDgyMzdhNjZiYzY2NDhlZWEwMGI0Y2Y4NGI5YWNjNzQiLCJwIjoiaiJ9


abeforgit (changes requested):
The way I understand it they want even more indentation.
This is how a nested list looks like without any css:
image
google docs:
image
word online:
image

`3rem` seems like a decent value for us

PR #614, opened 2023-02-15
`Heading` menu dropdown is not fully visible on small editor views

usrtim:


usrtim:
This PR will fix the problem with hidden overflow , I tested it and everything looks fine but I’m not sure if there will be any other side-effects.

elpoelma (changes requested):
I'm afraid this introduces another issue: on smaller displays the toolbar is not fixed at the top any longer. Additionally, the heading menu should also be fully visible on PC-sized displays but where the editor instance itself does not take up the whole page, so for that case, the @media css selector won't work as expected.

abeforgit:
Ok so
It seems you misunderstood the problem. The issue is that if the div container the editor has a limited height, the headings menu is cut off. Media queries are not relevant here since it has nothing to do with screen size.

usrtim:
> Ok so It seems you misunderstood the problem. The issue is that if the div container the editor has a limited height, the headings menu is cut off. Media queries are not relevant here since it has nothing to do with screen size.

@abeforgit as we agreed, the solution here was to remove overflow: hidden from main container. Please have a look in your spare time.

PR #616, opened 2023-02-16
disabled undo and redo button when there's no more steps

lagartoverde:


elpoelma (approved):
Looks good, maybe you can also set the cursor to 'not-allowed' when a toolbar button is disabled.

PR #618, opened 2023-02-16
Addition of margin on elements below paragraphs and tables

elpoelma:
This PR ensures that there is spacing added between paragraphs and other elements. Solves https://binnenland.atlassian.net/browse/GN-4060?atlOrigin=eyJpIjoiY2M5NDhkYmE0ZWU1NGYzYTk0YTM5YmJkNWM0MTJhOTMiLCJwIjoiaiJ9.


abeforgit:
this adds quite a bit of margin to list items as well, which is maybe a bit too much?
also, there is no margin for tables:
image

PR #620, opened 2023-02-17
Support for ordered list styles

elpoelma:
This PR mainly adds support for ordered list styles. A dropdown button is added to the order-list style menu so that users can select another style. Currently the following list-styles are included:
- decimal
- lower-alpha
- upper-roman

These styles are implemented by adding the custom `data-list-style` to ordered list styles which is picked up by the css. Another possibility is implementing this using inline styles. 

It should not be too difficult to make this list configurable using a component property if this would be desirable in the future.

Full list of changes:
- Renamed Toolbar::Icon to Toolbar::Button: this change seemed to made sense in a semantic way
- Addition of an optional `options` dropdown to Toolbar::Button. If a Toolbar::Button needs to support additional options included in a dropdown, they can be added in the optional `<:options>` named block.
- Add support for optional attributes in the `toggleList` command.

This PR is a solution to https://binnenland.atlassian.net/browse/GN-4052?atlOrigin=eyJpIjoiZWRmNGU1MGZmMmUzNDgxOWJiNTU0MWNjYzY0OTQ0Y2YiLCJwIjoiaiJ9.

It's possible that this PR needs some more design work.


lagartoverde (approved):
Looks good, maybe something a little bit nitpicky is that the arrow for the dropdown is very small and hard to click, maybe we should look for another design pattern for that

elpoelma:
> Looks good, maybe something a little bit nitpicky is that the arrow for the dropdown is very small and hard to click, maybe we should look for another design pattern for that

Yes that's fair, I now based it on the behaviour of other editors, but we can look into alternatives.

abeforgit (changes requested):
I think people probably want to keep the current list style going when they indent to a new level, bit annoying to make a deep alphabetic list now for example

PR #621, opened 2023-02-17
Added features to add table headings

lagartoverde:
I don't really like the toggle header column feature, I left it so you can test it, but I think we can leave only the toggle header row command.
Also need icons, we will have to ask designers


elpoelma (changes requested):
Functionality wise, it seems to work as expected. One thing to note: when enabling a column as a header, the seperators between the columns dissapear.
Additionally we should also look into what happens when you toggle a row as a header when one of the row cells is already part of a column header.

Before merging this, the icons should also be updated of course.

abeforgit:
> Additionally we should also look into what happens when you toggle a row as a header when one of the row cells is already part of a column header.

Saying we should look into something pretty much guarantees we won't :) 
To make this a bit more concrete: the current implementation is inconsistent with how our mark logic works, that is:
- if a selection contains some text that has a mark and some that doesn't, we say "the selection does not have that mark"
- this implies that when you toggle that mark, it toggles to "on" for the selection

this is actually modified already from prosemirror default behavior, which considers marks to be "on" in the case described above, making the toggle in that instance go to "off"

Something quite similar is happening here with the table commands. Prosemirror is consistent, so it says "if one of the cells in the row is a `<th>`, it is a header row, meaning the toggle will remove the `<th>`. 
If we want to be consistent in the same way (and I agree that we should be), we should flip that logic.

So, tldr: please make it so that if you toggle header row in a row where one of the cells is already a `<th>`, it toggles "on" instead of "off" as it does now, and same thing for columns

to do this you'll likely have to provide modified versions of the prosemirror commands. See `toggle-mark-add-first.ts` for how we did that with marks.

abeforgit:
https://github.com/appuniversum/ember-appuniversum/issues/382

PR #643, opened 2023-02-22
Addition of paragraph/heading indentation

elpoelma:
This PR adds the option to indent paragraph and heading nodes which are not located inside a list. Currently it is only possible to increase/reduce indentation using the corresponding toolbar buttons. We should determine whether using the tab key is an adequate shortcut for indentation.

Right now, the editor supports 4 levels of indentation, with each level carrying an indentation increase of 3rem.

The toolbar menu containing the indentation buttons is located inside the `Plugins::Indentation::IndentationMenu` component. This component also includes the functionality for list indentation. The previously used `Plugins::List::IndentationControls` component has been deprecated.

This PR is a solution to https://binnenland.atlassian.net/browse/GN-4062?atlOrigin=eyJpIjoiNGNlYmNlMmYxOTMwNGJhNGJlZTEwM2VkMjgzMDgyZjgiLCJwIjoiaiJ9


lagartoverde (changes requested):
2 Things: 
I would change the order of the unindent/indent button, this way is more logical to me and allows you to ident several times quicker.
If you indent some paragraphs a few times and then change them to a list it keeps the identation but the list is not indented

abeforgit (changes requested):
I agree with oscar about the indented paragraph in a list. Wrapping something in a list should just remove all prior indentation. 
I don't agree about changing the order of the indentation buttons. I think it's confusing now because the buttons hide when they are not applicable. I think you should make the entire indentation section always show, and just disable the buttons that are not applicable.

If we do want to keep the autohiding, we should hide the indentation section as one.

PR #666, opened 2023-02-28
Add `linkPasteHandler` plugin which detects and converts links in pasted content

elpoelma:
- Addition of a `linkPasteHandler` plugin which detects plain text links inside pasted content and converts them to a link node or adds a link mark.
- Use the new view `pasteHTML` method inside the Word paste handler.

Solution to https://binnenland.atlassian.net/browse/GN-4121?atlOrigin=eyJpIjoiMDM3NzdkM2E3ZTk2NDQxN2FjMWRkNjY3OGNiZDhlNjkiLCJwIjoiaiJ9.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Addition of linkPasteHandler plugin which detects and converts links in pasted content
 
+### Changed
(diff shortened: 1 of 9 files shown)


abeforgit:
Side effect: if I disconnect a link in the editor, then copy and paste it, it's a link again
rather unlikely to be a problem

abeforgit:
it also re-linkifies when dragging text

PR #668, opened 2023-02-28
Add toggle component which allows showing paragraph endings as pilcrows

elpoelma:
Add toggle component which allows showing additional formatting:
- Paragraph endings displayed as pilcrows (¶)
- Spaces as dots
- Returns as ⏎

This is implemented using the ` @guardian/prosemirror-invisibles` library (https://github.com/guardian/prosemirror-invisibles).

Disadvantages of this approach:
- Additional elements (decorations) instead of css only
- Firefox does not handle cursor behaviour well around `contentEditable: false` elements (such as decorations).

Solution to https://binnenland.atlassian.net/browse/GN-4066?atlOrigin=eyJpIjoiOGMxYmIyODkxOWE0NGJlNzg2NTgwZmJjOTAwOTc1MTEiLCJwIjoiaiJ9

### DIFF
--- a/app/styles/ember-rdfa-editor/_c-formatting-marks.scss
+++ b/app/styles/ember-rdfa-editor/_c-formatting-marks.scss
@@ -28,7 +28,7 @@
       z-index: 1;
     }
     &.invisible--space::before {
-      content: '.';
+      content: '\00B7';
     }
     &.invisible--break::before {
       content: '\23CE';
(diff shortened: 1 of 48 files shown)


app/styles/ember-rdfa-editor/_c-formatting-marks.scss
```diff
@@ -0,0 +1,43 @@
...
+      font-weight: 400;
+      font-style: normal;
+      line-height: initial;
+      width: 0;
+      top: 0;
+      left: 0;
+      z-index: 1;
+    }
+    &.invisible--space::before {
+      content: '.';
```
abeforgit, line 31:
```suggestion
      content: '\00B7';
```

https://www.compart.com/en/unicode/U+00B7

is a bit easier to distinguish from a real period

abeforgit (changes requested):

abeforgit (changes requested):
the marks don't work inside links and variables

PR #670, opened 2023-03-01
GN-4123: Inherit ordered list style correctly

dkozickis:
* Give `decimal` style by default when pressing on the "ordered list" menu button, without opening the button
* Fixup CSS 

### Before

https://user-images.githubusercontent.com/769698/222077169-a26d119b-ddfa-47d5-bd98-a357b09e2b2e.mp4

### After

https://user-images.githubusercontent.com/769698/222077191-c202887f-c49c-4f0f-bf5f-77c1897e1058.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Inherit ordered list style correctly 
+
 ## [3.2.1] - 2023-02-28
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


elpoelma (changes requested):
The first implementation of inheriting ordered list-styles was also done using the css `inherit` value. The issue here is that we can't rely on it as it forms issues when having an ordered list as a sublist of an unordered list (the ordered list will inherit the bullet-point-style of the unordered list).

elpoelma:
> > The first implementation of inheriting ordered list-styles was also done using the css `inherit` value. The issue here is that we can't rely on it as it forms issues when having an ordered list as a sublist of an unordered list (the ordered list will inherit the bullet-point-style of the unordered list).
> 
> I was not able to figure out how to update `toggleList` to assing `data-list-style` to `decimal` by default, when switching from unordered list to ordered, so I figures out a CSS fix in the end.

This solution seems fine to me!

elpoelma (approved):

PR #672, opened 2023-03-01
Disable toolbar buttons when not applicable instead of hiding them

elpoelma:
Solves https://binnenland.atlassian.net/browse/GN-4119?atlOrigin=eyJpIjoiNTAyZThmNTg5OTE0NGU1ODkwMTgzNTY0MWUyMTRlYzgiLCJwIjoiaiJ9.

Probably needs some design improvements on how disabled buttons are visualized as it looks a bit weird now. Maybe a better approach is to keep the same color on each toolbar button but only change the color on hover on enabled buttons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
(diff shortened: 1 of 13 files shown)


abeforgit:
image
it already looks better with the button border style removed imo

abeforgit:
image
even better if we set the background color of the disabled state to the toolbar grey

elpoelma:
> image even better if we set the background color of the disabled state to the toolbar grey

Indeed, I updated the css accordingly.

PR #678, opened 2023-03-02
text color plugin - allows user to change the color of the text

usrtim:
[screen-recorder-thu-mar-02-2023-11-00-41.webm](https://user-images.githubusercontent.com/115715476/222396425-a506370e-c7a7-4886-9341-a4b060e7a7a8.webm)

### DIFF
--- a/addon/components/plugins/text-style/color.ts
+++ b/addon/components/plugins/text-style/color.ts
@@ -13,25 +13,32 @@ type Args = {
 };
 export default class Color extends Component<Args> {
   @service declare intl: IntlService;
-  colors = ['black', 'red', 'green', 'blue', 'orange'];
-  @tracked selectedColor = this.colors[0];
+  @tracked selectedColor = 'black';
 
   get controller() {
     return this.args.controller;
   }
 
   get currentStyle() {
-    this.setActiveColorName();
-    const activeColor = this.selectedColor;
-
-    if (activeColor) {
-      return `${this.intl.t('ember-rdfa-editor.color')} ${activeColor}`;
-    } else {
-      return this.intl.t('ember-rdfa-editor.default-color');
-    }
+    return this.setActiveColor();
   }
 
   @action
+  onDone(selectedColorInfo: { _rgba: number[] }) {
+    const rgba = selectedColorInfo._rgba;
+    let color = 'rgba(';
+
+    rgba.forEach((code, index) => {
+      if (index === rgba.length - 1) {
+        color += ` ${code})`;
+      } else {
+        color += ` ${code},`;
+      }
+    });
+
+    this.setColorMark(color);
+  }
+
   setColorMark(color: string) {
     if (this.controller) {
       const state = this.controller.mainEditorState;
@@ -56,10 +63,12 @@ export default class Color extends Component<Args> {
           return tr.addMark($from.pos, $to.pos, mark);
         });
       }
+
+      this.controller.focus();
     }
   }
 
-  setActiveColorName = () => {
+  setActiveColor() {
     if (this.controller) {
       const state = this.controller.mainEditorState;
       const { selection } = state;
@@ -94,5 +103,5 @@ export default class Color extends Component<Args> {
     }
 
     return null;
-  };
+  }
 }
--- a/addon/components/plugins/text-style/color.ts
+++ b/addon/components/plugins/text-style/color.ts
@@ -5,6 +5,8 @@ import { SayController } from '@lblod/ember-rdfa-editor';
 
 import IntlService from 'ember-intl/services/intl';
 import { TextSelection } from 'prosemirror-state';
+import { Option } from '@lblod/ember-rdfa-editor/utils/_private/option';
+import { tracked } from '@glimmer/tracking';
 
 type Args = {
   controller?: SayController;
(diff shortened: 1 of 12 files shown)


abeforgit (changes requested):
- throws a range error with an empty document
- type 1 black char, change color, type another char -> color menu doesn't update
- color menu should show selected color after you've selected a color, not only when you start typing

usrtim:
> * throws a range error with an empty document
> 
>     * type 1 black char, change color, type another char -> color menu doesn't update
> 
>     * color menu should show selected color after you've selected a color, not only when you start typing

@abeforgit fixed all of these, tested more than 3 hours. I hope I didn't miss anything here. Also because this is my first time working on editor after prosemirror upgrade, please let me know if I could do things more simpler from code perspective. Thanks

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+        });
+      }
+    }
+  }
+
+  colorIsActive = (color: string) => {
+    return this.isActive() === color;
+  };
+
+  isActive = () => {
```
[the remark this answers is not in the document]
abeforgit: Additionally, the standard way in ember to do derived state is using getters. I recommend reading through this page: https://guides.emberjs.com/v4.8.0/components/component-state-and-actions/

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        const color = Object.values(node?.marks || {}).find((item) => {
+          return item.type.name === 'color';
+        })?.attrs?.color as string;
+
+        return color || 'black';
+      }
+    }
+
+    return '';
```
abeforgit, line 96: use `null` for cases where there is no sensible value to return.

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+export default class Color extends Component<Args> {
+  @service declare intl: IntlService;
+  colors = ['black', 'red', 'green', 'blue', 'orange'];
+
+  get controller() {
+    return this.args.controller;
+  }
+
+  get currentStyle() {
+    const currentColor = this.colors.find(this.colorIsActive);
```
abeforgit, line 21: you calculate the active color in the document
then you loop over the color array, comparing each color to the active color, to find... the active color?

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+      return this.intl.t('ember-rdfa-editor.default-color');
+    }
+  }
+
+  @action
+  enableColor(color: string) {
+    if (this.controller) {
+      const state = this.controller.mainEditorState;
+      const { selection } = state;
+      const { $from, $to, $cursor } = selection as TextSelection;
```
abeforgit, line 34: rather than casting, use the instanceof check you used in the isActive function

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  isActive = () => {
+    if (this.controller) {
+      const state = this.controller.mainEditorState;
+      const { selection } = state;
+      const marks = state.storedMarks;
+
+      if (marks?.length) {
+        const color = Object.values(marks).find((item) => {
```
abeforgit, line 68: use https://prosemirror.net/docs/ref/#model.MarkType.isInSet (also on line 68)

translations/nl-BE.yaml
```diff
@@ -33,6 +33,8 @@ ember-rdfa-editor:
   normalText: Gewone tekst
   heading: Koptitel
   insert-heading: Voeg koptitel toe
+  color: Color
```
abeforgit, line 36: Tekstkleur

translations/nl-BE.yaml
```diff
@@ -33,6 +33,8 @@ ember-rdfa-editor:
...
+  default-color: Color Black
```
abeforgit, line 37: Tekstkleur zwart

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+          return item.type.name === 'color';
+        })?.attrs?.color as string;
```
abeforgit, line 70: if you do cast: cast correctly. As you see in line 72, color is _not_ guaranteed to be a string, it can also be undefined! You can express that type as `string | undefined` or, even better, `Option<string>` .

addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,98 @@
...
+      const { selection } = state;
+      const { $from, $to, $cursor } = selection as TextSelection;
+      const markType = state.schema.marks.color;
+
+      const mark = markType.create({ color });
+      if (markType && $cursor) {
+        if (markType.isInSet(state.storedMarks || $cursor.marks())) {
+          this.controller.withTransaction((tr) => {
+            tr.removeStoredMark(markType);
+            return tr.addStoredMark(markType.create({ color }));
```
abeforgit, line 42: you defined this as a constant on line 37, why not use it? (same thing on line 46)

addon/components/plugins/text-style/color.ts
abeforgit: (and yes, I'd argue the name of that function isn't the best, for the same reason I don't like `isActive`)

abeforgit (changes requested):

addon/components/plugins/text-style/color.ts
usrtim: yes, you are right. In this commit I was mainly interested if functionality is working as expected, and for that reason forget this. You are right, doesn't make sense, sorry for that.

addon/components/plugins/text-style/color.ts
usrtim: If I remove casting from here, It will give me **"Property '$cursor' does not exist on type 'Selection'."** Also I noticed that every time we are using **$cursor** from selection, we are doing this type of casting.

usrtim:
@abeforgit please have a look if you have time. If you have any comments, I will fix that on weekend. During the functionality, I really tested it in all possible cases that I could think of, and everything seems to work fine.

usrtim:
@abeforgit I think we can close this PR because we have implemented new functionality here: https://github.com/lblod/ember-rdfa-editor/pull/715

PR #684, opened 2023-03-03
Add color highlighting toolbar button and mark

elpoelma:
This PR adds color highlighting functionality. I provides a `highlight` mark and a toolbar menu which allows users to highlight text. 

The toolbar menu consists of two parts:
- A button which allows users to quickly highlight text in a document
- An `options` button which allows users to select a different highlighting color using a color picker.

The toolbar menu differs a bit from the text styling menus as it does not show which highlight is applied to the currently selected text.


lagartoverde (changes requested):
I like it, just one thing I would change is I would add several predetermined color options first, and the a button to choose a custom color, but giving the user first the color picker might be a little bit overwhelming.
Just as google docs does, maybe with not so many color options xd
Screen Shot 2023-03-03 at 13 24 57
Screen Shot 2023-03-03 at 13 25 02
Might need also design input on this one, but it's already so cool

elpoelma:
> I like it, just one thing I would change is I would add several predetermined color options first, and the a button to choose a custom color, but giving the user first the color picker might be a little bit overwhelming. Just as google docs does, maybe with not so many color options xd Screen Shot 2023-03-03 at 13 24 57 Screen Shot 2023-03-03 at 13 25 02 Might need also design input on this one, but it's already so cool

I agree, the only issue is that I have limited control over the behaviour and visualization of the color picker. As it uses the standard `<input type='color'/>` tag. You'll also notice it looks a behaves a bit different in other browsers (such as firefox).

lagartoverde:
Yes but we can make a dropdown like @usrtim did, with some default color options, and then only open the color picker on the custom option

elpoelma:
> Yes but we can make a dropdown like @usrtim did, with some default color options, and then only open the color picker on the custom option

That's possible, I also just noticed that you can pass a list of predefined colors to an `<input type='color'>` element. I will first show the user those in combination with a button to open the colorpicker.

usrtim (approved):
LGTM

lagartoverde (approved):
https://user-images.githubusercontent.com/17006446/223693686-1b085ff4-0ab8-4351-a249-9cb6fe03816d.mov

Looks much better to me, now I have a small problem, when selecting text, we replace the background color, might be a design problem to solve this, but I don't think is very important. For me you can merge without it

addon/components/plugins/text-style/highlight.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this.controller.schema;
+  }
+
+  get state() {
+    return this.controller.activeEditorState;
+  }
+
+  @action
+  setHighlight(color: string) {
+    const mark = this.schema.marks.highlight.create({ value: color });
```
abeforgit, line 28: would like to see this wrapped in a command, so users of custom buttons can easily replicate the behavior

abeforgit (changes requested):
Something I don't like: there's no way to "remove" a highlight. You can set the color to white, which works fine as long as the background is white. But if the doc is ever rendered on a non-white background, the inline styles will still show the white markings.

addon/components/plugins/text-style/highlight.ts
elpoelma: All right, I moved the functionality into a command. I'll now look into the ability to remove highlights.

elpoelma:
> Something I don't like: there's no way to "remove" a highlight. You can set the color to white, which works fine as long as the background is white. But if the doc is ever rendered on a non-white background, the inline styles will still show the white markings.

I added a reusable color-selector component which allows you to either remove the current color, select a preset color or open the color picker.

PR #685, opened 2023-03-03
Ensure the list toggle button removes selected content completely out of all lists when untoggling

elpoelma:
This PR changes the behavior of the list toggle buttons slightly: when untoggling, instead of unindenting, the button now takes the selected content out of all nested lists.

Solves https://binnenland.atlassian.net/browse/GN-4096?atlOrigin=eyJpIjoiZWU5YWMxOGJjM2VmNDhjZDhiNzRhNzAxNWNkZDlmYjQiLCJwIjoiaiJ9.


elpoelma:
> in the following structure: image the list button is unresponsive when standing in `oneonetwo`

Nice catch, my latest commit already improves that behaviour by a bit, but there are still some issues with it.

abeforgit:
let me know when those are resolved

elpoelma:
> let me know when those are resolved

@abeforgit it should work better now.

PR #689, opened 2023-03-06
GN-4123: Fix switching unordered list to ordered

dkozickis:
Previously `data-list-style` was not applied correctly when switching from an unordered list to an ordered list, because we were not passing "new" attributes along with the existing attributes of the node, and because of that `data-list-style` was not getting set. This PR fixes that 🙏 

I did a sanity "regression" check on known cases (e.g. unordered->ordered->unordered) and it seems to be working correctly.

The good old "look for 4 hours for a one-line fix" 😅 d

### DIFF
--- a/addon/components/plugins/list/ordered.ts
+++ b/addon/components/plugins/list/ordered.ts
@@ -3,7 +3,10 @@ import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 import IntlService from 'ember-intl/services/intl';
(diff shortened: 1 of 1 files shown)


elpoelma:
> > Looks good, nice catch inside the `toggle-list` command. I would just ensure that if no `data-list-style` is present and the `ol` is not a child of another list, the default list-style-type is set as `decimal`.
> 
> I think we are safe with the fact that "data-list-style" is now always set on an ordered list. I was not able to find a corner case where it is not set thinking

It can happen on older documents for example.

elpoelma (approved):

PR #693, opened 2023-03-07
GN-4117: Correct cursor placement in ember node on select

dkozickis:
Looking for confirmation whether the plugin approach is correct?

https://user-images.githubusercontent.com/769698/223392664-5ecc4f93-27fe-4d44-8605-abb67dc7eeeb.mp4

### DIFF
--- a/addon/components/ember-node/embedded-editor.ts
+++ b/addon/components/ember-node/embedded-editor.ts
@@ -90,6 +90,10 @@ export default class EmbeddedEditor extends Component<Args> {
     return this.args.getPos();
   }
 
+  get controller() {
+    return this.args.controller;
+  }
+
   get schema() {
     return new Schema({
       nodes: {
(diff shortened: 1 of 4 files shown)


elpoelma:
> Looking for confirmation whether the plugin approach is correct?

This is what we want indeed.

addon/components/ember-node/embedded-editor.ts
```diff
@@ -203,11 +204,21 @@ export default class EmbeddedEditor extends Component<Args> {
   @action
   onSelected() {
     if (this.args.selected && this.innerView) {
+      const lastKeyPressedPluginState = lastKeyPressedPluginKey.getState(
+        this.outerView.state
+      );
```
elpoelma, line 209: I think it's better to use `this.controller.mainEditorState` here. (To ensure this will also work with multi-level nested editors in the future).

elpoelma (changes requested):
Seems to work well at first glance :smile:. I just have some comments. As an alternative you could also look into the history plugin.

addon/components/ember-node/embedded-editor.ts
elpoelma: Yes

abeforgit:
Ok for me once comments resolved

elpoelma (approved):

PR #695, opened 2023-03-08
Addition of a responsive toolbar

elpoelma:
This PR introduces the `ResponsiveToolbar` component, a modified version of the current `Toolbar` component which takes screen and container widths into account. 
The new toolbar component consists of two main sections:
- A `main` section which will typically contain often used controls such as text styling controls, table controls, history controls etc.
- A `side` section which can include additional controls. It will typically contain additional information and formatting controls.

Both of these sections are collapsible. If the controls of a section no longer fit in the available space, they are moved to a collapsible dropdown. The toolbar will first try to collapse the `side` section, and afterwards the `main` section

The component should be used in the following manner:

```
<ResponsiveToolbar>
<:main>
{{!-- A list of main controls --}}
</:main>
<:side>
{{!-- A list of side controls --}}
</:side>
<ResponsiveToolbar>
```

The toolbar will collapse each direct child of the `main` and `side` sections separately. If you do not wish a specific child to be collapsed, you may add the `data-ignore-resize` attribute.

This PR is a solution to https://binnenland.atlassian.net/browse/GN-4120?atlOrigin=eyJpIjoiYzk4MDcxZmNhZTJjNDJiNjkxYjQwOTM4YTNmYjFiNmIiLCJwIjoiaiJ9.

### DIFF
--- a/addon/components/responsive-toolbar.ts
+++ b/addon/components/responsive-toolbar.ts
@@ -3,25 +3,22 @@ import Component from '@glimmer/component';
 import { modifier } from 'ember-modifier';
 import { tracked } from 'tracked-built-ins';
 
+type ToolbarSection = {
+  reference?: HTMLElement;
+  dropdown?: HTMLElement;
+  enableDropdown: boolean;
+  showDropdown: boolean;
+};
+
 export default class ResponsiveToolbar extends Component {
   toolbar?: HTMLElement;
 
-  main: {
-    toolbar?: HTMLElement;
-    dropdown?: HTMLElement;
-    enableDropdown: boolean;
-    showDropdown: boolean;
-  } = tracked({
+  main: ToolbarSection = tracked({
     enableDropdown: false,
     showDropdown: false,
   });
 
-  side: {
-    toolbar?: HTMLElement;
-    dropdown?: HTMLElement;
-    enableDropdown: boolean;
-    showDropdown: boolean;
-  } = tracked({
+  side: ToolbarSection = tracked({
     enableDropdown: false,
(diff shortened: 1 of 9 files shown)


lagartoverde (approved):
Looks good to me :)

addon/components/responsive-toolbar.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  @action
+  toggleSideDropdown() {
+    this.side.showDropdown = !this.side.showDropdown;
+    if (this.side.showDropdown) {
+      this.main.showDropdown = false;
+    }
+  }
+
+  isOverflowing = () => {
```
abeforgit, line 90: any reason this is not a getter? (if there is, would like a code comment explaining why)

addon/components/responsive-toolbar.ts
```diff
@@ -0,0 +1,148 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { modifier } from 'ember-modifier';
+import { tracked } from 'tracked-built-ins';
+
+export default class ResponsiveToolbar extends Component {
+  toolbar?: HTMLElement;
+
+  main: {
```
abeforgit, line 9: would like to see a named interface here

abeforgit (changes requested):
works really nicely, good work. Comments are nitpicks. Feel free to merge after adressing

addon/components/responsive-toolbar.ts
elpoelma: Not really :stuck_out_tongue:

PR #709, opened 2023-03-14
GN-4131: Un-indent paragraph on backspace at the front of paragraph

dkozickis:
Un-indent paragraph on backspace at the front of paragraph (empty or with content)

https://user-images.githubusercontent.com/769698/224954215-90e6beca-4913-4c0b-90fd-82db0a8d6c57.mp4

### DIFF
--- a/addon/core/keymap.ts
+++ b/addon/core/keymap.ts
@@ -42,9 +42,7 @@ const reduceParagraphIndent: Command = (state, dispatch) => {
     $cursor.pos === 0 ||
     // Skip action if cursor is not at the first position of "child"
     $cursor.parentOffset !== 0 ||
-    // Skip action if parent is not paragraph
-    $cursor.parent.type.name !== 'paragraph' ||
-    // Skip action if paragraph has no existing "indentationLevel"
+    // Skip action node has no existing "indentationLevel"
     !$cursor.parent.attrs.indentationLevel
   ) {
     return false;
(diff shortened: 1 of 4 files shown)


elpoelma (changes requested):
Works well and as expected overall. I would just remove the constraint that the parent should be a `paragraph` node, this way it works with all parents which have the attribute `indentationLevel`, such as headings etc.

addon/core/keymap.ts
```diff
@@ -29,7 +29,46 @@ import { liftEmptyBlockChecked } from '@lblod/ember-rdfa-editor/commands/lift-em
 
 export type Keymap = (schema: Schema) => Record<string, Command>;
 
+const reduceParagraphIndent: Command = (state, dispatch) => {
```
elpoelma, line 32: Can you move this command to a seperate file?

addon/core/keymap.ts
```diff
@@ -29,7 +29,46 @@ import { liftEmptyBlockChecked } from '@lblod/ember-rdfa-editor/commands/lift-em
...
+      'indentationLevel',
+      currentIndentationLevel - 1
+    );
+
+    dispatch(tr);
+
+    return true;
+  }
+
+  return false;
```
elpoelma, line 67: In this case you should also return `true`. When `dispatch` is not defined running the command only checks whether the operation is possible, in this case it is.

elpoelma (changes requested):
Looks good now! :) Just some additional comments.

elpoelma (approved):

PR #714, opened 2023-03-15
GN-4133: Remove whitespace around inline pills

dkozickis:
* Extract required parts of `AuPill` into `Pill`
* Change `Pill` "inline" style to be `inline-block`
* Change `outline` for `Link` "ember-node" to have `outline-offset` `-2px` for it not to cover nearby content when focused


elpoelma (commented):
Looks good, I think, as an alternative you should be able to replace `outline` by `border` and no longer need the `outline-offset`

elpoelma:
> @elpoelma yep `border` gives a layout shift, see video
> 
> while there is no layout shift with `outline`, see video

Aha ok :)

elpoelma (approved):

PR #715, opened 2023-03-15
text color plugin

usrtim:
This PR adds text color functionality. It provides a **color** mark which allows users to set the color of the text.

### DIFF
--- a/addon/components/plugins/text-style/color.ts
+++ b/addon/components/plugins/text-style/color.ts
@@ -1,0 +1,97 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { SayController } from '@lblod/ember-rdfa-editor';
+import { clearColor, setColor } from '@lblod/ember-rdfa-editor/plugins/color';
+import { paintCycleHappened } from '@lblod/ember-rdfa-editor/utils/_private/editor-utils';
+import { modifier } from 'ember-modifier';
+import { TextSelection } from 'prosemirror-state';
+
+type Args = {
+  controller: SayController;
+  defaultColor: string;
+};
+
+export default class ColorMenu extends Component<Args> {
+  dropdownButton?: HTMLElement;
+
+  setupDropdownButton = modifier(
+    (element: HTMLElement) => {
+      this.dropdownButton = element;
+    },
+    { eager: false }
+  );
+  @tracked dropdownOpen = false;
+
+  get controller() {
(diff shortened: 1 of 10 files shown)


addon/components/plugins/text-style/color.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        if (node === null && selection.$from.parentOffset > 0) {
+          node = this.controller.mainEditorState.doc.nodeAt(
+            selection.$from.pos - 1
+          );
+        }
+
+        const markSet = markType.isInSet(node?.marks || []);
+        const color = markSet?.attrs?.color as string;
```
elpoelma, line 63: I think you can use `selection.$from.marks()` in order to get the marks which are active at the beginning of the selection.

elpoelma (changes requested):
Looks good and works well overall :smile:, I just have some comments.

elpoelma:
> When colored text is selected it looses color
> 
> Likely CSS `::selection` we have somewhere. Not sure if you want to deal with it or not.

Yes that's true, we have the same issue with text highlighting, I don't think it's super urgent.

elpoelma:
> > When colored text is selected it looses color
> > Likely CSS `::selection` we have somewhere. Not sure if you want to deal with it or not.
> 
> Yes that's true, we have the same issue with text highlighting, I don't think it's super urgent.

What you could do, @usrtim is setting the text color of selected content to `inherit`.

usrtim:
> > > When colored text is selected it looses color
> > > Likely CSS `::selection` we have somewhere. Not sure if you want to deal with it or not.
> > 
> > 
> > Yes that's true, we have the same issue with text highlighting, I don't think it's super urgent.
> 
> What you could do, @usrtim is setting the text color of selected content to `inherit`.

@elpoelma that was not working for me. I used different approach, please have a look.

addon/plugins/color/marks/color.ts
```diff
@@ -0,0 +1,22 @@
...
+        return {
+          color: value,
+        };
+      },
+    },
+  ],
+  toDOM(mark: Mark) {
+    const { color } = mark.attrs;
+
+    return ['span', { style: `color: ${color as string}`, mark: 'color' }, 0];
```
elpoelma, line 20: I think it would be better to use a `data-mark` attribute instead of `mark`, so it is legal.

usrtim: Done

elpoelma (approved):
Looks good now! :)

PR #721, opened 2023-03-20
GN-4152: No more `undefined` with `prosemirror-invisibles` enabled

dkozickis:
When the `prosemirror-invisibles` is enabled, an `undefined` is no longer thrown on some occasions.

The issue was that the Editor was importing a `*.js` file of `prosemirror-view`, and `prosemirror-invisibles` was importing `*.cjs` file of `prosemirror-view`.

At some point `prosemirror-view` was comparing provided `decorations` and checking their `instanceof`, but because `instanceof` was coming from different sources, the comparison was always `false` leading to an invalid attempt at reading a field in an object.

The real fix is adding `exports` field in `prosemirror-invisibles` - https://github.com/lblod/prosemirror-invisibles/commit/d1d3d045691af737c30d436ff7dc73e5479b098f


elpoelma (approved):
Looks good and prevents the `undefined` errors from showing up.

PR #729, opened 2023-03-24
GN-3998: Instead of joining into 'block_rdfa' node - select it when deleting at the end of it

dkozickis:
https://user-images.githubusercontent.com/769698/229424296-8a94810e-f42f-476e-b62f-21c11594f4e5.mp4

### DIFF
--- a/addon/core/keymap.ts
+++ b/addon/core/keymap.ts
@@ -26,14 +26,20 @@ import {
   insertHardBreak,
   reduceIndent,
   liftEmptyBlockChecked,
+  selectBlockRdfaNode,
 } from '@lblod/ember-rdfa-editor/commands';
 import selectParentNodeOfType from '../commands/select-parent-node-of-type';
 import { hasParentNodeOfType } from '@curvenote/prosemirror-utils';
 import { undoInputRule } from 'prosemirror-inputrules';
 
-export type Keymap = (schema: Schema) => Record<string, Command>;
+type KeymapOptions = { backspace?: { selectBlockRdfaNode: boolean } };
 
-const backspace = chainCommands(
+export type Keymap = (
+  schema: Schema,
+  options?: KeymapOptions
+) => Record<string, Command>;
+
+const backspaceBase: Command[] = [
   undoInputRule,
   selectBlockRdfaNode,
   reduceIndent,
@@ -51,8 +57,16 @@ const backspace = chainCommands(
     }
     return false;
   },
-  selectNodeBackward
-);
+  selectNodeBackward,
+];
+
+const getBackspaceCommand = (options?: KeymapOptions) => {
+  if (options?.backspace?.selectBlockRdfaNode) {
+    return chainCommands(selectBlockRdfaNode, ...backspaceBase);
+  }
+
+  return chainCommands(...backspaceBase);
+};
 
 const del = chainCommands(
   deleteSelection,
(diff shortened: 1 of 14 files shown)


lagartoverde (approved):
Looks good :)

abeforgit (changes requested):
For some reason this makes every block behave like the block_rdfa
I don't think there's anything wrong with your code but there's some interaction going on with the command chain I think.

fyi non-text nodes are selectable by default

lagartoverde:
> For some reason this makes every block behave like the block_rdfa I don't think there's anything wrong with your code but there's some interaction going on with the command chain I think.
> 
> fyi non-text nodes are selectable by default

Could you elaborate on this? Because I didn't see anything wrong and want to know what's wrong for the next one :)

abeforgit:
additionally this pr doesn't deal with delete. I'm fine with splitting that into another pr though, lets get this one consistent first

abeforgit:
the point is that the two examples: the list item and the first paragraph, are not block_rdfa nodes but they are still nodeselected

abeforgit:
unfortunately....
minimal repro:
clear the doc, use customhtml to insert:
```html
<div typeof="Sometype"><p>this is inside the rdfa node</p></div><p><br>text</p>
```
backspace away the "text" and keep backspacing, now it doesn't trigger the nodeselection
but if you click on the newline right before the rdfa node, backspacing _does_ trigger correctly

abeforgit:
interestingly, you get p much the same behavior if you disable your selectBlockRdfaNode command altogether
I suspect you're approaching the same implementation as the selectNodeBackward command which is already a part of the backspace chain

abeforgit:
Im actually quite surprised at that, it breaks my assumptions about "isolating"

abeforgit:
aah further clues: the culprit actually seems to be this ad-hoc construction in the backspace chain:
```ts
  (state, dispatch, view) => {
    const isInTable = hasParentNodeOfType(state.schema.nodes.table)(
      state.selection
    );
    if (joinBackward(state, dispatch) && dispatch && view) {
      const { state } = view;
      if (!isInTable) {
        selectParentNodeOfType(state.schema.nodes.table)(state, dispatch, view);
      }
      return true;
    }
    return false;
  },
```

If I move the selectNodeBackward command to BEFORE this construction, it works as I had imagined

so I think it's the "joinBackward" mentioned above seems to have more power than we thought
the assumption was that it would follow the same rules as the "default" backspace behavior when it came to isolating/selectable nodes. It seems it maybe doesn't? Not sure, but I'd investigate there for sure

abeforgit (changes requested):
I couldn't break it anymore, good job :) 
The logic is ready, but it needs to be opt-in. So rather simply export the command and add an explanation in the docs how to enable it

addon/core/keymap.ts
```diff
@@ -79,7 +94,7 @@ const del = chainCommands(
 /// * **Delete** and **Mod-Delete** to `deleteSelection`, `joinForward`, `selectNodeForward`
 /// * **Mod-Delete** to `deleteSelection`, `joinForward`, `selectNodeForward`
 /// * **Mod-a** to `selectAll`
-export const pcBaseKeymap: Keymap = (schema: Schema) => ({
+export const pcBaseKeymap: Keymap = (schema, options) => ({
```
abeforgit, line 97: its ok for now, I suspect we'll revisit the outside-facing interface when we actually turn this on

addon/nodes/block-rdfa.ts
```diff
@@ -9,6 +9,8 @@ export const block_rdfa: NodeSpec = {
     __tag: { default: 'div' },
   },
   defining: true,
+  isolating: true,
```
abeforgit, line 12: unfortunately this also breaks current behavior
I suggest simply exporting a separate variation of this node, or just documenting how to extend nodes the host might want to treat like this.
we're in the early phases of the configuration interface where we simply expose all wires and ask the consumer to plug them in, so I'm ok with just documenting

abeforgit (changes requested):
one more thing to do to avoid breaking the public api
also, I would like a page in the dummy app that has this active so we can start playing around with it (so a new page, so we can easily compare)

abeforgit:
asking for the extra page is kind of out of scope, depending on how you look at it, so feel free to make a ticket for it and not do it in this pr
but it does have to be made non-breaking for current consuming apps

PR #732, opened 2023-03-28
GN-4199: Support SVG without height/width

dkozickis:
Give default height to an `*.svg` image, because they might not have their own height/width information, which results in "invisible" (0/0px) image

### DIFF
--- a/addon/components/plugins/image/insert-menu.ts
+++ b/addon/components/plugins/image/insert-menu.ts
@@ -8,6 +8,7 @@ const DEFAULT_SVG_HEIGHT = 100;
 
 type Args = {
   controller: SayController;
+  defaultSvgHeight?: number;
 };
 
 export default class ImageInsertMenu extends Component<Args> {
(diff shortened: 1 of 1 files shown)


lagartoverde:
Do you have any example of a svg without height? Code seems good, and seems to work with the svgs I tested, but not sure if they have height information or not as they are base64 encoded

abeforgit:
we don't support base64 encoding yet

https://www.vlaanderen.be/img/logo/vlaanderen-logo.svg

This was one that didn't work before and now works fine

abeforgit (approved):

lagartoverde:
> https://www.vlaanderen.be/img/logo/vlaanderen-logo.svg

So base64 is working as long as you use the insert menu, just for information :)

abeforgit:
> > [vlaanderen.be/img/logo/vlaanderen-logo.svg](https://www.vlaanderen.be/img/logo/vlaanderen-logo.svg)
> 
> So base64 is working as long as you use the insert menu, just for information :)

yeah but we don't save the image on export. so we should block data urls, and wouldn't you know it theres a story planned to do exactly that this sprint :)

PR #733, opened 2023-03-28
"insert" card should be collapsible but open by default

usrtim:
This PR is linked with: https://github.com/lblod/frontend-reglementaire-bijlage/pull/73

### DIFF
--- a/addon/components/collapsible.ts
+++ b/addon/components/collapsible.ts
@@ -1,9 +1,8 @@
 import { action } from '@ember/object';
-import { EmberNodeArgs } from '@lblod/ember-rdfa-editor/utils/_private/ember-node';
 import Component from '@glimmer/component';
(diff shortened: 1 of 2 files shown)


addon/components/collapsible.ts
```diff
@@ -1,12 +1,18 @@
 import { action } from '@ember/object';
+import { EmberNodeArgs } from '@lblod/ember-rdfa-editor/utils/_private/ember-node';
 import Component from '@glimmer/component';
-import { tracked } from 'tracked-built-ins';
+import { tracked } from '@glimmer/tracking';
 
-export default class CollapsibleComponent extends Component {
```
abeforgit, line 5: this is not an ember node, and as such does not take EmberNodeArgs as arguments

abeforgit (changes requested):
besides the other comment, the code is fine and works

PR #735, opened 2023-03-31
unsaved changes dialog showing too often

usrtim:


abeforgit:
why do you shortcut the function, skipping the generation of mark attributes when the node attribute is not null?

abeforgit:
confirmed to not solve the problem (it sidesteps it by disabling important functionality)

PR #749, opened 2023-04-12
Also observe responsive toolbar children for resize

lagartoverde:
There was a problem in GN with the besluit type plugin, when you selected a besluit this plugin changed it's size in the toolbar resulting in an overlap describen in https://binnenland.atlassian.net/browse/GN-4212, this PR aims to solve that.


elpoelma (changes requested):
Looks already good, but I would add the observer to the children of the toolbar main and side sections seperately.

elpoelma (approved):

PR #750, opened 2023-04-12
Bugfix: Validate image URL before insertion

x-m-el:
fix: [GN-4209](https://binnenland.atlassian.net/browse/GN-4209?atlOrigin=eyJpIjoiZDdhNGRkN2FiNGI1NDdkYzg4MzgxYzU4NTYxMmU4YWEiLCJwIjoiaiJ9)

In the Image insert modal, when pressing the insert button, the URL is first validated. If it is not valid, an error message is shown. This will disappear when the user clicks in the input box to fix the error.
The image preview also only shows if a valid URL was inputted.

`@onChange` and `{{on "change" ..}}` didn't work (because this is used for reactivity?), hence why `{{on "focus" ..}` is used instead to hide the error message.

In principle this could also be purely reactive, where the error shows directly when the user begins typing (and the submit button is disabled as long as the url is invalid). But this felt a bit chaotic.

### DIFF
--- a/addon/components/plugins/image/insert-menu.hbs
+++ b/addon/components/plugins/image/insert-menu.hbs
@@ -17,7 +17,10 @@
           <AuLabel @required='true' for={{id}}>
             {{t 'ember-rdfa-editor.image.url-label'}}
           </AuLabel>
-          <AuInput id={{id}} @width='block' @value={{this.url}} />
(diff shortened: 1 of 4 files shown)


elpoelma (approved):
Looks good to me!

PR #755, opened 2023-04-14
Add inputrules for list generation

elpoelma:
This PR adds two inputrules for list generation:
- one for generating a bullet-list when the user types `-` or `*` followed by a space.
- one for generating an ordered list when the user types `1.` followed by a space.


abeforgit (changes requested):

abeforgit:
I'm not sure how the `undoInputRule` is supposed to work, but I find it a bit awkward that I can't start a line with `- ` even if I want to. It turns into a list initially, which is nice, but on backspace it then removes the entire construct.

PR #777, opened 2023-05-02
GN-4224: Fix backspacing into nested lists

dkozickis:
This fixes the issue with backspacing into nested list from outside the list

https://user-images.githubusercontent.com/769698/235661490-181fe0c3-c1fb-4ff2-9fcf-69896b14521a.mp4

Unfortunately, there is still a bug, when backspacing into nested list (into level 2) from level 1 of the list. I'd like to work on that separately, I've taken too much time on this one already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Addition of list inputrules
 ### Fixed
 - Also observe responsive toolbar children for resize
+- Fix backspacing into nested lists
 
 
 ### Changed


lagartoverde (approved):
Looks good, I noticed a problem but I think is not in scope of this PR, so only fix it if you know how to do it :), when you are in a nested list and you delete until the list item is removed you will have a weird indentation because you will be on a empty paragraph belonging to the list item above

lagartoverde:
https://user-images.githubusercontent.com/17006446/235914454-2948cf40-37ba-4841-bf2e-326792328d1f.mov

abeforgit:
> Looks good, I noticed a problem but I think is not in scope of this PR, so only fix it if you know how to do it :), when you are in a nested list and you delete until the list item is removed you will have a weird indentation because you will be on a empty paragraph belonging to the list item above

I think this is actually what we want though

abeforgit:
Note this still doesn't actually fix the bug, it avoids it. There's two things interacting here, one which is obviously a bug, and the other which is debatably weird backspacing behaviour around lists. 
If you try the demo on the prosemirror homepage: https://prosemirror.net/ 
you'll see that backspacing into nested lists actually works similar.
[Screencast 2023-05-04 09:54:53.webm](https://user-images.githubusercontent.com/16419862/236143671-d334e277-f910-4fa3-b4af-88a9d80d659f.webm)
This behavior is definitely different from "conventional editors", but it is consistent and has a definite logic behind it: https://discuss.prosemirror.net/t/backspace-and-enter-and-delete-i-guess/57

However, as you'll see in the video, what prosemirror _doesn't_ do is randomly switch list styles while backspacing. _That_ is a bug, and as evidenced by the case you described, is still not fixed.

About the prosemirror default behavior: I'm not crazy about it, and I think the alternative implementation you've made here is very valuable. But we would be deliberately stepping away from a well-tested and well-researched flow, just to be able to mimic competitors who have infinitely more resources than we do. That's not a decision we should make lightly, and we need to be sure it will add significant value for our users if we do decide to invest the time in supporting this.

I'd like to see the underlying bug fixed. Don't throw away your work though, I have a feeling we'll need it later. Document what it does and why it was made.

I will then open a ticket for analysis what we want to do with backspace around lists, and when and how many resources we will assign to it.

abeforgit:
Further context: jira has also made that choice to alter the default behavior to be more in line with Word and Gdocs. And even in the discuss thread there are differing opinions. So it's quite a can of worms. I don't mind opening it, but we have to be aware what we're getting into.

PR #784, opened 2023-05-05
GN-4224: Backspace into nested lists fix

dkozickis:
Buggy behaviour was Initially introduced here - https://github.com/lblod/ember-rdfa-editor/pull/659, Not sure what was the reasoning apart from the one mentioned in the PR, can you comment please @lagartoverde?

Video with a bit of a hectic commentary on a video here - https://share.cleanshot.com/yjj4q37J

wasn't able upload it to PR as it is too big

See "Content Expressions" here - https://prosemirror.net/docs/guide/

Ticket - https://binnenland.atlassian.net/browse/GN-4240

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Dependencies
+- Bumps `sass` from 1.62.0 to 1.62.1
+- Bumps `@typescript-eslint/eslint-plugin` from 5.59.0 to 5.59.5
+### Added
+- Update `EmberNodeConfig` type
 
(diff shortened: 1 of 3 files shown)


abeforgit:
I think it was to disallow tables in lists :thinking:

abeforgit:
with this change you can't get out of the list by pressing enter anymore...

lagartoverde:
I can confirm I can exit the list using the enter, could you specify maybe what are you doing arne?

> I am unable to backspace or delete the Table into the list. I am able to insert it into the list, yes. Can we merge this PR and I will disable the table button for the link nodes? How does that sound? cc @lagartoverde @abeforgit

Respect to this, I think I would be okay with this, there are other ways a user can get a table into a list item like pasting, but I don't think this is a realistic thing to worry about. Let's see also what Arne thinks

abeforgit:
So I could reproduce, but then I reloaded the template and I couldn't anymore :thinking:
so yeah no idea what was going on, maybe a weird state in my localStorage

abeforgit (approved):

PR #785, opened 2023-05-08
GN-4216: Update EmberNodeConfig type

dkozickis:
Required for 

* https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/137
* https://binnenland.atlassian.net/browse/GN-4216


lagartoverde:
I don't understand why this is needed but in nodes like the one in `ember-rdfa-editor/addon/plugins/link/nodes/link.ts` the linter doesn't complain

lagartoverde:
> > I don't understand why this is needed but in nodes like the one in `ember-rdfa-editor/addon/plugins/link/nodes/link.ts` the linter doesn't complain
> 
> We need it for `EmberNodeConfig` consumers in other repos

Yes, what I'm saying is that it should also complain here in the editor

PR #801, opened 2023-05-30
Allow the parent component to get the innerView

lagartoverde:
I need the inner view to set the selection when clicking on the outer pill of a variable, related to this PR https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/142


abeforgit:
I suppose `controller.activeView` updates too late
@elpoelma please have a look if this is the best way to do achieve the end goal (selection set in inner editor when any part of the surrounding component is clicked)

elpoelma:
> I suppose `controller.activeView` updates too late @elpoelma please have a look if this is the best way to do achieve the end goal (selection set in inner editor when any part of the surrounding component is clicked)

This way seems to be good enough for me. `controller.activeView` is indeed updated after the `on click` handler is called.

elpoelma (approved):

PR #833, opened 2023-06-15
Add utility functions to work with node groups

elpoelma:
### Overview
This PR adds two new utility functions:
- `getGroups(node: PNode): string[]`: this function returns the list of groups a given node belongs to
- `hasGroups(node: PNode, ...groups: string[]): boolean`: this function checks whether a given node belongs to a given series of groups.

These utility functions can be useful in order to implement functionality which is applicable to a specific node-group.

None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/utils/_private/node-utils.ts
+++ b/addon/utils/_private/node-utils.ts
@@ -1,0 +1,10 @@
+import { PNode } from '@lblod/ember-rdfa-editor';
+
+export function getGroups(node: PNode) {
+  return node.type.spec.group?.split(' ') ?? [];
+}
+
+export function hasGroups(node: PNode, ...groups: string[]) {
+  const nodeGroups = getGroups(node);
+  return groups.every((group) => nodeGroups.includes(group));
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - add pr template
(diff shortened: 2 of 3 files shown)


addon/utils/_private/node-utils.ts
```diff
@@ -0,0 +1,10 @@
+import { PNode } from '@lblod/ember-rdfa-editor';
+
+export function getGroups(node: PNode) {
+  return node.type.spec.group?.split(' ') ?? [];
+}
+
+export function hasGroups(node: PNode, ...groups: string[]) {
+  const nodeGroups = getGroups(node);
+  return groups.every((group) => nodeGroups.includes(group));
+}
```
x-m-el, line 10: Might be an idea to also support `hasGroup`?
Or support passing just a string too (but then the function name makes less sense and that's probably not javascripty).
As I'm guessing you'd more often check for one specific group than a list of them.

x-m-el (approved):
small comment, reset looks good :+1: 
(I didn't test it)

addon/utils/_private/node-utils.ts
elpoelma: yes but that function supports that. You can invoke it by using `hasGroups(node, 'group')` or `hasGroups(node, 'group1', 'group2')`

x-m-el: Great, I looked over the spread operator :+1:

PR #834, opened 2023-06-15
fix(paste): fix no-space paste from word

abeforgit:
### Overview

Sometimes when copying from word we'd get really weird html, where almost every character is wrapped in its own span. Our pastehandler removed those too aggressively.

- added a way to specify elements to be cleaned when empty, but not trimmed
- reinstated the DOM sanitzation we had before, but with the `style` attribute allowed (we need it for colors and highlights)

https://binnenland.atlassian.net/browse/GN-4112

### How to test/reproduce

quite difficult. The particular html that's generated only happens on desktop word, and only in some particular cases. 
rather than going through the hassle of setting it up, I think it's more productive to test a variety of pasting scenarios and see if any regressions pop up.

### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
my local linter actually doesnt throw the errors the CI throws :thinking:

abeforgit:
wow a rare unit test failure, very nice
I'll fix that tomorrow

PR #840, opened 2023-06-16
Improve behaviour of ember-nodes with editable content

elpoelma:
### Overview
Currently, ember nodes with editable block content are in a bit of a cumbersome state. At the basic level they work, but in both the firefox and chromium based browsers there are a lot of issues concerning selections and event handling. This PR tries to tackle these issues by providing an improved implementation of the `ignoreMutation` and `stopEvent` methods.

Both methods are based on tiptaps implementation which can be found on https://github.com/ueberdosis/tiptap/blob/d61a621186470ce286e2cecf8206837a1eec7338/packages/core/src/NodeView.ts .
None

### Setup
None

### How to test/reproduce
The dummy application contains a sample `card` ember-node which can be inserted. This `card` component should allow any block content.

### Challenges/uncertainties
- nested `contenteditable` attributes should be prevented. Ember nodes which can contain editable content should never be `contenteditable: false` as a whole, but only the parts which are `contenteditable: false` should be marked as so:
E.g.: It's preferred to write
```hbs
<AuCard @flex={{true}} as |c|>
  <c.header contenteditable="false">
    <AuHeading @level="2" @skin="4">
      Title
    </AuHeading>
    <p>
      Subtitle
    </p>
  </c.header>
  <c.content>
    {{yield}}
  </c.content>
</AuCard>
```
instead of:
```hbs
<AuCard @flex={{true}} contenteditable="false" as |c|>
  <c.header>
    <AuHeading @level="2" @skin="4">
      Title
    </AuHeading>
    <p>
      Subtitle
    </p>
  </c.header>
  <c.content  contenteditable="true">
    {{yield}}
  </c.content>
</AuCard>
```

Both firefox and chrome have issues handling nested `contenteditable` attributes.

- Note: when backspacing and removing text, chromium-based browsers will remove the surrounding div if it is empty and has no attributes, this is why the following code is found in this PR: `this.contentDOM.dataset.content = 'true';`. It is to prevent the removal of the `contentDOM` element.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/addon/utils/_private/ember-node.ts
+++ b/addon/utils/_private/ember-node.ts
@@ -1,3 +1,13 @@
+/**
+ * Contains code from https://github.com/ueberdosis/tiptap/blob/d61a621186470ce286e2cecf8206837a1eec7338/packages/core/src/NodeView.ts#L195
+ *
+ * MIT License
+ * Copyright (c) 2023, Tiptap GmbH
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
 import { hbs, TemplateFactory } from 'ember-cli-htmlbars';
 import {
   AttributeSpec,
@@ -94,8 +104,7 @@ class EmberNodeView implements NodeView {
   contentDOM?: HTMLElement;
   emberComponent: EmberInlineComponent;
   template: TemplateFactory;
-  stopEvent: (event: InputEvent) => boolean;
-  ignoreMutation: (mutation: MutationRecord) => boolean;
+  config: EmberNodeConfig;
 
   constructor(
     controller: SayController,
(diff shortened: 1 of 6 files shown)


x-m-el (changes requested):
- I would add your special note about contenteditable somewhere, like in the ember-node file (at the top?). It will not stop the mistake, but hopefully give something to come across when trying to fix the related bug.
- For stopEvent no comments are provided. I think I understand the code (and it  means you probably don't need a custom stopEvent anymore), but it would be nice for a small comment of what it will and will not stop, to know at a glance if you need to implement your own stopEvent function when using an ember-node.

- the comment At the start of the file has some comments about stopEvent and general usage. Could you check those if they are still correct? 

With ember-nodes being so nice to use, but the code being hard to grasp, comments are important here I think, especially things that can't be seen by looking at the Card example.

addon/utils/_private/ember-node.ts
```diff
@@ -108,14 +117,8 @@ class EmberNodeView implements NodeView {
     // We set the new node here and pass it to the component to render it.
     // However, this will create a DOM mutation, which prosemirror will catch and use to rerender.
     // to avoid a NodeView rerendering when we already handled the state change, we pass true to `ignoreMutation` (by default).
```
x-m-el, line 119: note that this comment is now changed too (the new `ignoreMutations` implementation has enough comments that this can probably just be removed)

x-m-el (changes requested):
- new comments look good and I understand them, great :+1: 
- Bug that happens on your branch but not master (Only in Firefox, not in Chrome!):
When typing in the content of card, my spaces disappear, and I can't type more than one space anyway. This might be related to https://binnenland.atlassian.net/browse/GN-4350 which seems to be a similar bug.
I have no idea how these code changes could make this bug happen though... 

For the rest it seems to work decently, at least as far as I can test these things with the basic card.

elpoelma:
> * new comments look good and I understand them, great +1
> 
>     * Bug that happens on your branch but not master (Only in Firefox, not in Chrome!):
>       When typing in the content of card, my spaces disappear, and I can't type more than one space anyway. This might be related to https://binnenland.atlassian.net/browse/GN-4350 which seems to be a similar bug.
>       I have no idea how these code changes could make this bug happen though...
> 
> 
> For the rest it seems to work decently, at least as far as I can test these things with the basic card.

Fixed it, the issue was that the content inherited the white-space rule of the ember-node itself (which has `whitespace: normal`). Editable content needs `white-space: break-spaces`.

x-m-el (approved):
Yes the bug is fixed :+1: 

I do wonder if using `data-content` is wise though. As far as I understand a plugin using ember-node might add some `data-` attributes themselves, and `content` is pretty broad term that might get reused. Maybe be more specific and use "ember-node-content", to avoid potential clashes with plugins?

elpoelma:
> Yes the bug is fixed +1
> 
> I do wonder if using `data-content` is wise though. As far as I understand a plugin using ember-node might add some `data-` attributes themselves, and `content` is pretty broad term that might get reused. Maybe be more specific and use "ember-node-content", to avoid potential clashes with plugins?

Yes that would be better indeed :)

x-m-el (approved):
Can merge, thanks :)

PR #847, opened 2023-06-20
GN-4323: `SayController::setHtmlContent` can now receive `shouldFocus` option

dkozickis:
### Overview

`SayController::setHtmlContent` can now receive `shouldFocus` option which determines whether editor will be focused or not after calling `setHtmlContent`. Defaults to `true` for back compatibility.

### Connected issues and PRs:
Allows to display multiple editors on the page without giving the focus to them, useful when we are displaying editors based on some search input

https://binnenland.atlassian.net/browse/GN-4323


x-m-el (approved):
maybe a bit overkill to already add support for more options, but I don't see a reason not to do it either so :+1:

elpoelma (approved):

PR #857, opened 2023-06-22
Addition of a locale plugin

elpoelma:
### Overview
This PR adds a keyed prosemirror plugin which allows setting the locale of a document/the editor. This prosemirror plugin can be used by other plugins in order to decide in which language they need to display their content.
Additionally, this PR also exposes a `locale` getter on instances of the `SayController` class.
The plugin is added to the list of default plugins and has a default locale of `nl-BE`.

Jira ticket: https://binnenland.atlassian.net/browse/GN-4365?atlOrigin=eyJpIjoiNmY5MzVjZjhlYTM1NGUyN2I1MjZiODhkODIwNzE2MTYiLCJwIjoiaiJ9

### How to test/reproduce
The keyed locale plugin should be visible in the `plugins` tab of the prosemirror-dev-tools

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Addition of a `locale` prosemirror plugin which allows other plugins/code to lookup the locale/language of the current document/editor instance.
+- Addition of a `locale` getter to the `SayController` class which returns the state of the `locale` plugin.
+
 ## [3.10.0] - 2023-06-22
 ### Fixed
 - better handle weird edgecases when copying from word
--- a/addon/core/say-controller.ts
+++ b/addon/core/say-controller.ts
@@ -13,6 +13,7 @@ import {
   Schema,
 } from 'prosemirror-model';
 import { Command, Selection, Transaction } from 'prosemirror-state';
+import { localePluginKey } from '@lblod/ember-rdfa-editor/plugins/locale';
 
 export default class SayController {
   @tracked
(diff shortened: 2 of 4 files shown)


x-m-el (commented):
-- didn't review + didn't check in detail! --

I don't think a user of the editor can now change the language at runtime (which also means it can't change this for the consuming plugins). I think this could be useful, so a button/toggle "set document language" can be implemented, to switch the language at export/publish/etc depending on the preference of the user.

However, I can also see that the usecase might be to always use one language, where you choose it at "creation" of your document, instead of making it possible to still change "ad-hoc". (and even then, it is possible, just gotta re-initialize the editor then)

supporting ad-hoc is more general, but also means plugins need to handle the ad-hoc way (it is definitely a bigger challenge), so I think this "hard-setting at initialization" approach might ultimately be the best approach to avoid unnecessary difficulties.

elpoelma:
> -- didn't review + didn't check in detail! --
> 
> I don't think a user of the editor can now change the language at runtime (which also means it can't change this for the consuming plugins). I think this could be useful, so a button/toggle "set document language" can be implemented, to switch the language at export/publish/etc depending on the preference of the user.
> 
> However, I can also see that the usecase might be to always use one language, where you choose it at "creation" of your document, instead of making it possible to still change "ad-hoc". (and even then, it is possible, just gotta re-initialize the editor then)
> 
> supporting ad-hoc is more general, but also means plugins need to handle the ad-hoc way (it is definitely a bigger challenge), so I think this "hard-setting at initialization" approach might ultimately be the best approach to avoid unnecessary difficulties.

I'm personally hesitating between using a plugin like this one and simply relying on the browser language of a user like the rest of GN (or other applications) do. In the second case we also could reuse ember-intl and it would be consistent with the rest of the interface.

PR #872, opened 2023-06-29
GN-4387: document language support

elpoelma:
### Overview
This PR introduces the concept of a document language. The `doc` node-spec has been modified to support a `lang` attribute. Additionally, a tracked `documentLanguage` setter and getter have been added to the `SayController` class.

This PR comes with some breaking changes:
- The manner in which the `doc` node-spec should be added to an editor schema has been changed. `doc` (imported from `@lblod/ember-rdfa-editor/nodes`) is now a function which accepts an optional `DocumentConfig` object in which it is possible to provide a default language to the `doc` node-spec. 
e.g.:
```js
doc({
  defaultLanguage: 'nl-BE',
}),
```
Note that the defaultLanguage provided here is only used if the parsed document does not already contain a language attribute.
- The behaviour of the `setHtmlContent` method of the `SayController` class has changed. This method now creates a whole new state based on the provided html string. This means plugin state is not carried over from the previous state (e.g. the state of the history plugin resets).

Additionally, this PR adds a new `SetDocAttribute` Prosemirror step. This step allows for updating attributes on the `doc` top-level node in an immutable way.

Jira ticket: https://binnenland.atlassian.net/browse/GN-4387?atlOrigin=eyJpIjoiNTljM2JiN2I0ZDAzNDY0MDk4Njk0MTk4NmIwZTczZDYiLCJwIjoiaiJ9

### How to test/reproduce

- Setting the language of a document can be tested by employing the `documentLanguage` setter on the controller in the Web console by executing e.g. `window.__PC.documentLanguage = 'en-UK';`. If using the Prosemirror-dev-tools, you will notice the `lang` attribute of the top-level `doc` node has changed. 
- In order for a more visual test, you can add the sample `counter` ember-node which should instantaneously react to a document language change.
- The language of the document should be exported when clicking on `Show Export Preview`. The document language should be preserved across reloads of the dummy app.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint


abeforgit (changes requested):
before I get into the review
could we maybe make the new setHTML behavior optional? either a flag, or just a new method. I think both behaviors can be useful, and less breaking changes are always nice

addon/utils/_private/steps.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+  apply(doc: Node): StepResult {
+    const newDoc = doc.copy(doc.content);
+    this.prevValue = newDoc.attrs[this.key];
+    //@ts-expect-error Nodes are normally immutable in prosemirror, this is an exception
+    newDoc.attrs = {
+      ...newDoc.attrs,
+      [this.key]: this.value,
+    };
+    return StepResult.ok(newDoc);
```
abeforgit, line 19: do you see a way to avoid this? I think (though not yet tested) this will break a pretty fundamental propery
we can now compare docs by reference, so we know that if `doc === otherdoc`, that they will also contain the exact same contents. With this we have a (admittedly rare, but still) case where that is no longer true. 
I don't see any partcular reason we can't just clone the doc node?

abeforgit (changes requested):

addon/utils/_private/steps.ts
elpoelma: The `.copy` call should normally clone the doc node already I think?

elpoelma:
> before I get into the review could we maybe make the new setHTML behavior optional? either a flag, or just a new method. I think both behaviors can be useful, and less breaking changes are always nice

That might be better indeed. I split off the new functionality to an `initialize` method.

addon/utils/_private/steps.ts
abeforgit: yes, totally. somehow thought it worked differently, shouldve checked the docs :stuck_out_tongue:

abeforgit:
if we then also export the doc node with default options we can avoid breaking entirely
something like 
```js
export docWithConfig = (/*...*/) => {/*...*/}
export doc = docWithConfig({
  defaultLanguage: 'nl-BE',
  content: 'block+',
})
```

elpoelma:
> if we then also export the doc node with default options we can avoid breaking entirely something like
> 
> ```js
> export docWithConfig = (/*...*/) => {/*...*/}
> export doc = docWithConfig({
>   defaultLanguage: 'nl-BE',
>   content: 'block+',
> })
> ```

Updated the PR :)

PR #886, opened 2023-07-10
Implement ember-application prosemirror plugin which gives access to ember application instance

elpoelma:
### Overview
This PR adds the `ember-application` prosemirror plugin which stores the current ember application instance in the prosemirror state. This allows other plugins/nodes to access the current ember application instance.

### How to test/reproduce
In order to check if the ember application instance has been stored correctly in the state, you can access the plugin state in the web console and check if it is correct and present. You should also be able to observe the plugin using the prosemirror-dev-tools.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Added
+- Addition of an ember-application prosemirror plugin which allows for accessing the current instance of the ember application given a prosemirror state.
 ### Dependencies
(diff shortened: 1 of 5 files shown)


x-m-el:
(will also review):
What is the use case of this, why do we need this?
Pure speculation: Is this so you can use ember's dependency injection and that way use e.g. intl, router etc in your ember-nodes or something?

If so, some small note *somewhere* how to use this plugin in a useful way would be nice (so both how it works and what to use it for, preferably with an actual example). No idea what would be the best, be it a comment in code, an actual example using it in the dummy app, or readme/wiki (but I think it'll be hard to find a good spot for it then, without it feeling like some "floating" documentation).
Just at the top `addon/plugins/ember-application/index.ts` might be the most straightforward with at least some chance people come across it if they need it :shrug:

x-m-el (approved):
see comment above. Tested like mentioned and looks good :+1:

PR #908, opened 2023-07-24
GN-4264 embedded-editor extensions

x-m-el:
### Overview
This PR is a bit of a mess of features, some needed for [template-comments plugin](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/220). The commits undo/redo some changes because of changing a lot of choices inbetween development. Best to mostly ignore those and squash this PR.

#### Changes
Embedded-editor (not used by template comments):
- don't let a schema be passed (this does not do anything anyway)
- give possibility to pass custom keymap and plugins. 
- explanation to not use this for block content

For template comments (which uses content directly):
- add a `ParagraphWithConfig` to allow custom paragraph nodes.
- let indentationMenu (button to indent something) accept extra nodes that can be indented

see https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/220 

### How to test/reproduce
Best tested with [template comments](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/220) which uses the new functionality.

Embedded editor:
Not really any way to test easily. But the configuration will be useful to fix some bugs (e.g. https://binnenland.atlassian.net/browse/GN-4433 ). If in review there is no certainty it works/is a good idea, I am fine to remove it from this PR and add them when fixing a bug that needs it.

### Challenges/uncertainties
~~This bug also appears in other instances using embedded-editor (like the variable plugin).~~
~~- when changing a mark (e.g. bold with CTRL+B), the button for this mark will not get "pressed" like with the normal editor. It might also have some other weird behavior (but didn't test further)~~ => Fixed by not using the "schema" inside a getter.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/ember-node/embedded-editor.ts
+++ b/addon/components/ember-node/embedded-editor.ts
@@ -61,6 +61,10 @@ type Args = EmberNodeArgs & {
   plugins?: Plugin[];
 };
 
+/**
+ * An embedded editor to use for *inline* content. This way you can specify extra content for an
+ * inline (atom) node. For block content, use content directly instead ({{yield}} in ember-nodes).
+ */
 export default class EmbeddedEditor extends Component<Args> {
   @service declare intl: IntlService;
   innerView: SayView | null = null;
--- a/addon/components/ember-node/embedded-editor.ts
+++ b/addon/components/ember-node/embedded-editor.ts
@@ -73,8 +73,6 @@ import { Plugin } from 'prosemirror-state';
 type Args = EmberNodeArgs & {
   placeholder: string;
   initEditor?: (view: SayView) => void;
-  /* override the schema */
-  schema?: Schema;
   /* override the keymap. Pass outerView where needed. See `embeddedConfig` option in baseKeymap as example) */
   keymap?: { [key: string]: Command };
   /* editor plugins to add */
@@ -97,6 +95,10 @@ export default class EmbeddedEditor extends Component<Args> {
     return this.args.node;
   }
 
+  get schema() {
+    return this.controller.schema;
+  }
+
   get pos() {
     return this.args.getPos();
   }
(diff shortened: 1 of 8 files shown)


elpoelma:
On the issue with the bold toggling using the `Ctrl+B` shortcut, that is fixed now I think, as the schema is now shared?

addon/components/ember-node/embedded-editor.ts
```diff
@@ -97,31 +93,45 @@ export default class EmbeddedEditor extends Component<Args> {
...
-      marks: {
-        inline_rdfa,
-        link,
-        em,
-        strong,
-        underline,
-        strikethrough,
-      },
-    });
+  get keymap() {
```
elpoelma, line 100: Maybe we should let `this.args.keymap` override the whole keymap so it is fully customizable?

addon/components/plugins/indentation/indentation-menu.ts
```diff
@@ -18,11 +20,25 @@ export default class IndentationMenuComponent extends Component<Args> {
     return this.controller.schema;
   }
 
+  get extraTypes() {
+    if (!this.args.extraTypes) return [];
+    if (this.args.extraTypes instanceof Array) return this.args.extraTypes;
+    return [this.args.extraTypes];
+  }
+
+  get allowedTypes() {
```
elpoelma, line 29: I would modify this so that all the indentable types are overriden by `this.args.types` and by default use `[this.schema.nodes.paragraph, this.schema.nodes.heading]`. In this way, it is fully customizable.

elpoelma (changes requested):
Looks good overall, just some comments.

Aside from the comments added below, I would also add a disabled state to the `mark` toolbar button so the mark button is disabled when it can't be applied.

addon/components/ember-node/embedded-editor.ts
x-m-el: it is done this way because undo/redo should work on the "outer state" and really the embedded-editor should not care about what undo/redo does. 

You could remove `...basemap`, so there only the undo/redo overrides are done. That might make more sense (and baseMap should really be inside keymap.ts, so a node using embedded-editor can choose to use baseMap and change whatever it wants from it).

addon/components/plugins/indentation/indentation-menu.ts
x-m-el: Yes, that makes more sense :+1:

x-m-el:
marks can now be disabled too if not allowed. +did all other comments (see resolved comments for more info).

Note: on the editor branch for template comments the disabled mark button does not happen yet. Enter still inserts normal paragraphs (even though they are not allowed), which means the marks are also allowed...

elpoelma (approved):
Looks good! Once a changelog has been added, you can merge.

PR #918, opened 2023-08-02
Fix parsing rules of ParagraphWithConfig and its inclusion in content

x-m-el:
### Overview
ParagraphWithConfig was recently developped, but has two problems:
- lists only accepted "paragraph" content, but not ones made with `ParagraphWithConfig`.
- `paragraphWithConfig` did not get parsed as such and would be converted to a normal paragraph.

This is fixed by:
- adding a group "paragraphGroup" to use in content. Better would be a group "paragraph", but this clashes with the typical name used in schema's (and changing that would be a breaking change). This group is then used in lists instead of just `paragraph`.
~~adding an attribute ParagraphType, so the parsing can know if the paragraph is a special one or a regular paragraph. Normal paragraphs did not change (and do not have extra data attribute). But custom paragraph will have a type, as long as "name" is passed.~~
=> instead added a custom node-spec "subType". This also get added as to dataset if it is not empty. This way it does not get added for normal paragraphs (keeping the old way of working).

https://binnenland.atlassian.net/browse/GN-4445
~~bug that you could not add lists in template comments (no issue was created for this)
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/241~~

### Setup
<del>a ParagraphWithConfig is used in [plugins repo #241](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/241).
Best to test this together with that. Note you need to
`npm link` ember-rdfa-editor
then in ember-rdfa-editor-lblod-plugins `npm link @lblod/ember-rdfa-editor`
And `ember ts:precompile` if you want types to work.</del>
This is not used anymore in template comment, so no setup needed anymore.

### How to test/reproduce
~~Add a template comment, see you can add lists.
When reloading, see that paragraphs are not converted to lists, or more importantly, in the structure see that it stays the same (and does not get converted to e.g. a regular paragraph).~~
Template comments don't use this anymore, hence this code is not used anywhere yet and hard to test
=> should add a custom node that uses this functionality.

### Challenges/uncertainties
- ParagraphGroup might not be the best name, but I'm not sure if there is anything better.
- ParagraphWithConfig is a bit messy and might need a refactor to clean up the toDom and parseDom a bit (like done with variables). But as the logic is quite small, it still reads clearly and I decided to not pre-refactor as I don't think it is needed (yet).

- !As template comments don't need this anymore, `ParagraphWithConfig` is not needed anymore. The code does work though, and it does give some customization. So we could still merge this. That way it exists for potential future usages, but also gives an example how to do a `NodeWithConfig`.


x-m-el:
Because you should not change the paragraphType, should this instead be a node-spec instead of part of the attributes? Not sure if you can dynamically set nodeSpec in parseDom

x-m-el:
Was able to set it as a nodespec which makes more sense. ready for review :+1:

addon/nodes/paragraphWithConfig.ts
```diff
@@ -1,16 +1,102 @@
...
-  marks = paragraph.marks,
-  group = paragraph.group,
-} = {}) => {
-  return { ...paragraph, content: content, marks: marks, group: group };
+export const paragraphWithConfig: (
+  config: ParagraphConfig,
+) => ParagraphNodeSpec = (config) => {
+  const name = config.subType !== '' ? config.subType : BASE_PARAGRAPH_TYPE;
+  return {
+    name: name,
```
elpoelma, line 43: Is the name attribute just for debugging purposes?

addon/nodes/paragraphWithConfig.ts
```diff
@@ -1,16 +1,102 @@
...
+        },
+        consuming: false,
+      },
+    ],
+    toDOM(node) {
+      const attrs: ParagraphDataAttributes = {
+        'data-indentation-level': node.attrs.indentationLevel as number,
+      };
+      const subType = (node.type.spec as ParagraphNodeSpec).subType;
+      if (subType && subType !== '') {
```
elpoelma, line 96: the `subType !== ''` check is not necessary here.

elpoelma (commented):
Code seems sane to me, still have to test it with the other PR. In the meanwhile some comments.

addon/nodes/paragraphWithConfig.ts
x-m-el: I think we don't use the name anywhere in code. I do often use it for debugging indeed.
So it is not required, but seemed less clean to keep it empty, because there is nothing bad about it being a sensible name I think? :)

elpoelma: I agree :)

elpoelma (approved):

PR #923, opened 2023-08-03
Add nodespec `indentable` that specifies which nodes can be indented

x-m-el:
### Overview
More nodes than just paragraphs and lists had to be indentable. Before a change was made to make the list configurable. However, this is cumbersome. This PR instead gives the choice to the node itself. 
A nodespec `indentable` is used to check if a node can be indented.
`allowedTypes` which was used previously is removed.

~~needed in https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/236~~ => closes, this PR isn't strictly needed anymore but does cleanup code, so worth to still keep around.
previous way with AllowedTypes added in: https://github.com/lblod/ember-rdfa-editor/pull/908/

### How to test/reproduce
add lists, headings and paragraphs, see if indenting still works as expected.

### Challenges/uncertainties
This creates less configurability, but in 99% of cases that isn't needed and specifying at node level will be good enough. This also cleans up the code a bit, the indent button does not need to know about nodes at all. It just fires the correct command.


elpoelma (changes requested):
Works as expected, just two comments:
- the `heading` node-spec should also have `indentable: true` set
- this is technically a breaking change I think? As the `allowedTypes` option has been removed.

x-m-el:
@elpoelma fixed the comment, however:
There is also a command called `reduceIndent` in `commands/reduce-indent.ts`. 
This has the same functionality, but specifically to use in a button command for backspace (which is different from `indentNode` command). There it only checks if a node has the attribute `indentationLevel` to know if it is indentable.

Is this more sensible? In a way the `indentable: true` is a bit redundant if we can just check for existance of `indentationLevel`?

Would you agree to:
- Remove `indentable`
- just check if `indentationLevel` exists in a nodespec's attributes.

elpoelma:
> @elpoelma fixed the comment, however: There is also a command called `reduceIndent` in `commands/reduce-indent.ts`. This has the same functionality, but specifically to use in a button command for backspace (which is different from `indentNode` command). There it only checks if a node has the attribute `indentationLevel` to know if it is indentable.
> 
> Is this more sensible? In a way the `indentable: true` is a bit redundant if we can just check for existance of `indentationLevel`?
> 
> Would you agree to:
> 
> * Remove `indentable`
> * just check if `indentationLevel` exists in a nodespec's attributes.

I agree, this removes the need for an `identable` property. Having both an `indentable` property and `identationLevel` attribute seems redundant to me anyways.

elpoelma (approved):
Works well :)

PR #937, opened 2023-08-16
Fetch prosemirror-invisibles via https instead of ssh

x-m-el:
### Overview
An issue was mentioned that git+ssh does not work on Jenkins, as it does not support ssh, with error 

> Host key verification failed & fatal: Could not read from remote repository on /usr/bin/git ls-remote -h -t ssh://git@github.com/lblod/prosemirror-invisibles.git.
 
As the repo is public, https can be used instead, which should fix this problem.
 
### How to test/reproduce
Not a real good way to test. This gives no difference locally and should fix the problem mentioned as far as I can find.


elpoelma (approved):
Looks good, `npm install` no longer changes the package-lock locally on my system.

PR #939, opened 2023-08-17
Add functions to find nodes up and down the document (in order of user-view)

x-m-el:
### Overview
This implements a function that searches a node (bases on a predicate) up/down the document. It does this in order of how a user views a document, or in easier terms, in increasing and decreasing position. This is often what we want to check when moving or inserting a node.
It only checks and returns positions between nodes.

The moving down algorithm works as follows:
- look before the passed node if can insert there.
- if not, go inside the node (to its children) and look to insert as much to the left as possible.
- if not, go even deeper. If no children, go to a space to the right of the original node and check there.
- if nothing in the parent of the passed node is found, we need to go up the tree.
- when going up, the same is done there.    
In simpler terms it really just loops from top to bottom. 

For moving up, the simplest algorithm is used:
loop over positions, see if it is between nodes and return if so.

A quick test with a small-ish document (5000 positions) was done: looping over positions (18ms) versus specific algorithm (12ms). 
This means that the position-looping algorithm is probably good enough (for now).

https://binnenland.atlassian.net/browse/GN-4442
Best way to test together with [PR plugins @229](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/229)

### Setup
to test together with [PR plugins @229](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/229), see setup there.

### How to test/reproduce
move anything up/down in the structure. Because the basic editor does not have a lot of structure, it is easiest to test in [PR plugins @229](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/229)

### Challenges/uncertainties
/

### DIFF
--- a/addon/utils/_private/document-walker.ts
+++ b/addon/utils/_private/document-walker.ts
@@ -64,7 +64,7 @@ export function* findNodePosDown(
     // passed node, so an offset has to be kept for the final result.
     let offsetChildren = 0;
     while (index < parent.childCount) {
-      const matchedChildOffsets = FindNodePosChildrenAsOffset(
+      const matchedChildOffsets = findNodePosChildrenAsOffset(
         parent,
         index,
         offsetChildren,
@@ -101,7 +101,7 @@ export function* findNodePosDown(
  * @param predicate the predicate to check for positions between nodes
  * @returns the offset (starting from the parent's startIndex) to the position that was found
  */
-export function* FindNodePosChildrenAsOffset(
+export function* findNodePosChildrenAsOffset(
   parent: PNode,
   startIndex: number,
   currentOffset: number,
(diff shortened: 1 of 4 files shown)


addon/utils/_private/document-walker.ts
```diff
@@ -0,0 +1,162 @@
...
+ * returns the offset from the start of the search where a match is found
+ * a global position can be found by adding the returned offset to the passed start position
+ * Checks in order of the "document", from left to right.
+ * @param parent the parent node to check the childs for
+ * @param startIndex the first index to check in the given parent (0 to check all children)
+ * @param currentOffset the current offset which is used by the recursion. An initial value can be passed but this does not change anything for the algorithm.
+ * @param predicate the predicate to check for positions between nodes
+ * @returns the offset (starting from the parent's startIndex) to the position that was found
+ */
+export function* FindNodePosChildrenAsOffset(
```
abeforgit, line 104: accidental PascalCase instead of camelCase here

abeforgit (changes requested):
just a small typo, good otherwise

looks sensible from a high level, and couldn't break it in (moderate) testing
(no need to wait for another review once you fix typo, just merge)

PR #946, opened 2023-09-01
GN-4433: Links - Bind enter-key to leave node/input-box

x-m-el:
### Overview
Enter-key did nothing in links. This adds the functionality that pressing enter-key leaves the node and puts the cursor after the link-node (which is after the pill).

https://binnenland.atlassian.net/browse/GN-4433

### How to test/reproduce
add a link
see that pressing enter leaves the node (sets cursor behind the node)

### Challenges/uncertainties
Created a modifier which is a lot more cleaner than adding code to the component. number-variable has similar logic and should also use this modifier.
I did not test if this is directly useable in the plugins repo without doing some special exporting.

This adds an explicit dependency of `ember-modifier`.

### DIFF
--- a/addon/modifiers/leave-on-enter-key.ts
+++ b/addon/modifiers/leave-on-enter-key.ts
@@ -1,26 +1,27 @@
 import { modifier } from 'ember-modifier';
 import { SayController, TextSelection } from '..';
 
-export default modifier(function leaveOnEnterKey(
-  element: HTMLElement,
-  controller: SayController,
-  startPosNode: number,
-) {
-  const state = controller.mainEditorState;
-  const node = state.doc.resolve(startPosNode).nodeAfter;
-  const leaveOnEnter = (event: KeyboardEvent) => {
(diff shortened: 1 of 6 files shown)


addon/modifiers/leave-on-enter-key.ts
```diff
@@ -0,0 +1,27 @@
...
+      controller.mainEditorView.dispatch(tr);
+      controller.mainEditorView.focus();
+    };
+
+    element.addEventListener('keyup', leaveOnEnter);
+
+    return () => {
+      element.removeEventListener('keyup', leaveOnEnter);
+    };
+  },
```
elpoelma, line 26: I think invoking `modifier` without the `{ eager: false }` argument is deprecated.

elpoelma (changes requested):
Code looks sane, but while testing this PR, ember gives an error on not finding the `leave-on-enter-key` modifier.

abeforgit (approved):

PR #952, opened 2023-09-04
Run the Embroider "safe" and "strict" scenarios in CI

Windvis:
This adds the Embroider `safe` and `strict` scenarios to Woodpecker again. They are allowed to fail since there are still some issues (which can be resolved in separate PRs. Once they succeed again we can add them to the other `ember-scenarios` job.

### DIFF
--- a/.woodpecker/.embroider-scenarios.yml
+++ b/.woodpecker/.embroider-scenarios.yml
@@ -1,0 +1,15 @@
+# These can be merged into the ember-scenarios job once they pass
+matrix:
+  scenario:
+    - embroider-safe
+    - embroider-optimized
+steps:
+  - name: ${scenario}
+    image: danlynn/ember-cli:4.8.0
+    failure: ignore
+    commands:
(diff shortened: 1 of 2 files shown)


Windvis:
CI seems to fail because there is no changelog change but this is not something that affects the consumer? Is a changelog entry _always_ required?

abeforgit:
yeah we don't do CI changes very often, so we didn't get around to adding an exception for it, and yeah I just add a CL entry if I do make a CI change
What should and should not go into the CL is a debate we don't really have the energy for :stuck_out_tongue:

Windvis:
Yea, that's fair. I just prefer changelogs that aren't noisy for consumers 😄. Anyway, I can also just ship this change as part of the other PR. That one has a changelog entry already. Or do you prefer to keep them split up (with a new entry for this one).

abeforgit:
no combining is fine :)

Windvis:
Closing and including this change in #953

Issue #945, opened 2023-09-01
Embroider compatibility


Windvis:
When trying to build the loket app under Embroider I encountered the following errors related to the editor addon:

<details>
<summary>1. Unsafe dynamic component</summary>

> Unsafe dynamic component: this.componentPath in node_modules/.embroider/rewritten-packages/@lblod/ember-rdfa-editor.0c1c90e7/utils/_private/ember-node.js

Which comes from this file: https://github.com/lblod/ember-rdfa-editor/blob/8261f1ef8a2b35dd8fc9650832ec4605c68497dd/addon/utils/_private/ember-node.ts#L151-L163

This should be solvable by passing the component class instead of the path and invoking it without the `component` helper:

```js
this.template = hbs`
  <this.component
    @getPos=this.getPos
    @node=this.node
    @updateAttribute=this.updateAttribute
    @controller=this.controller
    @view=this.view
    @selected=this.selected
    @contentDecorations=this.contentDecorations
  >
    {{#unless this.atom}}
      <EmberNode::Slot @contentDOM={{this.contentDOM}}/>
    {{/unless}}
  </this.component>`;
```
I'm not sure if this is something used outside of the addon itself, but if it is then I guess it's a breaking change 😄.
</details>

<details>
<summary>2. Ember Velcro component not found</summary>

The following error is thrown at build time:
> ERROR in ../rewritten-packages/@lblod/ember-rdfa-editor.0c1c90e7/components/toolbar/dropdown.hbs 1:0-58
Module not found: Error: Can't resolve '#embroider_compat/components/velcro' in '/Users/windvis/Development/Projects/digitaal-loket/frontend-digitaal-loket/node_modules/@lblod/ember-rdfa-editor'

I briefly looked at it and don't see anything wrong so this is probably an Embroider Bug, I'll try to find out more about this.
</details>

<details>
<summary>3. `crypto` fallback error</summary>

The addon includes some ember-auto-import configuration to provide fallbacks for node-only code. This config isn't used by Embroider, so the build fails.

```
ERROR in ../../@graphy/core.data.factory/main.js 1:15-32
Module not found: Error: Can't resolve 'crypto' in '/Users/windvis/Development/Projects/digitaal-loket/frontend-digitaal-loket/node_modules/@graphy/core.data.factory'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

[...]

PR #953, opened 2023-09-04
Resolve some Embroider issues

Windvis:
### Overview

This resolves some of the errors and warnings that are shown when building under the Embroider "safe" preset.

#945 
#952


Windvis:
There is still an issue with `@ember/string` but I'm trying to find out what the correct fix is here.

Windvis:
I went with option 1 for now. Personal preference would be to go for option 4 since those addons aren't really maintained, but that can be done in separate PRs.

abeforgit:
I think 4 should be trivial, `htmlSafe` is builtin since at least 3.28 it seems

Windvis:
It was built-in before that as well, as part of the fake `@ember/string` package, but it was moved to `@ember/template` later. So yea, should be trivial. I just wasn't sure about the `ember-feature-flags` usage in the addon (or the plans for that in the future).

abeforgit:
That's a holdover from a long time ago, and we're not planning on using it for anything at the moment, in fact I would really like to minimize the amount of different scenarios we support

so uh yeah safe to remove

abeforgit:
I don't really understand why the ci is failing on the peerdep, since the version you bumped it to doesnt even have a peerdep on ember anymore afaict

Windvis:
v11 of the resolver doesn't support Ember <4.8 it seems. Not sure what the best fix is.
- stay on v8 (+ ember/string devDev)
- use overrides in the scenarios. v8 on 3.28 and v9 on 4.4 LTS (even though it isn't being run at the moment.

abeforgit:
I mean if its this much of a pain im inclined to just drop 3.28 as well if its ok for loket

Windvis:
I just pushed the overrides. 🤞 

> I mean if its this much of a pain im inclined to just drop 3.28 as well if its ok for loket

Yea, I think it's fine to drop support and cut a major 👍 It's just nicer if that's a separate PR 😄.

PR #962, opened 2023-09-05
Fix leaveOnEnterKey Bug + extra modifiers LeaveWithArrowKeys and SelectNodeOnClick

x-m-el:
### Overview
https://github.com/lblod/ember-rdfa-editor/pull/946 seemed to still have a problem sometimes where pressing enter did not place the cursor correctly. With console logging, it seemed that the getter did not get the updated position value. (note: this getter is also done in embeddable-editor, but didn't come across the same problem there).
The fix is to just pass the `getPos` function and use that to get the position. This way it is definitely the correct (updated) position and the whole ember-tracking is fully skipped instead.

This also adds two other modifiers that keep the same syntax for ease of understanding.
Both are used in number-variable (usage not changed, this is purely a "refactor")
`leave-with-arrow-keys` lets you use the left and right arrow key to leave a certain element.
`select-node-on-click` selects the node passed when the given bound element is clicked. 

https://binnenland.atlassian.net/jira/software/c/projects/GN/boards/4?selectedIssue=GN-4433
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/285

### Setup
for the fix, no setup needed.
For the extra modifiers, `npm link` this repo and `npm link @lblod/ember-rdfa-editor` in the plugins repo. Then `ember serve` the plugins repo.

### How to test/reproduce
add a link, see that enter puts the cursor `after` the node, always. Even when typing in extra text.

For the other modifiers: see no change in usage of arrow keys and enter for a number variable.

### Challenges/uncertainties
This started out as a sort of test to see if modifier could work and clean up the code.
I am convinced these are amazing to use. They decrease a lot of the code in the components and keep related code together, help reusability and abstract away harder-to-grasp "on 'keyup' " listeners.

Class based modifiers feel very ugly because of all the null checking. But as the `modifier` function sets what is needed, I don't think there is a better way to handle this?

### DIFF
--- a/.changeset/silent-oranges-impress.md
+++ b/.changeset/silent-oranges-impress.md
@@ -1,0 +1,7 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+- Fix `leave-on-enter-key` modifier to always use the correct position to leave a node from.
+- Add `leave-with-arrow-keys` modifier, which can be used to leave an node's element with the left and right arrow key.
+- Add `select-node-on-click` modifier, which can be used to select a given node when clicking the bound element.
--- a/addon/components/ember-node/link.hbs
+++ b/addon/components/ember-node/link.hbs
@@ -18,7 +18,7 @@
       @placeholder={{t 'ember-rdfa-editor.link.placeholder.text'}}
       @decorations={{@decorations}}
       @contentDecorations={{@contentDecorations}}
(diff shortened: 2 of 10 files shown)


x-m-el:
I am not sure why it can't find `@ember/destroyable`. As far as I can see it should be part of the ember version used? https://github.com/emberjs/rfcs/blob/master/text/0580-destroyables.md

PR #963, opened 2023-09-05
Provide a Webpack config object

Windvis:
### Overview
To support Embroider some extra configuration is needed. To make it easier for consumers we provide this config as a CJS module.

Solves issue 3 in #945

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -15,31 +15,6 @@ module.exports = function (defaults) {
       cascade: true,
       sourcemap: true,
     },
-    autoImport: {
-      webpack: {
-        node: {
-          global: true,
-          __filename: true,
-          __dirname: true,
-        },
-        module: {
-          rules: [{ test: /\.handlebars$/, loader: 'handlebars-loader' }],
-        },
-        // plugins: [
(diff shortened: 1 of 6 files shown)


ember-cli-build.js
```diff
@@ -15,31 +15,6 @@ module.exports = function (defaults) {
       cascade: true,
       sourcemap: true,
     },
-    autoImport: {
```
Windvis, line 18: The addon already provides the autoImport config in the index.js file so this shouldn't be needed (unless something extra only for the dummy app is needed).

ember-cli-build.js
```diff
@@ -15,31 +15,6 @@ module.exports = function (defaults) {
...
-      webpack: {
-        node: {
-          global: true,
-          __filename: true,
-          __dirname: true,
-        },
-        module: {
-          rules: [{ test: /\.handlebars$/, loader: 'handlebars-loader' }],
```
Windvis, line 26: Not sure if this is still used somewhere?

abeforgit: I honestly have no idea what this does/did

abeforgit (approved):

PR #991, opened 2023-10-09
Improve types and documentation of EmberNode

piemonkey:
### Overview
Moving around and adding to existing documentation for EmberNodes. Especially focused on documentation for EmberNodeConfig, hopefully in a way that produces useful hints on any editor (tested in Neovim using Lsp and vscode).

Some type improvements, mostly making existing types a little more specific. Hopefully I haven't misunderstood constraints.

https://binnenland.atlassian.net/browse/GN-4429

### Setup
N/A

### How to test/reproduce
Try creating an ember node and checking what help the editor gives you.

### Challenges/uncertainties
I'm not 100% on whether I've accurately captured the type constraints or documented some of the options as I had to determine what they were doing from the code and git history.

It's unclear whether it's better to have the documentation attached to one type and linked from the 2 main entry point methods or whether it would be better to duplicate it for each of them.

### Checks PR readiness
- [-] UI: works on smaller screen sizes
- [-] UI: feedback for any loading/error states
- [-] Check cancel/go-back flows
- [-] Check database state correct when deleting/updating (especially regarding relationships)
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/sixty-dingos-behave.md
+++ b/.changeset/sixty-dingos-behave.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Improve types and documentation of EmberNode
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Includes some very nice documentation improvements :) 
The type changes look sensible and also don't seem breaking at first sight.

On the longer term I am wondering whether we should split up the `EmberNodeConfig` object which is passed to both `createEmberNodeSpec` and `createEmberNodeView`

PR #1012, opened 2023-10-31
Ensure links in outputted html open in new tab

elpoelma:
### Overview
This PR ensures that a `target: "_blank"` attribute is added to the `toDOM` of both the link mark and node to ensure that links in the outputted html is opened in a new tab.

### How to test/reproduce
- Start the dummy app
- Insert a link with a valid url and text
- Export the editor content to static html
- Ensure that when clicking the outputted links, they are opened in a new tab.

### Challenges/uncertainties
Technically, this change is breaking, as links in the exported html now open in a new tab by default. However, this change streamlines the behaviour with the behaviour of links in-editor.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/three-forks-try.md
+++ b/.changeset/three-forks-try.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Add `target: '_blank'` attribute to outputted html of a link
--- a/addon/plugins/link/mark/link.ts
+++ b/addon/plugins/link/mark/link.ts
@@ -19,7 +19,7 @@ export const link: MarkSpec = {
     },
   ],
   toDOM(mark) {
-    return ['a', mark.attrs, 0];
+    return ['a', { target: '_blank', ...mark.attrs }, 0];
   },
   hasRdfa: true,
   parseTag: 'a',
--- a/addon/plugins/link/nodes/link.ts
+++ b/addon/plugins/link/nodes/link.ts
@@ -45,7 +45,8 @@ const emberNodeConfig: (options: LinkOptions) => EmberNodeConfig = (
     toDOM(node) {
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const { interactive, placeholder, ...attrs } = node.attrs;
-      return ['a', attrs, 0];
+      const outputAttrs = { target: '_blank', ...attrs };
+      return ['a', outputAttrs, 0];
     },
   };
 };


lagartoverde (approved):
Works good

abeforgit:
do we want this for all links? to me this is a bit weird to decide in the editor. shouldn't this be the responsibility of the place that displays the final document?

elpoelma:
> do we want this for all links? to me this is a bit weird to decide in the editor. shouldn't this be the responsibility of the place that displays the final document?

Good point, I guess that the main application can decide to add the `target` attribute themselves to the outputted html as a postprocessing step too.

elpoelma:
> do we want this for all links? to me this is a bit weird to decide in the editor. shouldn't this be the responsibility of the place that displays the final document?

@lagartoverde does this approach sound good to you?

lagartoverde:
I think the postprocessing is weird as this is something we will want in 99% of the documents

abeforgit:
I think 99% is a gross overstatement, from what I can find online the question of whether to open external links in a new tab or not is still hotly debated, I'll check with design

abeforgit:
the WCAG at least _recommend_ limiting new tabs as much as possible: https://www.w3.org/WAI/WCAG21/Understanding/on-focus

abeforgit:
What I would support is making sure that the editor maintains the target attribute on links when importing html, and add a feature to the link editor to let the document creator choose what kind of link they'd like to insert.
but blanket setting all urls to open in a new tab is imo not a good idea.

lagartoverde:
> the WCAG at least _recommends_ limiting new tabs as much as possible: https://www.w3.org/WAI/WCAG21/Understanding/on-focus

I agree with this but I think our users expect them to be opened in a new tab, I mean the testers just report as a bug that the link doesn't open in a new tab, it wasn't even a question of should this open in the same tab or in a new one.

> What I would support is making sure that the editor maintains the target attribute on links when importing html, and add a feature to the link editor to let the document creator choose what kind of link they'd like to insert.
but blanket setting all urls to open in a new tab is imo not a good idea.

Giving the users the option would be great, I will comunicate this with the OP team and see what they think

abeforgit:
To be fair that's just the tester's opinion :grimacing: Which can make sense for your app but it not something the editor should make decisions about. Postprocessing always remains an option if you really want to force it.

abeforgit:
closing as wontfix, if there are more comments feel free to reopen

PR #1022, opened 2023-11-07
GN-4538: Allow to remove node via button with confirmation modal

dkozickis:
### Overview

Allows to remove node via button with confirmation dialog

https://binnenland.atlassian.net/browse/GN-4583

### Setup

1. Checkout
2. `npm run start` 

### How to test/reproduce

1. Go to "Editable Nodes POC"
2. Insert RdfaTest
3. See below

https://github.com/lblod/ember-rdfa-editor/assets/769698/032ed7c6-ef85-4eff-8c36-90cd98c7c70a

### Challenges/uncertainties

1. I'm not sure it needs a confirmation dialog at all. We're not asking user to confirm when he's deleting with backspace.
2. Changed "Convert to YYY node" button to be blue instead of red, so the "delete" button stands out
3. Figuring out best way to remove node by just pressing backspace/delete, right now when clicking on a node, caret appears at the closest text instead of "selecting" the node, so pressing backspace/delete naturally starts deleting the text.
  * When backspacing into a node we can adopt `selectBlockRdfaNode`
  * When selecting the node - ???

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations


abeforgit:
this just removes the node, and does not remove the relationships on nodes it was linked to, which is kinda the whole point of the button...

abeforgit (changes requested):
see above comment

abeforgit (changes requested):
- doing it in 3 steps means you need 3 undos to undo it, and in between undos you're in an inconsistent state
- you remove the entire node but only clear that node's properties, meaning if any of its children define relationships to other nodes in the doc, they aren't cleared. 
- see rdfaTest sample data, which contains duplicate definitions of the same resource (which is valid). If you remove one resource, it clears all the relationships of the other instance, even the relationships that are still valid

PR #1023, opened 2023-11-09
Hide text cursor when `ProseMirror-hideselection` class is applied to editor content

elpoelma:
### Overview
This PR ensures that the text-cursor in the editor is hidden when the  `ProseMirror-hideselection`  is applied.
This prevents two cursors from appearing at the same time (one real cursor and one gap-cursor for example)

Bef
[gapcursor-before.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/794347b9-bb81-443f-b3da-7814499e0950)
ore:

After:
[gapcursor-after.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/facac391-d597-4ed9-866b-4f550894408f)

None

### How to test/reproduce
- Start the dummy app in a chromium-based browser
- Insert a table
- Set the cursor before the table
- Notice that only one (horizontal) cursor shows up

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/late-cheetahs-share.md
+++ b/.changeset/late-cheetahs-share.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Hide caret when `ProseMirror-hideselection` class is applied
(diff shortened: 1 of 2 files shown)


piemonkey (approved):
Could only reproduce on Chrome but the fix works.

PR #1024, opened 2023-11-09
Wrapping selection in rdfa nodes

piemonkey:
### Overview
Add ability to wrap a selection in a content/literal or resource node.

Part of https://binnenland.atlassian.net/browse/GN-4579

### Setup
In the editable nodes PoC.

### How to test/reproduce
If there is no selection, nothing new appears.
Select any content in the editor and use one of 'Wrap with literal' or 'wrap with resource' from the sidebar. Adding a resource shows a modal as it's necessary to set the uri to use for the resource.

### Challenges/uncertainties
Prosemirror API around selections and commands.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations


addon/commands/_private/rdfa-commands/wrap-literal.ts
```diff
@@ -0,0 +1,27 @@
...
+import { wrapIn } from 'prosemirror-commands';
+import { v4 as uuidv4 } from 'uuid';
+import { findRdfaIdsInSelection } from '@lblod/ember-rdfa-editor/utils/rdfa-utils';
+
+export function wrapLiteral(): Command {
+  return (state, dispatch) => {
+    const childRdfaIds = findRdfaIdsInSelection(state.selection);
+    if (childRdfaIds.size !== 0) {
+      if (dispatch) {
+        console.error('Cannot wrap child rdfa nodes in a literal node');
```
abeforgit, line 11: that's a good pattern, to notify but not throw if you're running the command without checking :+1: 
We should use it more often

addon/utils/_private/rdfa-utils.ts
```diff
@@ -179,3 +173,47 @@ export function getResources(node: PNode): Set<string> {
...
+      return true;
+    },
+  );
+  return result;
+}
+
+/**
+ * Get all of the literal nodes which are children of the given node
+ */
+export function getChildLiterals(node: PNode) {
```
abeforgit, line 200: Seems to me this generates external properties for all descendant rdfa-aware nodes (not just literals, also resource nodes)
If intentional, the function could perhaps use a better name and docstring

addon/commands/_private/rdfa-commands/wrap-resource.ts
```diff
@@ -0,0 +1,67 @@
...
+    };
+    const wrapArgs: Parameters<typeof wrapIn> = [
+      state.schema.nodes.block_rdfa,
+      attrs,
+    ];
+    if (!wrapIn(...wrapArgs)(state)) {
+      return false;
+    }
+    if (dispatch) {
+      const objectId = uuidv4();
```
abeforgit, line 25: very useful to have the editor mint a new URI for you, but there will be cases where we will want to explicitly provide a pre-existing URI (e.g. when referring to external uris)

the command could perhaps take in a flag/optional arg to allow this (the UI also needs to have a toggle)

I'd wrap the URI generating logic in a util as well to make sure we use the same logic everywhere (even though it's only two lines)

If you'd prefer, you can do this in a separate PR, I'll leave that up to you

abeforgit (changes requested):
overal very solid and just about ready to merge, just a few comments/suggestions

addon/utils/_private/rdfa-utils.ts
piemonkey: Hmm, I was thinking that we wouldn't want to automatically link a wrapping resource to child resources, just literals, but I'm not sure what makes most sense now. Maybe this is one to decide after actually using the editor to make a big template. I'll at least ake the name and doc consistent for now.

addon/commands/_private/rdfa-commands/wrap-resource.ts
piemonkey: I'll do it in this one.

PR #1028, opened 2023-11-13
GN-4593: update `html-input-parser` to take rdfa-specific attributes into account.

elpoelma:
### Overview
This PR includes some changes to the `html-input-parser` implementation:
- Removal of all references to lump tags
- Include exhaustive list of allowed tags (sourced from https://github.com/cure53/DOMPurify/blob/1.0.8/src/tags.js)
- Include exhaustive list of allowed attributes (partly sourced from https://github.com/cure53/DOMPurify/blob/1.0.8/src/attrs.js). This list includes the default allowed attributes from `dompurify`, as well as rdfa-specific attributes, in order to ensure these are not removed.

This PR mostly ensures that rdfa-attributes are not stripped away on paste.

[GN-4593](https://binnenland.atlassian.net/browse/GN-4593?atlOrigin=eyJpIjoiNWQ1YTkzYWM1YjczNDgxZmE4NTFkYTVmZWRmMTdjNWUiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the dummy app
- Insert the sample decision
- Copy/cut the whole document
- Paste it
- Notice that the rdfa now stays preserved.

### Challenges/uncertainties
- Should we also run the `html-input-parser` (dompurify) when setting the content of a document using `initialize` or `setHTMLContent`?
- Is it necessary to keep track of a seperate `HTMLInputParser` class, we could just call dompurify directly instead?

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
We could do a simpler API like this:
image
It would allow users of this API to also pass all configuration options that are accepted by `dompurify`

abeforgit:
> We could do a simpler API like this: image It would allow users of this API to also pass all configuration options that are accepted by `dompurify`

sure, but let's cross that bridge when we come to it

PR #1035, opened 2023-11-17
GN-4583: Remove node properties on backspace

dkozickis:
### Overview

Performs correct action of removing node properties/backlink on backspace press when selection is present.

Also remove mentions of "alternative backspace", as it won't be "alternative" anymore 🤔 

[GN-4583](https://binnenland.atlassian.net/browse/GN-4583)

### Setup

1. Checkout

### How to test/reproduce

1. Use `RdfaTest` 
2. Select an RDFA node (e.g. Cmd + Click it)
3. Press backspace
4. Observe correct properties/backlinks were deleted
5. Observe selection was deleted/modified correctly
6. Select multiple RDFA nodes with text in between
2. Press backspace
3. Observe correct properties/backlinks were deleted
4. Observe selection was deleted/modified correctly

### Challenges/uncertainties

1. As we have discussed - this is not the best way to remove properties/backlinks on RDFA nodes being deleted, as this does not react on delete, selecting and typing in. To fix the above I am currently working on using `appendTransaction` with modification of `changedDescendants`
2. I have tried to implement a visible selector for the user, so he can press it directly to select the node, but I've decided against it, as it was not working properly. It was depending on `handleDOMEvents` and using [posAtDOM](https://prosemirror.net/docs/ref/#view.EditorView.posAtDOM) detect what node to select, but in the end it was not the best implementation, as `posAtDOM` was not alway selecting what I hoped it would, even with the `bias` param adjustments. Better way here would be to implement an `EmberView` for RDFA nodes, then selecting a node will be a breeze. Users still can use "Cmd + Click".
3. 
   ```ts 
   isolating: true,
   selectable: true,
   ```
   is a free way to select node on backspace, cannot not like it :)  

### Checks PR readiness
- [X] changelog
- [X] npm lint - target branch is failing
- [X] no new deprecations

### DIFF
--- a/addon/components/editor.ts
+++ b/addon/components/editor.ts
@@ -14,7 +14,6 @@ import { getOwner } from '@ember/application';
 import Owner from '@ember/owner';
 import { DefaultAttrGenPuginOptions } from '@lblod/ember-rdfa-editor/plugins/default-attribute-value-generation';
 import SayController from '@lblod/ember-rdfa-editor/core/say-controller';
-import { KeymapOptions } from '@lblod/ember-rdfa-editor/core/keymap';
 
 export interface RdfaEditorArgs {
   /**
@@ -33,7 +32,6 @@ export interface RdfaEditorArgs {
     [node: string]: NodeViewConstructor;
   };
   defaultAttrGenerators?: DefaultAttrGenPuginOptions;
-  keyMapOptions?: KeymapOptions;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


abeforgit (changes requested):
https://github.com/lblod/ember-rdfa-editor/assets/16419862/7914125e-55fe-460f-bc72-689efa253859

See recording: cursor ends up in wrong place after backspace

abeforgit:
@dkozickis confirmed that my comment is unrelated to your work

PR #1041, opened 2023-11-21
GN-4583: Remove node properties/backlink pointing to removed RDFA node

dkozickis:
### Overview

GN-4583: Remove node properties/backlink pointing to removed RDFA node
Also remove mentions of "alternative backspace", as it won't be "alternative" anymore 🤔 

[GN-4583](https://binnenland.atlassian.net/browse/GN-4583)

### Setup

1. Checkout

### How to test/reproduce

1. Use `RdfaTest` 
2. Select an RDFA node (e.g. Cmd + Click it)
3. Press backspace
4. Observe correct properties/backlinks were deleted
5. Observe selection was deleted/modified correctly
6. Select multiple RDFA nodes with text in between
2. Press backspace
3. Observe correct properties/backlinks were deleted
4. Observe selection was deleted/modified correctly
5. Select multiple RDFA nodes with text in between
2. Enter some text, e.g. "hello"
3. Observe correct properties/backlinks were deleted
4. Observe selection was deleted/modified correctly

### Challenges/uncertainties

2. I have tried to implement a visible selector for the user, so he can press it directly to select the node, but I've decided against it, as it was not working properly. It was depending on `handleDOMEvents` and using [posAtDOM](https://prosemirror.net/docs/ref/#view.EditorView.posAtDOM) detect what node to select, but in the end it was not the best implementation, as `posAtDOM` was not alway selecting what I hoped it would, even with the `bias` param adjustments. Better way here would be to implement an `EmberView` for RDFA nodes, then selecting a node will be a breeze. Users still can use "Cmd + Click".
3. 
   ```ts 
   isolating: true,
   selectable: true,
   ```
   is a free way to select node on backspace, cannot not like it :)  

### Checks PR readiness
- [X] changelog
- [X] npm lint - target branch is failing
- [X] no new deprecations


addon/commands/_private/rdfa-commands/remove-backlinks-targeting-resource-property.ts
```diff
@@ -0,0 +1,65 @@
...
+  resource: string;
+  /**
+   A transaction to use in place of getting a new one from state.tr
+   This can be used to call this command from within another, but care must be taken to not use
+   the passed transaction between passing it in and when the callback is called.
+   */
+  transaction: Transaction;
+};
+
+export const removeBacklinksTargetingResourceProperty = ({
```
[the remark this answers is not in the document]
elpoelma: It's a bit of a long name, but I think it is clear and can be left as is :)

addon/commands/_private/rdfa-commands/remove-properties-targeting-backlinks.ts
```diff
@@ -0,0 +1,61 @@
+import { Backlink } from '@lblod/ember-rdfa-editor/core/rdfa-processor';
+import { Command, Transaction } from 'prosemirror-state';
+import { getNodesByResource } from '@lblod/ember-rdfa-editor/plugins/rdfa-info';
+import {
+  getProperties,
+  getResource,
+} from '@lblod/ember-rdfa-editor/utils/_private/rdfa-utils';
+
+export const removePropertiesTargetingBacklinks = ({
```
elpoelma, line 9: I would just rename this to either `removePropertyTargetingBacklink` or `removePropertiesTargetingBacklink`. Just to be consistent :)

addon/plugins/remove-properties-of-deleted-nodes/index.ts
```diff
@@ -0,0 +1,131 @@
...
+          properties.forEach((property) => {
+            const removeBacklinksCommand =
+              removeBacklinksTargetingResourceProperty({
+                resource,
+                property,
+                transaction: sharedTransaction,
+              });
+
+            removeBacklinksCommand(nextState, (transaction) => {
+              nextState = newState.apply(transaction);
```
elpoelma, line 85: I think this should be:
```suggestion
              nextState = nextState.apply(transaction);
```

addon/plugins/remove-properties-of-deleted-nodes/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (node.attrs['__rdfaId']) {
+      prevNodesById.set(node.attrs['__rdfaId'] as string, node);
+    }
+  });
+
+  newState.doc.descendants((node) => {
+    if (node.attrs['__rdfaId']) {
+      nextNodesById.set(node.attrs['__rdfaId'] as string, node);
+    }
+  });
```
elpoelma, line 30: These two mappings are (I think) often calculated twice: as the `newState.doc` becomes the `oldState.doc` in the next iteration. We might be able to cache the mappings?

PR #1043, opened 2023-11-22
Support more complex rdfa aware nodespecs

piemonkey:
### Overview
- Biggest change here is to add a way to pass an array instead of single child node to `renderRdfaAware`.
- Renamed last trailing references to 'content' nodes to 'literal' nodes
- Improved some types and the name of a type (deprecated the old one to make it backwards compatible).

Will be needed for a PR in plugins, will link here when it's ready.

### Setup
N/A

### How to test/reproduce
It shouldn't add anything new, so test that everything works. The only new thing will only be used by the new plugins code.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/app/styles/ember-rdfa-editor/_c-annotation.scss
+++ b/app/styles/ember-rdfa-editor/_c-annotation.scss
@@ -62,6 +62,8 @@ $say-annotation-background-color: var(--au-gray-100) !default;
   [typeof]:not([typeof='foaf:Document']),
   [data-type],
   .mark-highlight-manual {
+    border-bottom: 0.2rem dotted var(--au-gray-300);
+    padding-bottom: 0.2rem;
     transition: border var(--au-transition);
 
     // Hide certain underlines
--- a/tests/dummy/app/templates/editable-node.hbs
+++ b/tests/dummy/app/templates/editable-node.hbs
@@ -5,7 +5,6 @@
   <:content>
     <EditorContainer
       @editorOptions={{hash
-        showRdfa="true"
         showRdfaHighlight="true"
         showRdfaHover="true"
         showPaper="true"


app/styles/ember-rdfa-editor/_c-annotation.scss
```diff
@@ -62,8 +62,6 @@ $say-annotation-background-color: var(--au-gray-100) !default;
   [typeof]:not([typeof='foaf:Document']),
   [data-type],
   .mark-highlight-manual {
-    border-bottom: 0.2rem dotted var(--au-gray-300);
```
elpoelma, line 65: What is the reason for the style changes?

elpoelma (approved):
Looks good overall!

abeforgit (approved):

app/styles/ember-rdfa-editor/_c-annotation.scss
piemonkey: The dotted line is no longer necessary to show what is contained within rdfa elements, as we now have the boxes to show this. Having both looks very broken. I guess maybe we should toggle whether this is displayed based on whether we're using the rdfa-aware editor or not?

elpoelma: It should also be possible to control these styles using the `rdfa-annotations` class. Which you can control using `editorOptions.showRdfa`. I suspect we will also remove a lot of the css of these rdfa-annotations, as the blocks we have now are IMO a better solution.

elpoelma: I would drop the style change from this PR, we can still remove it later and you can turn it off by setting `editorOptions.showRdfa` to `false`.

piemonkey: I've reverted the css change and removed `showRdfa` from the editable nodes route.

PR #1044, opened 2023-11-23
GN-6404: fix behaviour of embedded editor keymap

elpoelma:
### Overview
This PR solves issues with the keymap initialization of an embedded editor instance:
- The default keymap of an embedded editor is now correctly initialized
- An additional `shift+Enter` handler has been added to the default embedded editor keymap
- Only run `leave-on-enter` modifier when `Enter` has been pressed without modifier keys (`Ctrl`, `Alt`, `Shift` or `Cmd`)

[GN-6404](https://binnenland.atlassian.net/browse/GN-4604?atlOrigin=eyJpIjoiMGRiMjJmNTdhNDY0NDRjZmE0NDY0ZDEzNjgwZTQ2NDgiLCJwIjoiaiJ9)

### How to test/reproduce
**Before this fix**
- Set the allowed content of a link to `inline*`
- Insert a link with some text inside
- Notice that none of the editor shortcuts work inside the link (`Mod+b`, `Enter`, `Mod+i` etc.)

**After this fix**
- Set the allowed content of a link to `inline*`
- Disable the `leave-on-enter` modifiers in the link component
- Insert a link with some text inside
- Notice that the editor shortcuts work inside the link (`Mod+b`, `Enter`, `Mod+i` etc.), as well as `shift+Enter`

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/eighty-balloons-pump.md
+++ b/.changeset/eighty-balloons-pump.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Fix issue with setting keymap of embedded editor
--- a/.changeset/giant-hairs-applaud.md
+++ b/.changeset/giant-hairs-applaud.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+Add shift+enter handler to `embeddedEditorBaseKeymap`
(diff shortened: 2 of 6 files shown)


abeforgit:
gonna merge this but some notes:
- adding newline at end of node looks like it does nothing the first time, but this is because browsers dont' render a trailing <br> -> I don't think this is a regression
- I misunderstood the fix, my idea was to have leave-on-enter behavior for enter, and hard-break on shift-enter. But I thought this because I thought we added leave-on-enter for variable nodes as well as links, it seems we didn't.

PR #1045, opened 2023-11-23
fix(placeholders): fix placeholders not being selectable

abeforgit:
-------

### Overview

After the improvements to the gapcursor/position resolution from mouseclicks, placeholder nodes
threw erros when clicked on. This was because the position resolution logic tries to
find a position with a parent that is the same parent that the text cursor would have if
it would be standing where you clicked. It does this by walking upwards and comparing the value
of parent with the value of the node at the "inside" position (which prosemirror already calculates).
However, in the case of atom nodes, the first position checked already starts with a parent that
is the parent of the atom, meaning the walk upwards will never result in the correct position.
Additionally, the loop was missing a bounds check for depth=0, which caused the error

### Setup

best way to test is npm link in the plugin repo, and try out the article structures and their
placeholders

### Challenges/uncertainties

Switching on the node being an atom is an educated and afaict a relatively conservative bet.
There might be more scenarios where the walk upwards will not find the right position.
Additionally this may result in some more scenarios where a good gapcursor can not be found,
since I don't have a reproduction of the cases the original code was trying to solve.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/lovely-rivers-kneel.md
+++ b/.changeset/lovely-rivers-kneel.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+fix placeholders no longer being selectable with mouse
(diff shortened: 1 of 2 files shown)


elpoelma (approved):
Looks good! I would just move the depth check to the while loop condition, as calling `result.before` also fails at depth 0.

PR #1046, opened 2023-11-24
UI updates of relationship-editor modal

elpoelma:
### Overview
This PR includes a few UI adjustments to the relationship-editor modal, including:
- Seperate dropdowns for selecting a resource or literal in a document
- Adjustments to the radio-button components
- Slight refactor of some properties in the relationship-editor components

**Before**
image

**After**
image
image

### How to test/reproduce
Check if adding relationships still works as before (existing resource/literal, new resource/literal)

### Challenges/uncertainties
- I included some UI adjustments to make it (hopefully) a bit clearer to the user what they are selecting, especially by seperating the dropdowns for resources and literals. This is by no means a final design and feedback is very welcome.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
> for literals: select box still shows "select existing or create new one" but there's no option to create a new one anymore
> 
> I think the difference between selecting an existing resource and making a new one is a bit unintuitive: if you select one, the previously "required" uri base dissappears
> 
> we lost (or never had in this view?) the ability to specify an absolute URI for a new resource
> 
> In general I'm not quite sure of the value of making a new node to begin with in this particular modal. It's not super intuitive _where_ the new node will appear (it does appear where the cursor _was_, but I feel I've already forgotten my cursor context when I open this modal) I find the make first (with the wrapping buttons) -> then link approach more intuitive (but that's subjective, not against keeping this flow)

I agree that we should probably not include the functionality to create new nodes in this modal, I just wanted to keep feature parity with the previous version. I can update this PR to remove the functionality to create new resource/literal nodes.

elpoelma:
@abeforgit I updated this PR to remove the option to create new literal/resource nodes.

piemonkey:
Possibly not for this PR, more for future: Hmm, I'm thinking that as we have now cut out some of the functionality of this, it's possibly small enough to live in the sidebar? I was tempted to try that in the first place, but it seemed too overloaded. I think if we're not going for that then maybe we should add something to give the context of what we're adding to, e.g. the URI of the resource we're adding the relationship to.

Another maybe future thing: It probably makes sense to have either 'add relationship' or another 'add backlink' button (which uses much the same functionality) to add a backlink for a literal/resource. This is assuming that when creating a document you might find a node that is missing a backlink and currently you would need to then find the subject resource to add the relationship.

One actually for this PR: Maybe we should filter the existing resources to not include the current resource? I don't think relationships with the same subject and object make sense?

Otherwise, looks good.

addon/components/_private/rdfa-editor/relationship-editor/modal.hbs
```diff
@@ -24,76 +20,51 @@
...
-              />
-            {{/let}}
-          </AuFormRow>
-        {{/if}}
+        <AuFormRow @alignment="default">
+          {{#let (unique-id) as |id|}}
+            <AuLabel for={{id}} @required={{true}} @requiredLabel="Required">
+              Object
+            </AuLabel>
+            <PowerSelect
```
piemonkey, line 48: When the PowerSelect is open, if I type anything in the console, I get an error
```
Assertion Failed: <PowerSelect> If you want the default filtering to work on options that are not plain strings, you need to provide `@searchField`
```
We should either configure filtering or I think we can disable it using `onInput`.

elpoelma: Nice catch! I'll look into it.

PR #1051, opened 2023-11-30
GN-4588: improved document attributes support

elpoelma:
### Overview
This PR introduces a few new changes/features to improve support for custom document attributes and improved UI to edit/view document attributes. It also includes changes to support rdfa properties/backlinks of the document node.
Specifically, this PR includes the following changes:
- Add option for the document (top) node to be parsed using `parseDOM` parse-rules. When setting the content of a document, using either `setHTMLContent` or `initialize`, three options are possible:
  * The `topNode` (often `doc`) has no parse-rules: a default node of type `topNode` is created (without any attributes). The html provided to `setHTMLContent` or `intialize` is parsed as its content.
  * The `topNode` has 1 or more parse-rules: the parser searches the provided html for a node that matches a parse-rule of the `topNode`.
     - If a node is found: the node is parsed as the `topNode` and its content is parsed as the `topNode` content
     - If a node is not found: a default `topNode` node is created. The html provided to `setHTMLContent` or `intialize` is parsed as its content.
- Add `TransformUtils.setAttribute` utility function which is a wrapper for either updating the attribute of a document or a node.
- Add custom `SetDocAttributesStep` which allows for overwriting the `attrs` object of the doc-node.
- Update `rdfa-commands` to take document attributes into account
- Extend default `doc` node-spec to include rdfa-related attributes: `properties`, `backlinks`, `resource`, `rdfaNodeType` and `__rdfaId`.
- Remove `extraAttributes` parameter/option of `doc` node-spec. The `doc` node-spec can be easily extended/overriden using spread operators/custom `node-specs`. (:boom: breaking)
- When using the `setHTMLContent` method to also update the attributes of the topNode correctly. This should make `setHTMLContent` more-or-less equivalent with the `initialize` method. The main difference is that `initialize` creates a new state and `setHTMLContent` does not.
- Allow setting the `topNode` node as active (editable) node.
- Update rdfa components and attribute-editor to take the `topNode` node into account.
- Addition of a POC `document-language` pill to the editor toolbar. It opens a modal in which it is possible to select the document language.
- Addition of a `document-info` toolbar pill. It opens a modal in which it is possible to edit the attributes of the `topNode` node.
- Addition of a sample `RdfaWithDocAttrs` document

[GN-4588](https://binnenland.atlassian.net/browse/GN-4588?atlOrigin=eyJpIjoiNTZkMDhkMTE0ZmFmNGExOWE2YTFkMDM3YWJiM2EyODMiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the dummy app and visit the `Editable Nodes POC` route
- Select the `RdfaWithDocAttrs` document
- When setting the cursor at the root of the document, notice that it is possible to view info about the document and update the attributes/properties/backlinks
- Notice that it is possible to add relationships to other nodes
- Using the document-language pill, you should be able to change the language of the document
- Using the `Document Info` pill, you can open a modal to view/edit attributes of the document
- The attributes/properties/backlinks of the document are correctly serialized and parsed

### Checks PR readiness
- [x] changelog
- [x] npm lint


abeforgit:
first thing I noticed is you added the doc attr step back, do we need it?

elpoelma:
@abeforgit I updated this PR to set an `AllSelection` when jumping to the `doc` node. `AllSelections` have no visual feedback yet, but we can change that of course.

PR #1052, opened 2023-11-30
Add support for `about` attributes when parsing rdfa relationships

piemonkey:
### Overview
Adds `about` to the parsing of rdfa to establish properties and backlinks.

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/358

### Setup
Easiest used with the PR from plugins with article structure plugin changes.

### How to test/reproduce
In the latest version of that PR, look at the properties and backlinks for `say:heading` nodes. After refresh, these were not correctly maintained, this PR corrects that.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -75,6 +75,8 @@ export function getRdfaAttrs(node: Element): RdfaAttrs | false {
           logger('rdfaNodeType is not a valid type', value, node);
         }
         if (type === 'resource') {
+          // Attrs of resource nodeTypes have non-optional resource and properties fields, so
+          // ensure that they're set, either to the already parsed values or []/''
           attrs = {
             ...attrs,
             rdfaNodeType: type,


addon/core/schema.ts
```diff
@@ -55,38 +55,60 @@ export function getRdfaAttrs(node: Element): RdfaAttrs | false {
...
+                  : '',
+              properties:
+                attrs.properties && attrs.properties instanceof Array
+                  ? (attrs.properties as Property[])
+                  : [],
+            };
+          }
+          break;
         }
-        if (type === 'resource') {
```
abeforgit, line 77: I know this is already merged, but I didn't notice before

so if I understand correctly this is making sure resource nodes _have_ a resource attribute, even if there is none in the dom, right?
It's a bit janky cause it relies on the ordering of the rdfaDomAttrs constant, and Object.keys looping over the keys in order (which is probably in the spec, but still it's a bit fragile)

I'd do these checks after the loop to make it a bit more clear

addon/core/schema.ts
```diff
@@ -55,38 +55,60 @@ export function getRdfaAttrs(node: Element): RdfaAttrs | false {
...
-                ? attrs.resource
-                : '',
-            properties:
-              attrs.properties && attrs.properties instanceof Array
-                ? (attrs.properties as Property[])
-                : [],
-          };
+        case 'about': {
+          const about = value as unknown as string;
+          const property = node.attributes.getNamedItem('property')?.value;
```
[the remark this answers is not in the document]
piemonkey: Hmm, so the issue I'm seeing is that if I insert a structure, say a chapter, then the chapter resource has heading and title properties, each of which have the correct backlinks. If I then refresh, the heading is no longer correctly linked to the chapter (either the property on the chapter or the backlink on the heading.

It may well be that there is an issue with the parser that means it's missing these relations (they're in the DOM with an `about` and `property` attribute on the heading.

addon/core/schema.ts
abeforgit: ah yeah it would indeed be fine cause we don't overwrite when there's no value, so even if we hit resource later in the loop we would either overwrite it with the right value, or not do anything
I'd still put it outside of the loop personally, but its preference

piemonkey: Fair. The typing is easier this way though, so I'll leave it here. As soon as I set `rdfaNodeType` to be `resource`, it insists that the others are present.

PR #1054, opened 2023-11-30
Feature: inline rdfa nodes

abeforgit:
### Overview

Adds a new nodespec: inline_rdfa (which previously existed as a mark).
It works basically the same as the existing resource/literal nodes.
Adds equivalent wrapping commands as well

### Challenges/uncertainties

~~when 2 inline rdfa nodes are nested, trying to get a text selection _inside_ the nested node is
for some reason very difficult, it keeps selecting the entire node. I have no idea why.
Making the nodes not selectable is an option I'm seriously considering but it has its drawbacks.
I'm also not sure the toDOM spec is correct, seeing as in the inspector the inline node span gets
the draggable attribute for some reason, even though it shouldn't be draggable.
I'm gonna need some help to look into this.~~

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


abeforgit:
Note: i'm aware the "goto" relationship buttons don't work with inline nodes, but I don't have the time to keep looking for a solution

PR #1058, opened 2023-12-06
Improve behaviour of `delete` and `backspace` keys around atomic inline nodes

elpoelma:
### Overview
This PR attempts to provide improved behaviour of the `delete` and `backspace` keys around atomic inline nodes in a document. 
Specifically, it includes the following:
- Custom `select-node-forward` and `select-node-backward` commands with included support for atomic inline nodes. The commands are adapted from the [prosemirror-commands](https://github.com/ProseMirror/prosemirror-commands) package. These commands are tied to the `delete` and `backspace` keys respectively.
- An update to the logic of entering an embedded-editor with arrow-keys. Aside from clicking inside the embedded-editor, it is now only possible to enter an embedded-editor with the `arrowleft` and `arrowright` keys.

None

### How to test/reproduce
**Behaviour for inline nodes**
- Start the dummy application
- Insert a link and type some text in the link
- Notice that when putting the cursor after/before the link and pressing backspace/delete, the link gets selected and not directly removed. Another backspace/delete keypress deletes the link.

**Behaviour for block nodes (should be the same as before)**
- Insert a table node
-  Notice that when putting the cursor after/before the table and pressing backspace/delete, the table gets selected and not directly removed. Another backspace/delete keypress deletes the table.

### Challenges/uncertainties
- This PR is marked as `breaking`, not because it changes the API of this package, but because of the modified default behaviour around atomic inline nodes.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/curvy-cherries-sniff.md
+++ b/.changeset/curvy-cherries-sniff.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": major
+---
+
+Update logic of entering embedded-editor with arrow-keys
--- a/.changeset/green-files-flash.md
+++ b/.changeset/green-files-flash.md
@@ -1,0 +1,7 @@
+---
+"@lblod/ember-rdfa-editor": major
+---
+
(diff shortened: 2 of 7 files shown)


elpoelma:
Also created a PR on embeddable: https://github.com/lblod/frontend-embeddable-notule-editor/pull/187

PR #1067, opened 2023-12-12
Make rdfa preprocessor handle more cases correctly

abeforgit:
### Overview

How the preprocessor works:

some definitions:

- say you have a triple `<uri> predicate "object"`, where `"object"` is the value of an attribute (such as `content`), we call the tuple `(predicate, "object")` a **property** of `<uri>`
- if you have a triple `<uri> predicate <uri2>`, we call the tuple `(predicate, <uri2>)` a **relationship** of `<uri>`
- if you have a triple `<uri> predicate "object"`, where `"object"` is derived from the concatenated text content of a `node`, we call the tuple `(predicate, node)` also a **relationship** of `<uri>`

The core idea of the preprocessor is this: turn a random rdfa document, which has resources and predicates all over the place, into a document with 2 types of nodes:

- a resource node: defines a resource and, crucially, holds ALL properties and relationships of that resource, in both directions
- a literal node: the "manifestation" of a literal relationship on a resource

now, this split maps nicely on the spec for the rdfa parsing algorithm. Essentially, when visiting a node, there are 4 possible outcomes (that are relevant for us):

- a. the node defines a new subject (we mark it as a resource node)
- b. it "closes" a triple by defining its value with the concatenated textcontent (we mark it as a literal node)
- c. it does both a and b (we mark it as a literal node, see below)
- d. it does anything else (we don't mark it)

in the case of d, we can safely ignore it because of how the resource and literal nodes get processed later on. We attach all their properties and relationships after we fully parse the document, and the parser doesn't ignore d, so it captures all cases.
This also means that before the nodes ever get to prosemirror, all the relationships and properties are correctly attached, taking the entire document into account. 

### Special note
the case where an element has both an `about` and a `property` attribute, (this is case "c" from above) it will always get parsed as a literal node, even if it technically also defines a subject. This means there can be literal nodes
that point to a resource which does not have a resource node in the document. This doesn't break anything and is actually completely sensible. A node like this effectively defines a complete triple,
completely independant from the context it's in, with the value of `about` as the subject, the value of `property` as predicate and the concatenated textcontent as the object.

Because of this, I've also added a message in the link-ui that informs the user that no node was found for a backlink when it is clicked. It's an info message, not an error, cause it should signify
the above valid case.

I could have disabled the link instead when no valid node could be found, but that's quite costly as we'd run over the entire document every state change, and defeats the point of
calculating the resource mapping in a lazy way. It's a rare case anyway, so I much prefer this solution.

Note this also signals a legitimate usecase for a literal->resource-with-nested-literal conversion feature.

Once a resource node for the resource in question is created, the link "goto" behavior will work again, as it's completely dynamic. The only thing that will be missing is the outgoing link on the newly
created node. This link will get populated on refresh, but should also be populated on insert. However, it's a rare case, so I'm keeping that out of scope here.

### How to test/reproduce

try to come up with weird rdfa edge cases and see if they get parsed/serialized correctly, using [rdfa play](https://rdfa.info/play) as a reference
note: inline rdfa isn't parsed yet in this branch, so use blocks only

### Challenges/uncertainties

`in-list` is not yet supported (it may work, I didn't try yet)

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/proud-balloons-complain.md
+++ b/.changeset/proud-balloons-complain.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Improved parser based on a better definition of resource and literal nodes
--- a/addon/components/_private/rdfa-editor/relationship-editor/index.hbs
+++ b/addon/components/_private/rdfa-editor/relationship-editor/index.hbs
@@ -88,6 +88,17 @@
       <p>This node does not have any backlinks</p>
     {{/if}}
   </div>
+  {{#if this.statusMessage}}
+  <div>
+    <AuAlert
+      @skin={{this.statusMessage.type}}
+      @closable={{true}}
+      @onClose={{this.closeStatusMessage}}
+    >
+      {{this.statusMessage.message}}
+    </AuAlert>
+  </div>
+  {{/if}}
 </AuContent>
 
 <this.Modal
--- a/addon/components/_private/rdfa-editor/relationship-editor/index.ts
+++ b/addon/components/_private/rdfa-editor/relationship-editor/index.ts
@@ -30,17 +30,29 @@ type Args = {
   node: ResolvedPNode;
 };
 
+interface StatusMessage {
+  message: string;
+  type: 'info' | 'warning' | 'error' | 'success';
+}
+interface StatusMessageForNode extends StatusMessage {
+  node: PNode;
+}
+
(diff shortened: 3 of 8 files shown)


piemonkey:
Lgtm. One thing I realised is that maybe the long comment explaining this in the PR should live in the code?

abeforgit:
good call, I'll add it

PR #1074, opened 2023-12-18
GN-4614: automatically add http or mailto to links

piemonkey:
### Overview
Uses Linkify.js to match links and add http or mailto when relevant. Also fixes a few usability bugs around links:
- Fix a crash after reloading the page and using the sidebar
- Some missing translations
- Selection when inserting a link (so hitting insert and then typing writes the text of the link instead of replacing the new link node with the character typed)

https://binnenland.atlassian.net/browse/GN-4614

### Setup
N/A

### How to test/reproduce
Type a valid or invalid link (or email) into either of the href boxes (in pop-over or sidebar) and see that the http: or mailto: part is added automatically.

### Challenges/uncertainties
There are more usability problems I wanted to fix, but I wasn't able to figure out how to do them as embedded-editor nodes seem to have problems around reactivity in the ember components (e.g. changing a property with a `@tracked` decorator not updating in the UI).

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/mean-dancers-agree.md
+++ b/.changeset/mean-dancers-agree.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+When adding the href for a link, automatically add `http://` or `mailto:` to the href attribute to generate a valid `<a>` tag.
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -67,7 +67,11 @@ module.exports = {
         semi: 'off',
         '@typescript-eslint/no-unused-vars': [
           'error',
-          { argsIgnorePattern: '^_' },
+          {
+            argsIgnorePattern: '^_',
+            args: 'after-used',
+            ignoreRestSiblings: true,
+          },
         ],
         '@typescript-eslint/semi': ['error', 'always'],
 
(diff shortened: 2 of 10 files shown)


elpoelma (changes requested):
The automatic conversion of incomplete links works as expected (tested with both http and mailto links). 

The automatic focussing of the inner editor on creation does not completely work as expected: 
- I don't know if we can focus the innerview in the `didInsertContentWrapper` method? That way it would only run on creation of the innerview.
- When first pressing the `arrowUp` key and then inserting a link, the link is not autofocussed, as the condition in the `onSelected` handler fails.

Two possible solutions:
- Focus the innerview (if needed) in the `didInsertContentWrapper` method
- Rename `selectInnerOnCreation` to `focusInnerOnSelection` and always focus the innerview when the link is selected.

elpoelma (commented):
This solution looks good to me! The only thing I would change about it is that we set the selection by default at the start (instead of at the end).

piemonkey:
Hmm, I'm not sure which is most intuitive, but I also think that other than using arrow keys to navigate, this logic would only be used in very edge cases or when there is no content, so no difference. So I've updated the default to be `atStart`.

elpoelma (approved):

PR #1079, opened 2023-12-29
Improve handling of datatype

piemonkey:
### Overview
I wasn't going to make this a PR to a PR but then I realised that some parts of it are questionable, so we should think about which of the commits to keep.

In trying to correct the handling of `datatype` attributes when parsing RDFa, I realised that the datatype `XMLLiteral` has the difficult side-effect that any changes to the HTML content of that element are reflected in the RDF data, so it will be incredibly hard to pass an element with this attribute through the editor without modifying the data. These changes are enough to make a very minimal blackbox test work, but they will fail as soon as there are elements which are picked up as RDFa aware within the element with datatype `XMLLiteral`.

I'm not sure of the value of value of having this artificially simple example working, but some of the parts should be kept and it highlights the need to handle this in some way for this kind of test, either by handling triples with an object which is an `XMLLiteral` typed string differently (or ignoring them), or by stripping this attribute for these tests.

Made to #1077.
#1086 split out and those commits removed from this PR.

### How to test/reproduce
Changes most easily seen in tests, but can also copy the `dataset` test from `test-cases.ts` into `sample-data.ts` to insert it into the editor dummy app.

### Challenges/uncertainties
N/A

### DIFF
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -177,7 +177,7 @@ export function renderRdfaAttrs(
             nodeOrMark.attrs.resource) as string,
           property: contentPred.predicate,
           resource: null,
-          datatype: (nodeOrMark.attrs.datatype as string) ?? null,
+          datatype: nodeOrMark.attrs.datatype as string,
         }
       : {
           about: (nodeOrMark.attrs.subject ||
@@ -195,7 +195,7 @@ export function renderRdfaAttrs(
       about: backlinks[0].subject,
       property: backlinks[0].predicate,
       'data-literal-node': 'true',
-      datatype: (nodeOrMark.attrs.datatype as string) ?? null,
+      datatype: nodeOrMark.attrs.datatype as string,
     };
   }
 }
@@ -231,19 +231,21 @@ export function renderRdfaAware({
   delete clone.resource;
   delete clone.__rdfaId;
   delete clone.rdfaNodeType;
-  const renderAttrs = renderRdfaAttrs(renderable);
+  const renderAttrs = { ...clone, ...renderRdfaAttrs(renderable) };
   let { property } = renderAttrs;
-  if (renderable.attrs.datatype === 'rdf:XMLLiteral') {
+  let datatype = renderAttrs.datatype || clone.datatype;
+  if (datatype === 'rdf:XMLLiteral') {
     contentContainerAttrs = {
       ...contentContainerAttrs,
-      datatype: renderable.attrs.datatype,
+      datatype,
       property,
     };
     property = null;
+    datatype = null;
   }
   return [
     tag,
-    { ...clone, ...renderAttrs, property },
+    { ...renderAttrs, property, datatype },
     renderInvisibleRdfa(renderable, rdfaContainerTag, rdfaContainerAttrs),
     [
       contentContainerTag,
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -231,9 +231,19 @@ export function renderRdfaAware({
   delete clone.resource;
   delete clone.__rdfaId;
   delete clone.rdfaNodeType;
+  const renderAttrs = renderRdfaAttrs(renderable);
+  let { property } = renderAttrs;
+  if (renderable.attrs.datatype === 'rdf:XMLLiteral') {
+    contentContainerAttrs = {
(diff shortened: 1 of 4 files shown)


addon/nodes/paragraphWithConfig.ts
```diff
@@ -50,7 +51,7 @@ export const paragraphWithConfig: (
         default: DEFAULT_ALIGNMENT,
       },
       indentationLevel: {
-        default: 0,
+        default: null,
```
elpoelma, line 54: What is the reason to change the default to `null`?

elpoelma (changes requested):
Implementation of supporting the `datatype` attribute looks sane to me overall.
I agree that correctly support the `rdf:XMLLiteral` datatype will be quite hard, as we have several prosemirror nodes that may serialize just a bit differently.

I also don't know if it is very important to support this datatype completely correctly. I agree that moving it to the `content-container` attributes already helps a ton.

I also added some additional comments.

addon/nodes/paragraphWithConfig.ts
piemonkey: The `toDOM` checks whether `indentationLevel` is an integer and adds an attribute if it is, so if we default to 0, it adds a `data-indentation-level=0` attribute to every paragraph, unless we specifically set `indentationLevel` to be null. This is part of minimising changes to the DOM that have no practical input, so reducing the changes if they're inside an element with a `rdf:XMLLiteral` datatype.

addon/nodes/paragraphWithConfig.ts
```diff
@@ -96,18 +97,18 @@ export const paragraphWithConfig: (
     ],
     toDOM(node) {
       const { alignment, indentationLevel } = node.attrs;
-      let style = '';
+      const attrs: ParagraphDataAttributes = {};
       if (alignment && alignment !== DEFAULT_ALIGNMENT) {
-        style += `text-align: ${alignment}`;
+        attrs.style = `text-align: ${alignment}`;
```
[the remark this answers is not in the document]
piemonkey: Fair. I'll separate that commit.

addon/core/schema.ts
```diff
@@ -224,9 +231,19 @@ export function renderRdfaAware({
   delete clone.resource;
   delete clone.__rdfaId;
   delete clone.rdfaNodeType;
+  const renderAttrs = renderRdfaAttrs(renderable);
+  let { property } = renderAttrs;
+  if (renderable.attrs.datatype === 'rdf:XMLLiteral') {
```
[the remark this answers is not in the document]
piemonkey: I've added a commit which addresses this, as well as no datatype being set in many cases.

addon/core/schema.ts
```diff
@@ -231,19 +231,21 @@ export function renderRdfaAware({
   delete clone.resource;
   delete clone.__rdfaId;
   delete clone.rdfaNodeType;
-  const renderAttrs = renderRdfaAttrs(renderable);
+  const renderAttrs = { ...clone, ...renderRdfaAttrs(renderable) };
   let { property } = renderAttrs;
-  if (renderable.attrs.datatype === 'rdf:XMLLiteral') {
+  let datatype = renderAttrs.datatype || clone.datatype;
```
elpoelma, line 236: Is the second part (`clone.datatype`) necessary here?, as `clone` is already included in `renderAttrs` right?

piemonkey: Part of the problems I'm working around here is that datatype can be set to `null` in the return of `renderRdfaAttrs`, which means that `renderAttrs.datatype` is null, even if `clone.datatype` is defined. Maybe worth a comment in the code?

abeforgit:
I think we'll need proper support for datatypes on all properties and also on literal nodes, including the ability to change them on the fly
the xml type is definitely the worst one for us, and at the same time probably the least interesting for real usage (or at least not in a way where one would care too much about the structure changes)
I wanna think about this again in the light of full datatype support, but this is already an improvement indeed

addon/core/schema.ts
piemonkey: Done

piemonkey:
Should I merge this (to the feat/content-predicate branch) or hold off?

abeforgit:
should be superceded by #1130

PR #1083, opened 2024-01-03
GN-4659: do not clean empty elements when they have attributes

elpoelma:
### Overview
This PR includes a possible solution to [GN-4659](https://binnenland.atlassian.net/browse/GN-4659?atlOrigin=eyJpIjoiNjE5YmUzM2I0NzQwNGQzZmFiZTY1ZWRhYmE1ZGY4M2MiLCJwIjoiaiJ9), which describes a bug with empty elements containing relevant attributes being removed when pasting/setting the content of the editor.

This PR ensures that empty elements containing any attribute are not pruned away.

Additionally, this PR includes some changes to tests of the `html-input-parser` module to ensure that empty elements with attributes are allowed.

[GN-4659](https://binnenland.atlassian.net/browse/GN-4659?atlOrigin=eyJpIjoiNjE5YmUzM2I0NzQwNGQzZmFiZTY1ZWRhYmE1ZGY4M2MiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the dummy app
- Create an empty document
- Set the alignment of the first empty paragraph to right/center
- Reload the document
- Ensure that the alignment is not reset

### Challenges/uncertainties
The bug this PR is trying to solve is caused by the fact that the the docx-cleaner is now always executed when:
- Pasting html content to the editor
- When setting the html content of the editor

Two alternative solutions to this problem are:
- Add the p and header tags to the allowed empty elements
- Only run the docx-cleaner when pasting, not when setting the whole content of the editor. (This issue will then still occur when pasting from word/anywhere else)

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/loud-cooks-float.md
+++ b/.changeset/loud-cooks-float.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Do not clean up empty elements when they have attributes
--- a/addon/utils/_private/ce/paste-handler-helper-functions/cleanEmptyElements.ts
+++ b/addon/utils/_private/ce/paste-handler-helper-functions/cleanEmptyElements.ts
@@ -1,22 +1,13 @@
 import { traverseElements } from './traverseElements';
-import { RDFA_ATTRIBUTES } from '../../constants';
 
 const ALLOWED_EMPTY_ELEMENTS = ['BR', 'IMG', 'TR', 'TD'];
 const NOTRIM_ELEMENTS = ['SPAN'];
 
-function hasRdfaAttributes(element: Element) {
-  for (const attr of RDFA_ATTRIBUTES) {
-    if (element.hasAttribute(attr)) {
-      return true;
-    }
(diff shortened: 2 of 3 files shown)


piemonkey:
I can't see any issues with this fix, but I'm not sure why we were stripping empty elements to begin with. Unfortunately I can't see why it was added to the `@prezly/docx-cleaner` package that it was taken from as the git history of that project seems mangled.

elpoelma:
> I can't see any issues with this fix, but I'm not sure why we were stripping empty elements to begin with. Unfortunately I can't see why it was added to the `@prezly/docx-cleaner` package that it was taken from as the git history of that project seems mangled.

As I understand it the clipboard html data that is coming out of word may contain several 'useless' html elements, like empty `span` elements that just have a style attribute containing a font-family. IMO it is not a big problem to not remove these, as a lot of them will still be filtered away when converting them to prosemirror nodes.

PR #1084, opened 2024-01-03
Fix text and highlight colour toggles when selecting multiple table cells

piemonkey:
### Overview
Fix setting text colors and highlights when multiple table cells are selected (instead of just toggling for the last cell).

This issue was actually in the implementation of the set-color and set-highlight commands as they were not handling some selection types well. In switching to a command which handles these selections well, I realised that our command was based on a prosemirror internal command which [incorrectly treats marks with different attributes as equal](https://github.com/ProseMirror/prosemirror/issues/771). I modified this command to do a shallow comparison of the mark's attributes, so e.g. red and green highlights are treated as if they are different marks, as the user would expect.

https://binnenland.atlassian.net/browse/GN-4661

### Setup
N/A

### How to test/reproduce
Select multiple table cells and choose a highlight or text colour, see that the colour is applied to all cells. See video in [Jira issue](https://binnenland.atlassian.net/browse/GN-4661) for illustration.

Secondary improved behaviour, selecting text that already has a highlight (or colour) applied, and applying the same highlight again, now clears the highlight instead of having no effect. This mimics e.g. applying bold. If any part of the selection does not have the highlight, the highlight is applied to the whole selection instead of being cleared.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/core/say-controller.ts
+++ b/addon/core/say-controller.ts
@@ -1,12 +1,13 @@
 import { SayStore } from '@lblod/ember-rdfa-editor/utils/_private/datastore/say-store';
 import Owner from '@ember/owner';
 import { unwrap } from '@lblod/ember-rdfa-editor/utils/_private/option';
+import { shallowEqual } from '@lblod/ember-rdfa-editor/utils/_private/object-utils';
 import { datastoreKey } from '@lblod/ember-rdfa-editor/plugins/datastore';
 import { rangeHasMarkEverywhere } from '@lblod/ember-rdfa-editor/commands/toggle-mark-add-first';
 import SayView from '@lblod/ember-rdfa-editor/core/say-view';
 import SayEditor from '@lblod/ember-rdfa-editor/core/say-editor';
 import { tracked } from '@glimmer/tracking';
-import { MarkType, Schema } from 'prosemirror-model';
+import { Attrs, MarkType, Schema } from 'prosemirror-model';
 import {
   Command,
   EditorState,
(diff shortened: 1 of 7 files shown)


elpoelma (approved):
Includes some very nice changes! :) 
Maybe it is also possible to include similar adaptions to the `isMarkActive` method on the controller? Currently it only checks the mark for the first range, but does not take selections with multiple range (e.g. table-cell selections) into account.

addon/core/say-controller.ts
```diff
@@ -96,11 +97,12 @@ export default class SayController {
...
-  isMarkActive(markType: MarkType) {
+  isMarkActive(markType: MarkType, attrs?: Attrs) {
     const state = this.activeEditorState;
     const { from, $from, to, empty } = state.selection;
     if (empty) {
-      return !!markType.isInSet(state.storedMarks || $from.marks());
+      const mark = markType.isInSet(state.storedMarks || $from.marks());
+      return !!mark && (!attrs || shallowEqual(attrs, mark.attrs));
     } else {
       return rangeHasMarkEverywhere(state.doc, from, to, markType);
```
elpoelma, line 107: In this case, we could loop over all the ranges of the selection, as well as pass the optional `attrs` object to `rangeHasMarkEverywhere`

piemonkey:
I've added this behaviour to that function, but at the moment it's only used in one place, where there are no attributes applied to the mark, so that path isn't actually used. It'll be useful for later though.

addon/core/say-controller.ts
elpoelma: It will fixed the status of the bold toggle when having multiple table cells selected in which not all of them are bold.

piemonkey: Oops, yes. Fixed now.

piemonkey:
As discussed, I'll merge this and imitate the logic for handling multiple selection ranges from this PR in a new PR to improve this behaviour for marks using `isMarkActive`.

PR #1086, opened 2024-01-03
Only set attributes on paragraphWithConfig node when non-default

piemonkey:
### Overview
Was originally part of #1079 but separated as it does not rely on RDFa work.

Prevent empty style attributes and data attributes with indentationLevel = 0 by default for paragraph nodes.

Cherry-picked from #1079.

### Setup
N/A

### How to test/reproduce
In the editor, press enter to insert a new paragraph and inspect the html. The paragraph should just be a `<p>` element rather than `<p data-indentation-level=0 style="">`.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/eight-moons-pump.md
+++ b/.changeset/eight-moons-pump.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+When creating paragraph elements, do not set a style attribute if there is no style, instead of setting an empty attribute.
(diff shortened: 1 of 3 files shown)


elpoelma (commented):
Good solution for preventing style attributes.

On the `data-indentation-level` attribute: we could also keep track of a constant `DEFAULT_INDENTATION_LEVEL = 0;` (similar to the default alignment), set this as the default of the `indentationLevel` attribute, and not serialize the `indentationLevel` if it is equal to the default indentation level. This way, paragraphs that were indented but are unindented back to indentation-level 0 do not have their indentation-level serialized either,

elpoelma (approved):

PR #1088, opened 2024-01-04
Absolutely position toolbar dropdowns to avoid unreachable items

piemonkey:
### Overview
Based on a ticket which was from Loket, which was including the editor by importing the `ember-rdfa-editor` package and giving it a very small box to render in, but is no longer using the editor.

Only relevant in embedded scenarios with very short editor windows.

This uses the floating UI option to use the absolute positioning strategy rather than the velcro default of fixed. This means the browser handles the case that the dropdown extends past the extent of the editor div in a way that at least means all the controls are accessible.

https://binnenland.atlassian.net/browse/GN-4014

### Setup
Modify `app/styles/ember-rdfa-editor/_c-dummy.scss` to remove `overflow: hidden` from the `.c-dummy` class (L10) and set `height: 200px` instead of 100% for `.c-dummy__content` (L81).

### How to test/reproduce
When opening the text style dropdown, the menu should now extend beyond the editor display.

### Challenges/uncertainties
It's unclear where this will be useful as it's no longer used in Loket any more.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/healthy-papayas-check.md
+++ b/.changeset/healthy-papayas-check.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Allow dropbown menus to extend beyond editor container when embedded as a very small window
--- a/addon/components/toolbar/dropdown.hbs
+++ b/addon/components/toolbar/dropdown.hbs
@@ -1,6 +1,6 @@
 
 <div class="say-dropdown" ...attributes>
-  <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} as |velcro|>
+  <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} @strategy="absolute" as |velcro|>
     <button
       {{this.reference}}
       class='say-dropdown__button {{if this.dropdownOpen "is-active" ""}}'


elpoelma:
It seems to me that it is mostly the `overflow: hidden` rule that causes the popup menu to disappear when the height of the editor is very small. It doesn't seem to me that it is caused by `position: fixed` or `position: absolute`. 

On the dummy-app of embeddable, this issue is simply resolved by removing `overflow: hidden` from the dummy app css.

The original line of css which caused this issue (see https://github.com/lblod/ember-rdfa-editor/blob/e4b35ec8da2cb7268d2e04dc4f8e5041a199b55c/app/styles/ember-rdfa-editor/_c-container.scss#L17) has also been removed a few months ago.

So I think this issue is actually solved already. The developers using the editor package can choose themselves if they add `overflow: hidden` or not.

elpoelma:
What we could do is add an `overflow-y: scroll` (or something similar) to the dropdown menus in order to ensure that the users can scroll through the menu when the window height itself is too short.

piemonkey:
Sorry, my reproduction instructions weren't helpful. You're right that the most common cases are fixed by not setting `overflow: hidden;`, but changing for the absolute rather than fixed strategy means that the small browser window situation works correctly.

Unfortunately, trying to get the menu to scroll is tricky as the menu is a child of the header bar, so it has no information about how much space it has to render in. As well, the sizing of the buttons is using flex, so any attempt to restrict the size of the menu would mean the buttons are very compressed, so for this route we'd need to change their styling.

For a too small window:
- As it is, with the 'fixed' strategy, the menu is not part of the layout, so it simply extends off the edge of the window. Scrolling never brings the menu into view.
- With the 'absolute' strategy, scrolling on the menu scrolls the whole window (not just the editor part), so the menu can at least be scrolled into view.

elpoelma:
> Sorry, my reproduction instructions weren't helpful. You're right that the most common cases are fixed by not setting `overflow: hidden;`, but changing for the absolute rather than fixed strategy means that the small browser window situation works correctly.
> 
> Unfortunately, trying to get the menu to scroll is tricky as the menu is a child of the header bar, so it has no information about how much space it has to render in. As well, the sizing of the buttons is using flex, so any attempt to restrict the size of the menu would mean the buttons are very compressed, so for this route we'd need to change their styling.
> 
> For a too small window:
> 
>     * As it is, with the 'fixed' strategy, the menu is not part of the layout, so it simply extends off the edge of the window. Scrolling never brings the menu into view.
> 
>     * With the 'absolute' strategy, scrolling on the menu scrolls the whole window (not just the editor part), so the menu can at least be scrolled into view.

Ok I see. While the 'absolute' strategy causes the whole window to scroll indeed. It also causes the following issue (recorded on chrome):
[Screencast from 2024-01-08 13-32-20.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/9f25d34e-bad3-4942-b45b-1904351b628b)

elpoelma:
> Sorry, my reproduction instructions weren't helpful. You're right that the most common cases are fixed by not setting `overflow: hidden;`, but changing for the absolute rather than fixed strategy means that the small browser window situation works correctly.
> 
> Unfortunately, trying to get the menu to scroll is tricky as the menu is a child of the header bar, so it has no information about how much space it has to render in. As well, the sizing of the buttons is using flex, so any attempt to restrict the size of the menu would mean the buttons are very compressed, so for this route we'd need to change their styling.
> 
> For a too small window:
> 
>     * As it is, with the 'fixed' strategy, the menu is not part of the layout, so it simply extends off the edge of the window. Scrolling never brings the menu into view.
> 
>     * With the 'absolute' strategy, scrolling on the menu scrolls the whole window (not just the editor part), so the menu can at least be scrolled into view.

For the scrollbar: we can set `flex-shrink` to 0 on the different dropdown items, which should fix the shrinking (I think).

Of course, I don't think anyone will use the editor with such a small window size...

piemonkey:
Yeah, the weird layout isn't great but I figured it's better than having a menu that you can't use. Happy to close this if we're not going to use it.

elpoelma:
> Yeah, the weird layout isn't great but I figured it's better than having a menu that you can't use. Happy to close this if we're not going to use it.

True, @abeforgit what do you think?

elpoelma:
> Was the consensus to merge this then? If not we should just close this as it's not like we can't just do it again if relevant.

Yes you can merge :)

PR #1089, opened 2024-01-04
Keep paragraph config e.g. indent when including soft breaks

### LINKED ISSUE #1076: When using a soft break within an indented paragraph, the indentation gets removed when loading said paragraph into the editor
Environments where this bug was reproduced:

**OS:** MacOS
**Browser:** Chrome & Firefox

Package version this was tested on: `8.1.0`

Steps to reproduce: 
- Open the dummy application.
- Input some text and indent it using the indentation tool within the toolbar.
- Add a soft break to the indented text using `shift + enter` and add some more text.
- Reload the dummy application.
- The indentation (and thus the `data-indentation-level` value) gets removed.

Remark: When not using a soft break within a paragraph, the indentation works as expected.

Example video:

https://github.com/lblod/ember-rdfa-editor/assets/18174827/c4cc702a-dd89-402f-b2df-5dd9ddeee8f1

piemonkey:
### Overview
As reported in #1076 we were losing information from `paragraphWithConfig` nodes when they contain `<br>` tags. This is due to a parse rule which skips any `<p>` tags which contain block content (or rather non-phrasing content), but which was treating `<br>` incorrectly. This now correctly treats `<br>` as phrasing content. I also added some comments to clarify what this code is doing.

Fixes #1076

### Setup
N/A

### How to test/reproduce
As indicated in #1076 :

- Open the dummy application.
- Input some text and indent it using the indentation tool within the toolbar.
- Add a soft break to the indented text using shift + enter and add some more text.
- Reload the dummy application.
- The indentation (and thus the data-indentation-level value) gets removed.

### Challenges/uncertainties
It's unclear the reason why `<p>` elements with block content are being ignored as `paragraphWithConfig` nodes. Ideally this would be in a comment in the code.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/nodes/paragraphWithConfig.ts
+++ b/addon/nodes/paragraphWithConfig.ts
@@ -66,6 +66,9 @@ export const paragraphWithConfig: (
           }
           return false;
         },
+        // If this rule matches (a paragraph with block node content),
+        // the paragraph itself is skipped (but it's content is still parsed).
+        // Paragraphs with block content are not allowed in the HTML spec.
         skip: true,
       },
       {
--- a/addon/nodes/paragraphWithConfig.ts
+++ b/addon/nodes/paragraphWithConfig.ts
@@ -11,14 +11,7 @@ export type ParagraphDataAttributes = {
 
 export type ParagraphNodeSpec = NodeSpec & { subType: string };
 
-let BLOCK_SELECTOR = '';
-NON_BLOCK_NODES.forEach(
-  (tag) => (BLOCK_SELECTOR = `${BLOCK_SELECTOR}${tag}, `),
-);
-// Also include br tag as while it's normally handled as a block by the editor, it's also used for
-// 'soft breaks' (when adding a newline with shift-enter), so we need to include it here.
-BLOCK_SELECTOR = `:not(${BLOCK_SELECTOR} br)`;
-
+const BLOCK_SELECTOR = `:not(${NON_BLOCK_NODES.join(', ')})`;
 const BASE_PARAGRAPH_TYPE = 'paragraph';
 const matchingSubType = (node: HTMLElement, subType: string) => {
   // basic paragraph has no subtype in its dataset and an empty subType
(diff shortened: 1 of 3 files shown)


addon/nodes/paragraphWithConfig.ts
```diff
@@ -59,14 +58,11 @@ export const paragraphWithConfig: (
         getAttrs(node: HTMLElement) {
           const nonBlockNode = node.querySelector(BLOCK_SELECTOR);
```
elpoelma, line 59: We should probably rename this variable to `blockNode`.

elpoelma (changes requested):
This solution works, but I think it would be a bit more clear to add the `br` tag (and other missing phrasing content tags found on https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) to the `NON_BLOCK_NODES` set. 

The reason why `p` tags containing block nodes are skipped (only their content is parsed) because that is not allowed in the html spec. I agree that we could add a comment to clarify this.

piemonkey:
The reason I didn't just add it to `NON_BLOCK_NODES` is because that's generated from the `PHRASING_CONTENT` list of element types, which is used elsewhere and also has it's own comment explaining why this differs from the html spec. `NON_BLOCK_NODES` isn't used elsewhere that I can see, but it seems strange to use a different logic for the two lists.

For me, even with the comment, the lack of clarity as to why these nodes are skipped still remains. The confusion isn't about why block nodes within a `<p>` makes the content invalid. The confusion is what the correct behaviour should be. Should we ignore the `<p>` element or should we ignore the block content that it contains?

addon/nodes/paragraphWithConfig.ts
piemonkey: Hmm, good point.

elpoelma (approved):
Looks good now!

PR #1098, opened 2024-01-08
Improvements to the UX of links

piemonkey:
### Overview
Some UX improvements I wanted to add when working on setting hrefs from urls. When doing that ticket I struggled to make these work, so gave up. I re-visited them to see if there was something missing from ember-nodes, but it turns out I was just misusing them. I also made 2 very minor fixes by tweaking the layout.

In terms of UX, this to me feels much more natural (or at least close to other editors). If we're unsure about any parts of it, we can split either of the 2 UX tweaks or 2 layout fixes by removing some commits.

Issues fixed:
https://binnenland.atlassian.net/browse/GN-4504
https://binnenland.atlassian.net/browse/GN-4480

### Setup
N/A

### How to test/reproduce
See linked issues for reproduction instructions for those. For the UX tweaks:
- Type a valid url or email address as normal text, select it, then click the 'insert link' button on the toolbar. The link should use the url (with added http:// or mailto:) as the href, instead of requiring to manually enter both the text and the href.
- With no selection, click the insert link button, then enter a valid url or email address (without changing the selection). This will set this to be the text content of the link. Then click somewhere else to de-select the link. This will automatically set the href to be the one from the text content.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/commands/wrap-selection.ts
```diff
@@ -20,11 +27,16 @@ export function wrapSelection(nodeType: NodeType): Command {
...
+      let wrappers = findWrapping(nodeRange, nodeType);
       if (!wrappers) {
         return false;
       }
       if (dispatch) {
+        if (attrsFromWrapped) {
+          const attrs = attrsFromWrapped(nodeRange);
+          const newWrap = findWrapping(nodeRange, nodeType, attrs);
+          wrappers = newWrap ?? wrappers;
+        }
```
elpoelma, line 39: Would it not be better to move this part before the dispatch and calculating the wrapping only once? (based on if `attrsFromWrapped` have been passed or not).

elpoelma (changes requested):
The two new features look very sane to me from a UX point of view, they both work as expected in firefox and chromium.

I'm not fully convinced on the css solutions for the [GN-4504](https://binnenland.atlassian.net/browse/GN-4504) and [GN-4480](https://binnenland.atlassian.net/browse/GN-4480) issues. While they seem to solve [GN-4504](https://binnenland.atlassian.net/browse/GN-4504), I can still reproduce  [GN-4480](https://binnenland.atlassian.net/browse/GN-4480) (in firefox). 

I think it would be better to find the underlying cause of both issues, which is caused by the same/similar bug in firefox: sometimes a cursor is shown in an input/contenteditable without it being correctly focused.

A similar issue caused by the same underlying bug in firefox can be reproduced by typing some text in a link, adding a newline/line before the link, and pressing arrow-up/arrow-down.

[Screencast from 2024-01-09 10-51-20.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/9a893e37-549b-415e-a383-d36ed7a917fa)

These issues do not occur on chromium, as far as I tested.

addon/commands/wrap-selection.ts
piemonkey: My thinking was that since `attrsFromWrapped` is passed in, we don't have any way to know if it's heavy (or has side-effects, etc.), so we should only run it if we're actually `dispatch`ing a result rather than checking if the command can run. This is probably over thinking it though and it would be better being easier to read.

elpoelma: I can see that, but I suppose it is then our responsibility to pass in a function that is not too heavy.

abeforgit (approved):
afaict the css changes don't have a negative impact, and the new ux is very nice, so good to merge for me

PR #1124, opened 2024-01-16
Fix `looseMatches is undefined` error

elpoelma:
### Overview
This PR adds an empty array fallback when initializing the variable `looseMatches` in the `decentLocaleMatch` function, to prevent it from having the `undefined` value.

### How to test/reproduce
Before this fix:
- Open the dummy app
- Set the browser locale to an unsupported language (e.g. french)
- Observe the `looseMatches` is undefined error

After this fix:
- Open the dummy app
- Set the browser locale to an unsupported language
- The error should no longer show up, the language shown is the default language (english)

### DIFF
--- a/.changeset/poor-socks-pay.md
+++ b/.changeset/poor-socks-pay.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Fix `looseMatches` is undefined error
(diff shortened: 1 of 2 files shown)


elpoelma:
> Argh, it's been a while since this one got me. Damn `Record<string, T>` types. I don't think a typescript flag exists for this and I don't know of an eslint rule... I should also not make last minute tweaks and not re-test everything...

no worries :) 
I think adding `"noUncheckedIndexedAccess": true` to the `tsconfig.json` might help.

PR #1127, opened 2024-01-22
GN-4657: Serialize table column width to exported HTML

dkozickis:
### Overview

GN-4657: Serialize table column width to exported HTML

* Exporting `colgroup` element in the `serialize` method of `table` node, as it contains all the information to control width of displayed columns
* Expand `showExportPreview` of `DebugTools` to "copy" CSS from parent window (editor), so the preview is closed to the "real" one
* Normalize `rows`/`columns` before calling `insertTable`. `tableAddRows`/`tableAddColumns` become `string` in the input. It was also possible to delete columns/rows and click on the insert table, the error was thrown in console without any indication to the user.

https://binnenland.atlassian.net/browse/GN-4657

### Setup

1. Checkout

### How to test/reproduce

1. `npm run start`
2. http://localhost:4200/
3. Insert the table, resize the columns, click "Show Export Preview" 🙏 

### Challenges/uncertainties

1. Last column alway takes up remaining space, might feel a bit off in different environments, but this is what we decided to do 🤔 

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/.changeset/three-lamps-ring.md
+++ b/.changeset/three-lamps-ring.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Mark paragraphs as `defining`, as they may contain important attributes (such as alignment and indentation) that should not be lost when pasting
--- a/.changeset/wild-eggs-tap.md
+++ b/.changeset/wild-eggs-tap.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Remove custom `transformPasted` hook as it conflicts with the `defining` node-spec attribute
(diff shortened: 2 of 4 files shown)


addon/components/debug-tools.ts
elpoelma, line 1: This approach does not seem to work on firefox, only on chromium based browsers.

elpoelma (changes requested):
Seems to work well overall. The different column widths (except for the last one of course) are correctly exported.

Adding the styles to the export preview is a nice addition, but the implementation does not work on firefox, only on chromium

addon/components/debug-tools.ts
elpoelma: Using the `wnd.document.write` methods does seem to have an effect on firefox.

PR #1130, opened 2024-01-24
Support datatype and language

abeforgit:
### Overview

Refactors the "property" and "backlink" interfaces, now renamed to "OutgoingTriple" and "IncomingTriple" respectively.

The main difference is that there is no more property "type", but the `object` of the property is now an object (instead of a string) that adheres closely to the RDFjs `Term` interface.

There are 6 types of objects for outgoing triples:
- Literal
- NamedNode
- BlankNode (with limited support)
- ContentLiteral
- ResourceNode
- LiteralNode

and 2 for incoming triples:
- ResourceNode
- LiteralNode

which should be pretty self-explanatory.

In particular, Literal, LiteralNode and ContentLiteral objects now also have `datatype` and `language` attributes, again analogous to the rdfjs interfaces (this is of course the main reason for the PR).

These attributes are now also parsed and serialized correctly.

To support this, there is also a major overhaul included of the add/edit modals for both "properties" and "relationships". I've kept the separation still in the UI.

For this, I made a generic form that can handle all types of outgoing properties.

Additionally, cause I needed _some_ validation anyway, I went ahead and added the [yup](https://github.com/jquense/yup/tree/master) validation library, along with schemas for each of the object types, and even a custom `curie` validator for predicates, namednode object values, and datatypes.

Was this a smart time to do this? absolutely not. But I fell in the form rabbithole and couldn't get out. At least it will prevent us from chasing down bugs arising from invalid RDFA values. And I did enjoy using the library, it's very well typed and has a good set of builtin validators, and has builtin support for async validation.

As a bonus, it is now also possible to edit relationships.

### Checks PR readiness

- [x] npm lint


elpoelma (commented):
It seems that 'content-literals' are not correctly serialized/parsed. Content-literals are reparsed as normal literals.

addon/core/rdfa-processor.ts
```diff
@@ -6,41 +7,99 @@ import {
...
+  | 'LiteralNode'
+  | 'ResourceNode'
+  | 'ContentLiteral'
+  | 'BlankNode';
+
+export type SayRDFLiteral = Pick<
+  RDF.Literal,
+  'value' | 'termType' | 'language'
+> & {
+  datatype: Omit<RDF.NamedNode, 'equals'>;
```
elpoelma, line 26: What is the main reason for omitting the `equals` method?

addon/core/rdfa-processor.ts
```diff
@@ -6,41 +7,99 @@ import {
...
+  | NamedNodeTriple
+  | BlankNodeTriple
+  | ResourceNodeTriple
+  | LiteralNodeTriple
+  | ContentTriple;
+
+export type IncomingResourceNodeTriple = {
+  termType: 'ResourceNode';
   subject: string;
   predicate: string;
```
elpoelma, line 91: Wondering if we should also use `RDF.NamedNode` objects for the predicates/subjects.

addon/core/rdfa-processor.ts
abeforgit: so we can work with plain POJOs in the forms, having to add the equals method everywhere means we need to use the Dataset constructors from rdfjs everywhere

abeforgit:
> It seems that 'content-literals' are not correctly serialized/parsed. Content-literals are reparsed as normal literals.

yep working on it, cheers

addon/core/rdfa-processor.ts
abeforgit: we could... it's just they can't be anything else so it's maybe overkill

addon/utils/_private/rdfa-parser/rdfa-parser.ts
```diff
@@ -964,6 +964,8 @@ export class RdfaParser<N> {
     node: N,
     activeTag: IActiveTag<N>,
     attributes: Record<string, string>,
+    contentDatatype?: ModelNamedNode<N>,
```
elpoelma, line 967: Are these necessary?

addon/core/rdfa-processor.ts
elpoelma: Fair enough, it doesn't really seem necessary in this case indeed.

addon/utils/_private/rdfa-parser/rdfa-parser.ts
abeforgit: yes, its how I distinguish between normal triples and contentliterals

abeforgit:
still issues with linking to literalNodes

addon/utils/_private/rdfa-parser/rdfa-parser.ts
abeforgit: sorry that's actually not true, these indeed aren't necessary, it's only in the resourceNode function that they matter

elpoelma:
[Screencast from 2024-01-26 15-33-36.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/bf0b82c1-20c3-43f8-b03a-1c6adc316e71) Found another bug with the datatype parsing/serialization.

abeforgit:
on ti

abeforgit:
@elpoelma should be fixed now

elpoelma:
The issue seems solved indeed :)
image
After a reload, the backlink contains both a language and a datatype property.

When loading the `RdfaTest` sample document, the outgoing title links have the datatype `http://www.w3.org/2001/XMLSchema#string` (even if these do not occur in the source document). After a reload they are gone.

addon/core/rdfa-processor.ts
```diff
@@ -112,10 +168,13 @@ export function preprocessRDFa(dom: Node) {
...
   for (const [node, object] of datastore.getContentNodeMap().entries()) {
-    const { subject, predicate } = object;
-    const incomingProp: Backlink = {
+    const { subject, predicate, datatype, language } = object;
+    const incomingProp: IncomingTriple = {
+      termType: 'LiteralNode',
       subject: subject.value,
       predicate: predicate.value,
+      datatype: datatype ?? { termType: 'NamedNode', value: '' },
+      language: language || '',
```
elpoelma, line 177: We probably need to change this to
```suggestion
      language: datatype ? '' : language ?? '',
```

addon/core/rdfa-processor.ts
```diff
@@ -125,63 +184,112 @@ export function preprocessRDFa(dom: Node) {
...
+      ) {
+        return [
+          {
             predicate: quad.predicate.value,
             object: {
-              type: 'resource',
-              resource: quad.object.value,
+              termType: 'ContentLiteral',
+              datatype: contentDatatype ?? { termType: 'NamedNode', value: '' },
+              language: contentLanguage ?? '',
```
elpoelma, line 270: This probably needs to be changed to:
```suggestion
              language: contentDatatype ? '' : contentLanguage ?? '',
```

elpoelma, line 270: This probably needs to be changed to:

```suggestion
              language: contentDatatype ?'' : contentLanguage ?? '',
```

abeforgit:
no idea where those string types come from
I'm assuming the parser adds them as defaults, ill look into it tomorrow

elpoelma:
> no idea where those string types come from I'm assuming the parser adds them as defaults, ill look into it tomorrow

It seems so: https://www.npmjs.com/package/rdf-data-factory#creating-literals

elpoelma:
I tested with a few other data-factory packages ([@rdfjs/data-model](https://github.com/rdfjs-base/data-model) and [graphy](https://github.com/blake-regalia/graphy.js)) and they gave the same results. So the default `http://www.w3.org/2001/XMLSchema#string` seems to be following spec (as well as the default `langString` datatype when a language is provided)

elpoelma:
What if we take the following approach?:
- When a literal/content-node has a language => set the datatype always to `xml:langString`.
- The user can only set the language of a literal property when the datatype is set to `xml:langString`; or if the datatype is empty (in this case we implicitly set the datatype to `xml:langString`. => we disable the language input-element if datatype is other than `xml:langString`/empty.
- When creating a new literal property, prefill the datatype with `xml:string`; or leave it empty and implicitly set the datatype  to `xml:string` when no other datatype is provided and no language is provided.

This way we still follow spec (default datatype of `xml:string` or `xml:langString` in case of provided language), if another datatype is provided, do not allow the language attribute.

abeforgit:
so what I ended up doing was just using a proper datafactory implementation: https://github.com/rubensworks/rdf-data-factory.js

I also updated the types to be more in line with rdfjs, and another small tweak to the IncomingTriple interface to be more consistent as well

I still see a problem with resourcenodes that multiple links to literals on the same predicate, but with different language/datatypes. I don't think thats a regression from these latest commits, I think that's been broken from the beginning. I have an idea where to look but ill do that tomorrow

abeforgit:
I didn't do anything clever in the form, it just shows language if there and datatype if not
I'm relying on the parser to properly ignore language if there is a datatype attribute

I also updated the serializer to prioritize language for that reason
It's a bit mind-bending to think about cause it _seems_ like the opposite of the rdfa spec but it works out in the end.

elpoelma (commented):
Seems to work well overall, but this bug: [Screencast from 2024-01-26 15-33-36.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/bf0b82c1-20c3-43f8-b03a-1c6adc316e71) is back again it seems.

Adding a relationship with datatype `xml:string` is converted to a relationship with `xml:langString` after two reloads.

PR #1133, opened 2024-01-27
Migrate off of ember-cli-typescript

abeforgit:
### Overview

Migrates off of ember-cli-typescript onto the recommended setup. 
Config taken from a new `ember addon` output from the cli, as well as the ember-cli-typescript docs.

Why now: I was having issues with compiling types while working on the plugins with `npm link` and `yalc`, and I wanted to eliminate one of the possible causes. 

in this pr:
- config and package.json changes to match the modern setup
- update all files to comply with 2 new strictness rules: `type` imports and using indexed access for undeclared properties
these were turned on in the recommended config, and for v2 and codesplitting we need the type imports anyway. The conversion was quick as vscode has an automatic fix for these. 

one strictness rule I turned off was no unchecked index access, which causes things like `array[0]` to require a nullcheck. This is more safe, but also a massive pita, so I turned it off  for now
- includes a `yalc`ed version of our fork of prosemirror-invisibles until I can make a pr there. There was a missing config in the package.json which meant ts couldn't find
the types. -> this was actually the real problem that triggered this pr, but it was lost among hundreds of false positives.
- fix some various areas where we had questionable types for a while, but where ts let it slide until now

as far as I know I introduced no behavioral changes

One more thing that's different in this setup is that `ember serve` no longer reports type errors. To get live typechecking, run `npm exec tsc -- --noEmit --watch` in a separate terminal (or just rely on your editor's typescript language server integration)
### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
CI failing due to yalced dep, ill put it in draft mode for now

tsconfig.declarations.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "declarationDir": "declarations",
+    "emitDeclarationOnly": true,
+    "noEmit": false,
+    "rootDir": "."
+  },
+  "include": ["addon", "addon-test-support"]
```
elpoelma, line 9: I think we need to add the `types` folder here too so it picks up our custom declarations.

tests/test-helper.ts
```diff
@@ -2,7 +2,7 @@
 //@ts-nocheck
 // TODO: linting disabled cause of really bad and incomplete qunit typings, fix
 import * as QUnit from 'qunit';
-import Application from 'dummy/app';
+import Application from 'dummy/tests/dummy/app/app';
```
elpoelma, line 5: Why did you include this change. My editor and tsc complains because of this change. It works well with `dummy/app`

elpoelma (changes requested):
Most type errors seem gone :) Only some errors from the `prosemirror-invisibles` left.

tests/test-helper.ts
abeforgit: probably some fat-fingered vim commands if I'd have to guess

abeforgit: I am curious tho why your tsc complained and mine didn't, especially considering this is in a ts-nocheck file

elpoelma: My bad, it wasn't ts complaining, it was a runtime error

elpoelma (approved):
Seems to work well now! :)
Also tested with `npm link` + `tsc:watch` on editor-plugins. This seems to work.
I would just add the `declarations` folder to the `.gitignore`

PR #1135, opened 2024-01-30
GN-4626: Make list styles persist on paste/copy

dkozickis:
### Overview

This PR does following

* Changes how ordered lists receive their styling in the browser - instead of using `data-list-style` attribute we now use `style` attribute with `list-style-type` property directly on the rendered DOM element. This ensures, that when user is copying lists to other editor - the style gets carried over.
* Introduces Cypress Visual Regression Testing (VRT) - I have described Visual Regression Testing in the [README.MD](https://github.com/lblod/ember-rdfa-editor/pull/1135/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R351-R376) ([rendered version](https://github.com/lblod/ember-rdfa-editor/blob/7306d18fa77eb2da0ee611c53e16ca09b879453e/README.md#visual-regression-tests)) for this PR, and adds two first VRT tests for how the lists are rendered.

https://binnenland.atlassian.net/browse/GN-4626

### Setup

1. Checkout
2. `npm run start`

#### Copying lists

1. Create an ordered list
2. Change the type (to e.g. Roman numbers)
3. Copy the list to another editor (e.g. Google Docs)
4. Observe that the type carried over

#### Visual Regression Testing

1. See the README in this PR
2. Try to run `npm run cypress:run:vrt` locally

#### Copying lists

1. In our editor, nested lists take their parent style implicitly via [global CSS](https://github.com/lblod/ember-rdfa-editor/blob/064fec2b19ccdc2b2d48660c72f3bd66a900dc67/app/styles/ember-rdfa-editor/_c-content.scss#L232), because of that the style is not persisted when copying. Still trying to figure out how to approach this. I think it might be worth adding parent attribute inheritance to _our_  version of `sinkListItem`, although similar approach was _removed_ from `prisemirror-schema-list` 5 years ago - https://github.com/ProseMirror/prosemirror-schema-list/commit/c0fd5cea5dc90160b19a3ab1bd9c449a0a849be1

#### VRT

1. Not sure if VRT will work locally in Docker on Linux/Windows machines, had to do with Docker local networking.
6. Need to fine tune acceptable difference, so far it's working fine, but some false positives might start appearing as we add VRT tests.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/tricky-windows-knock.md
+++ b/.changeset/tricky-windows-knock.md
@@ -1,0 +1,8 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+GN-4625: Copy lists to other editors correctly
+
+Adds `list-style-type` to list elements in the rendered HTML, so other editors  
+can render lists with the same style as the editor.
--- a/addon/components/debug-tools.ts
+++ b/addon/components/debug-tools.ts
@@ -71,7 +71,6 @@ export default class RdfaEditorDebugTools extends Component<DebugToolArgs> {
       }),
       parent: element,
     });
-
     this.htmlEditor.dispatch({
       changes: { from: 0, insert: this.formattedDebuggerContent },
     });
--- a/addon/plugins/list/nodes/list-nodes.ts
+++ b/addon/plugins/list/nodes/list-nodes.ts
@@ -9,6 +9,13 @@ type OrderedListAttrs = typeof rdfaAttrs & {
   style?: OrderListStyle;
 };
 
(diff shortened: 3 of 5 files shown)


abeforgit (changes requested):
Afaict it doesn't actually work?

input:
image
output:
image

abeforgit (approved):
yeah mustve been looking at the wrong tab or smthing

PR #1137, opened 2024-02-07
GN-4625: Move to Playwright E2E

dkozickis:
### Overview

Migrate to using Playwright instead of Cypress.

???

### Setup

1. Checkout
2. `npm install`

#### Available commands

* `e2e:open` - opens Playwright UI locally 
* `e2e:open:docker` - opens Playwright UI in Docker and attempts to connect to your XServer 
* `e2e:run` - runs Playwright tests locally, skipping tests tagged with `@vrt`
* `e2e:run:docker` - runs all Playwright tests in Docker
* `e2e:run:vrt` - runs visual regression tests in Docker
* `e2e:run:vrt:update` - updates visual regression tests after running them in Docker
* `e2e:show-report` - shows the report of the last run

#### Visual regression tests (VRT)

To run visual regression tests locally you need to run `npm run e2e:run:vrt`.  
Afterwards if you need to update the snapshots you can run `npm run e2e:run:vrt:update`.

> [!IMPORTANT]
> Playwright is run in Docker to ensure consistent snapshot comparison results, as it guarantees the same environment for each run.

### Challenges/uncertainties

Not sure about behaviour on other host systems, e.g. Linux or Windows.

@abeforgit I am interested to see if you are able to run Dockerized version on your machine with XServer, if you have time to check 🤔 

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/sharp-hornets-love.md
+++ b/.changeset/sharp-hornets-love.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+GN-4660: Percentage resize for tables instead of absolute pixels
--- a/addon/plugins/table/commands/insertTable.ts
+++ b/addon/plugins/table/commands/insertTable.ts
@@ -2,10 +2,9 @@ import { Command } from 'prosemirror-state';
 
 import { unwrap } from '@lblod/ember-rdfa-editor/utils/_private/option';
 import { PNode } from '@lblod/ember-rdfa-editor';
-import { getEditorViewWidth } from '@lblod/ember-rdfa-editor/utils/_private/editor-view';
 
 export function insertTable(rows: number, columns: number): Command {
-  return (state, dispatch, view) => {
+  return (state, dispatch) => {
     const {
       schema,
       selection: { $from },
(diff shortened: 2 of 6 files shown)


abeforgit (commented):
so the x11 forwarding setup is not working for me and I couldnt find an answer within the hour I gave myself, but thats not blocking for this pr to merge if it works for @piemonkey and @elpoelma 

I do know elena is on wayland, so she might have issues too, but afaik the dockerized setup only matters for VRT right? for functionality tests running playwright natively should be enough?

elpoelma:
I justed tried the x11 forwarding setup. I doesn't seem to work on wayland (as expected, maybe it could be possible with xwayland), but it also does not seem to work on x11.

I'm now looking into running playwright locally, which also has some caveats it seems.

elpoelma:
Update: got it to work on fedora wayland natively (not yet using the x11 forwarding). Was confused on ubuntu being the only 'officially' supported distro, but it also works fine on other distros.

elpoelma:
> @elpoelma good to merge?

Good to merge for me. I can run it locally as expected, and the docker set-up works (albeit without x11 forwarding, but not super crucial IMO). 

One final question: doesn't the fact that playwright uses its own custom browser binaries make the vrt tests also run on the 'same' environment every time. That should be reliable enough no?

elpoelma (approved):

elpoelma:
> > One final question: doesn't the fact that playwright uses its own custom browser binaries make the vrt tests also run on the 'same' environment every time. That should be reliable enough no?
> 
> The rendering is still going to be different between different OS

true, true.

PR #1138, opened 2024-02-09
GN-4660: Percentage resize for tables instead of absolute pixels

dkozickis:
### Overview

Percentage resize for tables instead of absolute pixels

Changes to our fork of `prosemirror-tables` here - https://github.com/lblod/prosemirror-tables/pull/2

Fixes

* https://binnenland.atlassian.net/browse/GN-4660
* https://binnenland.atlassian.net/browse/GN-4657 (will need a change in [GN](https://github.com/lblod/frontend-gelinkt-notuleren/pull/627))

### Setup

1. Checkout
2. `npm i`

### How to test/reproduce

1. `npm run start`
2. Insert table, resize column, see if that still works 😅 

### Challenges/uncertainties

Inserting a lot of extra column and resizing them might yield strange results, as they don't have "real width", the browser just makes the best effort to render them, so they might seem unresizeable initially, but resizing them actually works.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/sharp-hornets-love.md
+++ b/.changeset/sharp-hornets-love.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+GN-4660: Percentage resize for tables instead of absolute pixels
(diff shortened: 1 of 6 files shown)


elpoelma (changes requested):
AFAICT it does not really work as expected? The resizing behaviour seems to function incorrectly on my end:
[Screencast from 2024-02-12 15-44-55.webm](https://github.com/lblod/ember-rdfa-editor/assets/24712836/df783eb7-956c-4a4f-abaa-bb99b469cf2f)

Maybe I'm doing something wrong in the configuration on my end?

Update: it was on my end, nuking the `node_modules` seemed to help.

elpoelma (approved):

PR #1140, opened 2024-02-14
Add questionable hacks for NamedNodes with external subjects

piemonkey:
### Overview
Draft PR as I thought I should put it somewhere... I thought this was necessary for the blackbox tests on the plugin repo, but it seems that isn't the case. It is necessary in order to compare parsing on rdfa.info/play however...

https://github.com/lblod/ember-rdfa-editor/pull/1139
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/387
https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/388

### Setup
N/A

### How to test/reproduce
Import the HTML from any article structures into the editor, then compare the exported HTML to the original in rdfa.info/play.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


addon/utils/_private/rdfa-parser/rdfa-parser.ts
```diff
@@ -1297,7 +1304,8 @@ export class RdfaParser<N> {
   ) {
     // Validate IRIs
     if (
-      (subject.termType === 'NamedNode' && subject.value.indexOf(':') < 0) ||
+      // More hacking... Does including '#' make sense?
+      (subject.termType === 'NamedNode' && subject.value.indexOf(':') < 0 && subject.value.indexOf('#')) ||
```
elpoelma, line 1308: I don't think this is necessarily correct, as IRIs do not need to contain a hash symbol, that is just an often used delimiter.

piemonkey: This isn't requiring them to contain `#`, the logic is the other way around. If it does not contain `:` and does not contain `#`, then it's not an IRI, so ignore it. This is here to catch the case that we have in article structure nodes where the subject is `#` (or at least, that's what we have after my hack to handle the nodes with `resource="#"`).

elpoelma: Ok I see what you mean, was a bit confused. Would if not be necessary to negate the `indexOf('#')` then?

piemonkey: Oh, good spot. It was meant to mimic the `:` test, so `subject.value.indexOf('#') < 0`. What I wrote is nonsense as it's only false if it's testing a string that starts with `#`...

abeforgit:
closing to clean up pr list

PR #1142, opened 2024-02-15
Add option for inlining styles for table borders

piemonkey:
### Overview
Allows control of table border styles by (dev) users of the editor. When no style is passed, it defaults to the current behaviour, which is the CSS applies a border in the editor but this is not serialized/exported. If a style is passed, it's serialized.

[PR to embeddable to expose this](https://github.com/lblod/frontend-embeddable-notule-editor/pull/226)
https://binnenland.atlassian.net/browse/GN-4697

### Setup
Inlined styles can be tweaked in the schema for test pages of the dummy app.

### How to test/reproduce
Create a table and see the border styles are applied correctly and then click 'show export preview' or copy-paste it into e.g. google docs. With no style passed, behaviour should be unchanged, i.e. thin border in the editor, nothing in the copy-pasted content.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/table/index.ts
+++ b/addon/plugins/table/index.ts
@@ -33,22 +33,12 @@ export class TableView extends PluginTableView {
   }
 
   private addAttrs(node: Node): void {
-    const {
-      class: nodeClasses,
-      style,
-      ...attrs
-    } = node.attrs as Record<string, unknown>;
+    const { class: nodeClasses, style } = node.attrs as Record<string, unknown>;
     if (typeof nodeClasses === 'string') {
       this.table.classList.add(...nodeClasses.split(' '));
     }
     if (typeof style === 'string') {
       this.table.style.cssText = `${this.table.style.cssText} ${style}`;
-    }
-    for (const attrName in attrs) {
-      const attr = attrs[attrName];
-      if (typeof attr === 'string') {
-        this.table.setAttribute(attrName, attr);
-      }
     }
   }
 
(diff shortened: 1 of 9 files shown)


addon/plugins/table/index.ts
```diff
@@ -29,14 +29,27 @@ export class TableView extends PluginTableView {
...
+      class: nodeClasses,
+      style,
+      ...attrs
+    } = node.attrs as Record<string, unknown>;
     if (typeof nodeClasses === 'string') {
       this.table.classList.add(...nodeClasses.split(' '));
     }
+    if (typeof style === 'string') {
+      this.table.style.cssText = `${this.table.style.cssText} ${style}`;
+    }
```
elpoelma, line 46: It seems we now both define the border-styles in inline css **and** external css for the dynamic table view. We can probably remove the external css on tables **or** not apply the border inline-css to the dynamic node-view.

addon/plugins/table/index.ts
```diff
@@ -29,14 +29,27 @@ export class TableView extends PluginTableView {
...
+    for (const attrName in attrs) {
+      const attr = attrs[attrName];
+      if (typeof attr === 'string') {
+        this.table.setAttribute(attrName, attr);
+      }
+    }
```
elpoelma, line 52: I don't know if we necessarily want this, as table `attrs` will not always be valid html attributes.

elpoelma (changes requested):
Looks good overall, table is now correctly exported with inline styles. I just have some questions on the `table` node-view.

elpoelma (approved):

abeforgit (approved):

PR #1143, opened 2024-02-16
Replace postpack script with `clean` as this breaks linking

piemonkey:
### Overview
`npm link` is broken by the files being removed (it uses a symlink). `yalc publish --push` works but `yalc push` seems to clear the declarations directory. Make it a manual step and add declarations to a dockerignore file.

N/A

### Setup
N/A

### How to test/reproduce
`npm link` or `yalc publish && yalc push` into another TS project and no longer see the types fail to import.

### Challenges/uncertainties
`npm link` still seems to see type problems as it is treating the types from different locations as different, e.g.:
```
Type 'import("<...>/ember-rdfa-editor-lblod-plugins/node_modules/prosemirror-model/dist/index").MarkSpec' is not assignable to type 'import("<...>/ember-rdfa-editor/node_modules/prosemirror-model/dist/index").MarkSpec
```

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+node_modules
+declarations
+.eslintcache
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "lint:js:fix": "eslint . --fix",
     "lint:types": "tsc --noEmit",
     "prepack": "tsc --project tsconfig.declarations.json",
-    "postpack": "rimraf declarations",
+    "clean": "rimraf declarations",
     "start": "concurrently -c \"auto\" -P \"npm:serve -- {@}\" \"npm:typecheck\" --",
     "typecheck": "tsc --noEmit --watch",
     "serve": "ember serve",


elpoelma (changes requested):
Removing the `postpack` script indeed prevents the `declarations` from being removed.
There can still occur issues with `yalc push` however in the following flow:
1. `yalc publish`: runs both the `prepack` and `postpack` scripts
2. change some types
3. `yalc push`: does not run the `prepack` and `postpack` scripts => types are not updated

Two possible solutions:
- Always run `yalc push` with the `--scripts` flag
- Add `preyalc` and `postyalc` scripts (which always run)

elpoelma:
I think the migration to pnpm (https://github.com/lblod/ember-rdfa-editor/pull/1145) makes this PR less necessary. If we would want this, we can always fork yalc in the future and fix the issues pertaining to npm script handling.

abeforgit:
closing as we did quite a deep dive and have a couple of new workflows around this problem that hopefully avoid the many pitfalls we found (the more you look the worse it gets)

PR #1144, opened 2024-02-19
GN-4723: Addition of `table-tooltip` component allowing users quick access to table-editing features

elpoelma:
### Overview
This PR includes several changes leading into the implementation of a `table-tooltip` component. This component allows users quick access to table-editing features.
Specifically, this PR includes the following changes/features:
- Addition of the `floating-ui` package and implementation of a custom (for now private) `floating-ui` modifier. Compared to the modifier provided by the `ember-velcro` package, it includes the following features:
   * Support for virtual elements (https://floating-ui.com/docs/virtual-elements)
   * Support for transform positioning (similar to the `vue` and `react` implementations of `floating-ui`: https://floating-ui.com/docs/usefloating#transform). By default the modifier uses `transform` positioning, but this can be disabled.
   * Support for fully customizable middleware-pipelines
- Addition of a (private) `selection-tooltip` component. This is a simple tooltip which displays under the current editor selection. It does so by leveraging the `floating-ui` modifier in combination with usage of virtual elements. A virtual element is used here, as the editor selection does not (always) correspond to a specific element, but rather to a position/series of positions. https://floating-ui.com/docs/virtual-elements contains more information on how to use virtual elements
- Addition of a (public) `table-tooltip` component. A simple tooltip which displays only when inside a table. It provides several table-actions including:
  * Add row after
  * Add row before
  * Add column after
  * Add column before
  * Delete row
  * Delete column
  * Delete table
  This tooltip is implemented to be as non-intrusive as possible: it only displays when clicking; and disappears when typing.

[GN-4723](https://binnenland.atlassian.net/browse/GN-4723?atlOrigin=eyJpIjoiMmVjNzhmNTc2ZDZhNGI2M2FkNGZmNWVkZDU2MjM1NDgiLCJwIjoiaiJ9)

### How to test/reproduce
- Open the dummy app on the `home` page
- Insert a table with some content inside
- Clicking inside any cell, the table-tooltip should appear
- The different actions provided by the table-tooltip should work
- When scrolling the viewport, the tooltip should correctly move along with the selection
- The tooltip should correctly disappear when the selection is out-of-view. (This is why the `contextElement` of the virtual-element is needed)
- When starting to type in the cell, the tooltip should disappear
- Observe that the tooltip element uses `transform` based positioning and the `position: fixed` approach.

### Challenges/uncertainties
- I did not opt into using the `ember-velcro` package for the implementation of this tooltip as it is missing several needed features. We might be able to upstream some features (such as virtual-element support and transform-based position) in the future.

### TODOs/Future work
- [x] Add `toggle header row` and `toggle header column` actions to tooltip
- [x] (Optional) Look into using a grid layout for the table-tooltip
- [ ] Look into supporting the `arrow` middleware (https://floating-ui.com/docs/arrow) (not crucial for this PR)
- [ ] Look into supporting a 'headless' version of the `floating-ui` modifier, in which the user itself has more control over how the styling is set/computed. (see the https://floating-ui.com/docs/vue and https://floating-ui.com/docs/react packages) (not crucial for this PR)
- [ ] Replace usages of `ember-velcro` with our custom `floating-ui` modifier (not crucial for this PR)
- [ ] Apply `size` modifier to dropdown components => allows for displaying a scrollbar when viewport is too short.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/great-experts-scream.md
+++ b/.changeset/great-experts-scream.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+Add custom `floating-ui` modifier with support for virtual elements, custom middleware options and transform-positioning
--- a/.changeset/quiet-llamas-shake.md
+++ b/.changeset/quiet-llamas-shake.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+Addition of a `table-tooltip` component which allows users to quickly access table-editing features
--- a/addon/components/_private/selection-tooltip.hbs
+++ b/addon/components/_private/selection-tooltip.hbs
@@ -1,0 +1,14 @@
+{{#if this.visible}}
+  <div
+    {{this.floatingUI
+      referenceElement=this.referenceElement
+      placement="bottom"
+      middleware=this.tooltipMiddleWare
+      strategy="fixed"
+      useTransform=true
+    }}
+    class="say-editor-selection-tooltip"
+  >
+    {{yield}}
+  </div>
+{{/if}}
--- a/addon/components/_private/selection-tooltip.hbs
+++ b/addon/components/_private/selection-tooltip.hbs
@@ -8,6 +8,7 @@
       useTransform=true
     }}
     class="say-editor-selection-tooltip"
+    ...attributes
   >
     {{yield}}
   </div>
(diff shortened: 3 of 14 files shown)


abeforgit:
small detail: the column resize handle renders above the popup, otherwise nice work!

abeforgit (changes requested):
small detail: the column resize handle renders above the popup, otherwise nice work!

elpoelma:
> small detail: the column resize handle renders above the popup, otherwise nice work!

Nice catch, thanks! This should be solved. It seems the `z-index` css rule of the `resize-handle` is not really necessary.

abeforgit (approved):

PR #1153, opened 2024-03-05
Correctly hide/disable sidebar when necessary

elpoelma:
### Overview
This PR includes an important fix with relation to the rendering of the sidebar.
When the `hideSidebar` argument is passed or no `aside` block is used, the sidebar is completely removed from the dom. This fixes an issue where the right side of the editor is unusable when the sidebar is 'hidden' (it was not hidden but transparent and was overlaid is an absolute way over the editor).

None

### How to test/reproduce
- Start the dummy app
- Remove the `aside block` or pass the `hideSidebar` argument
- Notice that the editor now has full-width
- Notice that the right side of the editor (where the sidebar normally renders) is now correctly usable and clickable

### Challenges/uncertainties
I think it a bit strange that the sidebar is rendered in an absolute manner (which caused this bug), we should probably change its styling in a seperate PR.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (approved):

elpoelma:
I'll update the e2e snapshots and merge.

PR #1154, opened 2024-03-06
Playwright updates

dkozickis:
### Overview

* Bump `@playwright/test` to `1.42.1` to allow for overriding styles in tests.
* Remove prosemirror devtools icon from screenshots - it's a detail that we don't care about, which styling can change without our knowledge e.g. after updating devtools.

### How to test/reproduce

Screenshots should pass

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations


e2e/models/screenshot.css
```diff
@@ -0,0 +1,3 @@
+[data-test-id="__prosemirror_devtools_collapsed_button__"] {
```
piemonkey, line 1: I seem to be still seeing the prosemirror dev tools (running with `npm run e2e:open`). If I inspect the snapshot then I see the element with this attribute but I don't see any sign of this CSS being applied.

piemonkey (commented):
Looks good but getting screenshots failing.

e2e/models/screenshot.css
piemonkey: Hmm, that makes debugging tricky. Are the tests in `npm run e2e:open` working for you?

piemonkey: Huh, had to refresh to get the video... That makes sense, as well as what you said about the snapshot failing as the system is different. So, only check snapshot tests on docker.

PR #1155, opened 2024-03-07
Improved RDFa backwards compatibility

elpoelma:
### Overview
This PR attempts to improve the backwards compatibility of rdfa-aware nodes.
Specifically, it converts each rdfa-aware nodespec to a function producing a node-spec based on an options argument, which is defined as follows:
```typescript
type Config = {
  rdfaAware?: boolean;
};
```
This, of course includes a breaking change as a lot of the nodespecs have been converted to functions. Specifically, the following node-specs have been converted to a function producing a node-spec:
- `block_rdfa`
- `inline_rdfa`
- `invisible_rdfa`
- `repaired_block`
- `list_item`, `bullet_list` and `ordered_list`
- `heading`

The following mark-specs have also been converted to a function:
- `inline_rdfa`

The following node-specs where already produced by a function (their options argument has been extended to include the `rdfaAware` option)
- the table node-specs
- `link`

The `doc` node-spec has been dropped and the `docWithConfig` function has been renamed to `doc`

The `rdfaAware` option always has a default value of false, the new features are thus opt-in.

Regarding the dummy pages:
- The 'index', 'dummy plugins' and 'space invisible' pages are not rdfa-aware
- The `editable nodes POC` page is rdfa-aware

### How to test/reproduce
- Start the dummy app

- Visit the `index`,  'dummy plugins' or 'space invisible' page
- Notice, (using e.g. the prosemirror devtools) that the different node-specs use the classic RDFa attributes
- Using the different sample configs, notice that the RDFa stays preserved (using the classic method)

- Visit the `editable nodes POC` page
- Notice that the different RDFa node-specs use the new API
- Using the different sample configs, notice that the RDFa stays preserved (using the new API)
- The 'editable' RDFa nodes should work as expected

### Challenges/uncertainties
This PR contains quite a few breaking changes, as it converts some node-specs to functions producing node-specs.
**Initially I tried the following approach:**
- Define an optional `rdfaAware` boolean property on node-specs, this property can be supplied to a node-spec using the spread operator: e.g. `block_rdfa: {...block_rdfa, rdfaAware: true }`
- Adapt the `toDOM`, `parseDOM` and `attrs` properties of a node-spec to be reactive to this `rdfaAware` property

This approach had the big following advantage:
- No (new) breaking changes: the `rdfaAware` property has a default value of false/undefined. The node-specs stays node-specs and do not become functions

But this approach did not work out:
- The `toDOM` and `parseDOM` node-spec properties get pulled out of their contexts, so it is not possible to get access to the `this` object inside them. I did not (yet) see an other option on how to access `this.rdfaAware`.

**In order to limit the breaking changes, we could also take the following approach:**
- Introduce seperate `configurable` functions that produce node-specs:
  * `configurableBlockRDFa`
  * `configurableInlineRDFa`
  * `configurableList`
  * ...
- Keep the current `block_rdfa`, `inline_rdfa` ... variables as normal node-specs with `rdfaAware: false`

This approach does have the disadvantage to introduce quite a few new variables/functions.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ export default class EditorComponent extends Component {
         block_rdfa: block_rdfa(),
       },
       marks: {
-        inline_rdfa,
+        inline_rdfa: inline_rdfa(),
         em,
         strikethrough,
         strong,
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ export default class EditorComponent extends Component {
           cellContent: 'block+',
           inlineBorderStyle: { width: '0.5px', color: '#CCD1D9' },
         }),
-        heading,
+        heading: heading(),
         blockquote,
         horizontal_rule,
         code_block,
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ export default class EditorComponent extends Component {
     // A prosemirror schema which determines how documents are parsed and written to the DOM.
     return new Schema({
       nodes: {
-        doc: docWithConfig({
+        doc: doc({
           defaultLanguage: 'nl-BE',
         }),
         paragraph,
@@ -102,7 +102,7 @@ export default class EditorComponent extends Component {
         code_block,
         text,
         hard_break,
-        block_rdfa,
+        block_rdfa: block_rdfa(),
       },
       marks: {
         inline_rdfa,
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -1,7 +1,7 @@
 import { v4 as uuidv4 } from 'uuid';
 import { Mark, type Attrs, type DOMOutputSpec } from 'prosemirror-model';
 import { PNode } from '@lblod/ember-rdfa-editor/index';
-import { unwrap } from '../utils/_private/option';
+import { isSome, unwrap } from '../utils/_private/option';
 import type {
   ContentTriple,
   IncomingLiteralNodeTriple,
@@ -11,6 +11,44 @@ import type {
 import { isElement } from '@lblod/ember-rdfa-editor/utils/_private/dom-helpers';
 
 // const logger = createLogger('core/schema');
+
+export const classicRdfaAttrSpec = {
+  vocab: { default: undefined },
+  typeof: { default: undefined },
+  prefix: { default: undefined },
+  property: { default: undefined },
+  rel: { default: undefined },
+  rev: { default: undefined },
+  href: { default: undefined },
+  about: { default: undefined },
+  resource: { default: undefined },
+  content: { default: undefined },
+  datatype: { default: undefined },
+  lang: { default: undefined },
+  xmlns: { default: undefined },
+  src: { default: undefined },
+  role: { default: undefined },
+  inlist: { default: undefined },
+  datetime: { default: undefined },
+};
+
+export function getClassicRdfaAttrs(
+  node: Element,
+): Record<string, string> | false {
+  const attrs: Record<string, string> = {};
+  let hasAnyRdfaAttributes = false;
+  for (const key of Object.keys(classicRdfaAttrSpec)) {
+    const value = node.attributes.getNamedItem(key)?.value;
+    if (isSome(value)) {
+      attrs[key] = value;
+      hasAnyRdfaAttributes = true;
+    }
+  }
+  if (hasAnyRdfaAttributes) {
+    return attrs;
+  }
+  return false;
+}
 
 export const rdfaAttrSpec = {
   properties: { default: [] },
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -178,20 +178,16 @@ export function renderInvisibleRdfa(
 }
 
 export function renderRdfaAttrs(
-  nodeOrMark: NodeOrMark,
+  rdfaAttrs: RdfaAttrs,
 ): Record<string, string | null> {
-  if (nodeOrMark.attrs['rdfaNodeType'] === 'resource') {
-    const contentTriple: ContentTriple | null = (
-      nodeOrMark.attrs['properties'] as OutgoingTriple[]
-    ).find(
+  if (rdfaAttrs.rdfaNodeType === 'resource') {
+    const contentTriple: ContentTriple | null = rdfaAttrs.properties.find(
       (prop) => prop.object.termType === 'ContentLiteral',
     ) as ContentTriple | null;
 
     return contentTriple
       ? {
-          about: (nodeOrMark.attrs['subject'] ||
-            nodeOrMark.attrs['about'] ||
-            nodeOrMark.attrs['resource']) as string,
+          about: rdfaAttrs.subject,
           property: contentTriple.predicate,
           datatype: contentTriple.object.language.length
             ? null
@@ -201,15 +197,11 @@ export function renderRdfaAttrs(
           resource: null,
         }
       : {
-          about: (nodeOrMark.attrs['subject'] ||
-            nodeOrMark.attrs['about'] ||
-            nodeOrMark.attrs['resource']) as string,
+          about: rdfaAttrs.subject,
           resource: null,
         };
   } else {
-    const backlinks = nodeOrMark.attrs[
-      'backlinks'
-    ] as IncomingLiteralNodeTriple[];
+    const backlinks = rdfaAttrs.backlinks as IncomingLiteralNodeTriple[];
     if (!backlinks.length) {
       return {};
     }
@@ -240,6 +232,7 @@ export type RdfaRenderArgs = {
   contentContainerTag?: string;
   contentContainerAttrs?: Record<string, unknown>;
 } & ({ content: DOMOutputSpec | 0 } | { contentArray: unknown[] });
+
 export function renderRdfaAware({
   renderable,
   tag,
@@ -251,15 +244,18 @@ export function renderRdfaAware({
   ...rest
 }: RdfaRenderArgs): DOMOutputSpec {
   const clone = { ...attrs };
+
+  //TODO: this should not be needed
   delete clone['properties'];
   delete clone['backlinks'];
   delete clone['subject'];
   delete clone['resource'];
   delete clone['__rdfaId'];
   delete clone['rdfaNodeType'];
+
   return [
     tag,
-    { ...clone, ...renderRdfaAttrs(renderable) },
+    { ...clone, ...renderRdfaAttrs(renderable.attrs as RdfaAttrs) },
     renderInvisibleRdfa(renderable, rdfaContainerTag, rdfaContainerAttrs),
     [
       contentContainerTag,
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -1,5 +1,5 @@
 import { v4 as uuidv4 } from 'uuid';
-import type { Attrs, DOMOutputSpec, Mark } from 'prosemirror-model';
+import { Mark, type Attrs, type DOMOutputSpec } from 'prosemirror-model';
 import { PNode } from '@lblod/ember-rdfa-editor/index';
 import { unwrap } from '../utils/_private/option';
 import type {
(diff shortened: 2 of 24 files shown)


README.md
```diff
@@ -96,16 +96,16 @@ export default class EditorComponent extends Component {
           cellContent: 'block+',
           inlineBorderStyle: { width: '0.5px', color: '#CCD1D9' },
         }),
-        heading,
+        heading: heading(),
```
piemonkey, line 99: Thinking about this, I wonder if the best approach is to, rather than change the nodespec to a function, to add a `WithConfig` function version, much in the same way as `docWithConfig`. We could then also export a constant version which is just the result of calling the function with no argument. e.g.

```
export const headingWithConfig: (config?: Config) => SayNodeSpec;
/** @deprecated Use headingWithConfig instead */
export const heading = headingWithConfig();
```

This would potentially avoid having to change any clients that don't want to use any of the new functionality. I guess unless we want it to be an explicit change that users know about?

addon/marks/inline-rdfa.ts
```diff
@@ -1,37 +1,62 @@
...
-      tag: 'span',
-      // default prio is 50, highest prio comes first, and this parserule should at least come after all other nodes
-      priority: 10,
-      getAttrs(node: string | HTMLElement) {
-        if (typeof node === 'string') {
+type Options = {
+  rdfaAware?: boolean;
+};
+
+export const inline_rdfa: (options?: Options) => SayMarkSpec = ({
```
piemonkey, line 15: It's not clear to me how the rdfaAware version of this mark should be used, as I thought the rdfa editing UI required `inline_rdfa` to be a node. So we would need some way to either automatically switch between mark and node using config or prompt the user to move it if they change the config.

README.md
piemonkey: Oops, just re-read your comment and realised that this is essentially your 'to limit the breaking changes' comment in your summary. I guess I see that as not too much overhead as it doesn't introduce any duplication, just an extra exported function for each RDFa aware nodespec compared to what existed before.

tests/dummy/app/controllers/editable-node.ts
```diff
@@ -83,23 +84,28 @@ export default class EditableBlockController extends Controller {
...
     nodes: {
-      doc: docWithConfig({
-        defaultLanguage: 'nl-BE',
-      }),
+      doc: {
+        ...doc({
+          defaultLanguage: 'nl-BE',
+          rdfaAware: true,
+        }),
+        ...sharedRdfaNodeSpec,
```
piemonkey, line 92: This doesn't feel like the right way to do this. It requires anyone who wants to start using the `rdfaAware` features to know that they also need to do this for some of the nodes. To me, having a `config` option implies that switching that option should take care of turning on the feature rather than being part of the picture. I think spreading (or not) the shared nodespec should be done in the specs themselves.

addon/utils/_private/constants.ts
```diff
@@ -3,6 +3,8 @@
  * we've added a, del, ins to the list since we assume they only contain phrasing content in the editor
  * we've removed br from the list to be inline with editor behaviour, which treats it as a block
  **/
+
+export const HEADING_ELEMENTS = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] as const;
```
piemonkey, line 7: This takes the JSdoc from `PHRASING_CONTENT`, we should position it differently.

addon/plugins/table/nodes/table.ts
```diff
@@ -146,45 +174,77 @@ export function tableNodes(options: TableNodeOptions): TableNodes {
...
-          },
-          ['tbody', 0],
-        ];
+        if (rdfaAware) {
+          return [
+            'table',
+            {
+              ...renderRdfaAttrs(node.attrs as RdfaAttrs),
+              class: 'say-table',
+              style: `width: 100%; ${tableStyle || ''}`,
```
piemonkey, line 193: I can't remember the reason why the `width: 100%;` is in the `serialize` but not the `toDOM`, but I do remember there being one... Is there a reason to change it now?

addon/core/schema.ts
```diff
@@ -12,6 +12,44 @@ import { isElement } from '@lblod/ember-rdfa-editor/utils/_private/dom-helpers';
...
+  datatype: { default: undefined },
+  lang: { default: undefined },
+  xmlns: { default: undefined },
+  src: { default: undefined },
+  role: { default: undefined },
+  inlist: { default: undefined },
+  datetime: { default: undefined },
+};
+
+export function getClassicRdfaAttrs(
```
piemonkey, line 35: It strikes me that each usage of this seems to be in an `if (rdfaAware) { getRdfaAttrs(node) } else { getClassicRdfaAttrs(node) }` type construction. Would it be cleaner to have `getRdfaAttrs` take `rdfaAware` and have the `if` in just one place?

piemonkey (commented):
Initial playing around looks good. Here are some code-based comments, I'll keep testing for any use-based comments that might come up.

README.md
elpoelma: Yeah that's true, it would indeed not introduce too much overhead, although I would not deprecate the older 'static' version.

tests/dummy/app/controllers/editable-node.ts
elpoelma: That's fair, I'm wondering if we should automatically enable 'editable'/'isolating'/'selectable' when `rdfaAware` is enabled, or if we should add additional 'opt-in' options for these to the `Config` object.

addon/marks/inline-rdfa.ts
elpoelma: We can also leave the mark as is (just a non-configurable node-spec), and only make the `inline_rdfa` node-spec configurable.

tests/dummy/app/controllers/editable-node.ts
elpoelma: As (at the moment at least), we do not always want the new rdfa-aware nodes to be editable/isolating/selectable.

addon/utils/_private/constants.ts
elpoelma: Done

addon/core/schema.ts
elpoelma: I updated both the `getRdfaAttrs` and `rdfaAttrSpec` functions to take in an `rdfaAware` parameter, so this should reduce the code duplication by quite a bit.

addon/plugins/table/nodes/table.ts
elpoelma: the `toDOM` function is only used when not using the custom tableview right?

README.md
piemonkey: I like deprecating as I'm imagining the path of someone migrating to `rdfaAware: true`.
- If there is no deprecation, they have no idea which static NodeSpecs need to be switched to a function, they need to go through the code or hope for a list in the docs.
- If we deprecate the static version, either
  - they're migrating straight to `rdfaAware: true` and they can replace the deprecated statics,
  - or they are migrating from `rdfaAware: false` and they can look at what config options are available for each function NodeSpec.

addon/marks/inline-rdfa.ts
piemonkey: True, though it makes the migration more manual... Maybe this is one of those times where we can't cleanly make everything automatic though.

tests/dummy/app/controllers/editable-node.ts
piemonkey: Is it something that the (dev) users are going to want to alter, or is it always going to be the same for a NodeSpec?

To me it seems most likely that while some NodeSpecs want different settings, they are like that for a reason, so we don't really want (dev) users playing with them as it's cause weird problems.

If we do want to let users play with them, then I think setting sensible defaults when they pass `rdfaAware` but then allowing them to be set explicitly with other config options, is the way to go.

addon/plugins/table/nodes/table.ts
piemonkey: I think so. I think the reason I didn't change this when adding `tableStyle` was that it hadn't already been set, so it's more a @dkozickis question on whether there is an edge case where this is used.

README.md
elpoelma: Yeah, that's fair. I'm just not sure about the naming of the configurable node-specs.

elpoelma: yes, we're indeed talking about deprecating non-function node-specs, I'm just wondering if we can provide a better API than consumers of this package having to use syntax like `block_rdfa: block_rdfa()` the whole time.

elpoelma: I restored the classic node-specs, and create `WithConfig` functions where relevant. I'm now thinking of a better solution on how to configure node-specs.

abeforgit: I'm not too worried about convenience in the config process, as it's already pretty terrible (on purpose)
clarity and explicitness over convenience for now, and then we add a convenience layer later on

abeforgit:
It's pretty magical switching between the two "styles" (rdfa-aware vs not) and seeing the document stay stable, great stuff
(I do realize we probably can't guarantee stability, but having it this close is already fantastic)

README.md
elpoelma: I deprecated all static node-specs that now have a configurable option

tests/dummy/app/controllers/editable-node.ts
elpoelma: I guess we can indeed turn editable/selectable/isolating on by default (when `rdfaAware` is enabled and where applicable), dev users can indeed override them.
It would turn it on for:
- block_rdfa
- inline_rdfa
- heading?
- doc

addon/marks/inline-rdfa.ts
elpoelma: I deprecated the mark version with a message to use the node version instead.

README.md
```diff
@@ -96,13 +96,13 @@ export default class EditorComponent extends Component {
...
         blockquote,
         horizontal_rule,
         code_block,
         text,
         hard_break,
-        block_rdfa,
+        block_rdfa: block_rdfa(),
```
piemonkey, line 105: Typo: This should be `blockRdfaWithConfig()`.

tests/dummy/app/controllers/index.ts
```diff
@@ -65,6 +64,7 @@ import { chromeHacksPlugin } from '@lblod/ember-rdfa-editor/plugins/chrome-hacks
 import type { PluginConfig } from '@lblod/ember-rdfa-editor';
 import { emberApplication } from '@lblod/ember-rdfa-editor/plugins/ember-application';
 import { getOwner } from '@ember/application';
+import { heading } from '@lblod/ember-rdfa-editor/plugins/heading/nodes/heading';
```
piemonkey, line 67: It feels strange to be using deprecated APIs in our own test app, should we use the `WithConfig` versions?

piemonkey (approved):
It's good to think we can move away from this long-running branch workflow soon... Just a couple of tiny tweaks.

README.md
elpoelma: Fixed

tests/dummy/app/controllers/index.ts
elpoelma: Yeah that's fair, will update.

elpoelma:
I made the following additional compatibility improvements:
- `getRdfaAttrs` and `rdfaAttrSpec` can now be called without providing a config object. If no config object is passed, `rdfaAware` is set to `false`
- The now deprecated `rdfaAttrs` is set equal to `rdfaAttrSpec({ rdfaAware: false })`
- The `extraAttributes` config option has been re-added to the `docWithConfig` function.

piemonkey (approved):
Oops, missed this. Yes, good to merge from my side.

elpoelma:
In [`4891d3b` (#1155)](https://github.com/lblod/ember-rdfa-editor/pull/1155/commits/4891d3bf904be34e066375d3479a7593ee81c90e) I have included some improvements to the types of the `say-serializer` to they are more backwards-compatible.

elpoelma:
I readded the `keyMapOptions` argument in [`3fc50ef` (#1155)](https://github.com/lblod/ember-rdfa-editor/pull/1155/commits/3fc50ef4a6d83ef98cf3b26220d5ff9eb5185e18) and marked it as deprecated. When it is enabled, a warning should also be displayed in dev builds.

piemonkey:
We should disable that 'parser bug' blackbox test for now, so we don't miss things on failing tests...

abeforgit (approved):

elpoelma:
> We should disable that 'parser bug' blackbox test for now, so we don't miss things on failing tests...

Yep, will do that in the other PR.

PR #1156, opened 2024-03-07
Hierarchical styles for ordered lists

dkozickis:
### Overview

Hierarchical styles for ordered lists

Allows to enable "hierarchical" style for ordered list, which will render

```
1. First level
  1.1 Second level
    1.1.1 Second level
```

Kaleidos request, no tix :(

### Setup

1. Checkout
2. `npm run start` 

### How to test/reproduce

Repeat after me

https://github.com/lblod/ember-rdfa-editor/assets/769698/68b78801-e7cb-4c16-b82f-749328a47a68

https://github.com/lblod/ember-rdfa-editor/assets/769698/bcc12bfb-5779-4e6a-8084-00537a2f1936

### Challenges/uncertainties

Trying to switch to "normal" style (e.g. roman) in the middle of hierarchical list will have no effect, which might not be 100% obvious.

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations


abeforgit (changes requested):
- cant change non-hierarchical list into hierarchical list
- cant do the reverse either
- inconsistent spacing with alternating list:
image
the red marks a place where it's obviously wrong, the orange could be due to the `1` character being weird or something, can't really tell
- for alternating: to be consistent with our other available styles, I think going for lowercase letters looks better

#### Not a regression
the following is confusing behavior that is not related to your pr, but might come up when reviewing:


[recording.webm](https://github.com/lblod/ember-rdfa-editor/assets/16419862/8db54cc3-4185-4abc-8abc-f0d88da0406a)

abeforgit:
@SchrodingersCat00 @brenner-company requested your reviews here to see if it works for you guys

brenner-company:
@abeforgit I'll try and give this a look on monday 🤞

brenner-company:
@dkozickis Thanks for you effort regarding this PR!

Just to be sure, this is the way we implemented our temporary (forked) version of those hierarchical styles (mostly the same as with this PR, but with some differences):
- The ability to change non-hierarchical list into hierarchical list (and the reverse) as mentioned by @abeforgit before. This enables the user to something like in the screenshot below this list. A hierarchical list can both be 'wrapped' (not shown in the screenshot below) and continued by a normal style list.
- Within our own version, we limit the usage of the hierarchical list to one instance to avoid nested list of that same hierarchical style list. We do this by disabling the hierarchical list button when needed (see screenshot below).
- We've implemented the limitation above also because we've encountered a bug when creating a new hierarchical list that is not starting on the root level if the document and is 'wrapped' by a normal style list (see last screenshot below). In this example, the second instance of a hierarchical list continues the numbering of the last second level item 🤷‍♂️. It seems this same issue is present within this version of the hierarchical list.

CleanShot 2024-03-13 at 18 42 46

CleanShot 2024-03-13 at 18 52 36

CleanShot 2024-03-13 at 18 57 13

I hope this helps a bit, but is a lot to cover of course. If you have any follow-up questions, feel free to ask!

abeforgit:
to be clear: limiting to one list on the page is _not_ an acceptable solution for this feature
what is acceptable is only allowing the whole list to be hierarchical, and not allowing different styles for the different levels _if_ the list is hierarchical

abeforgit:
> I'm not entirely sure it makes sense to be able to switch to normal style in the middle of the hierarchical 🤔 It's not indicated correctly though, as buttons are enabled at all times, that's bad UX.
having the buttons enabled or disabled depending on the list level your cursor is in also doesn't make sense imo. So I'd propose that choosing a non-hierarchical style at any point in the list converts the entire list

brenner-company:
> to be clear: limiting to one list on the page is _not_ an acceptable solution for this feature

Correct, this was only implemented in our version because we were encountering the numbering issue I mentioned within https://github.com/lblod/ember-rdfa-editor/pull/1156#issuecomment-1995208655. In the meantime, I've heard from @abeforgit, he found a fix for this particular issue 👍

brenner-company:
> EDIT: it does seem that kaleidos needs to be able to make a list like it seems that kaleidos also needs to be able to make a list like
> 
> ```
> 1
>   1.1
>     a
>       1.1.1.1
> ```
> 
> so it does at least make sense for their users

Just a quick remark that is not the case, but I'll try to expand on that this afternoon 🤞

abeforgit (approved):
Working very nicely now imo

I'm heavily nitpicking now, but in a situation like here:

image
It _looks_ like these are 2 different styles on the same level, but that's not the case, the hierarchical is actually a child of the "a" item there. It's completely correct and also behaves correctly, but just because we're quite deep the hierarchical list marker becomes so long that it starts aligning with the parent marker.

Honestly not sure if that's solvable or even really a problem, but just mentioning it in case you have ideas. Otherwise lgtm

brenner-company:
Looks good with the new changes 👍

I'll just need to double check this with Johan to see if everything is covered/correct.

abeforgit:
was asked to make a tracking ticket, here it is: https://binnenland.atlassian.net/browse/GN-4754

brenner-company:
Johan is giving a thumbs-up for this version, but when demoing I encountered the following issue:

Starting with a regular (type) list that contains a hierarchical list, when you add or remove items to/from that regular list, the numbering of the items within the hierarchical list don't adjust correctly. But when you reload the page (and thus the editor), it's shown the correct way.

https://github.com/lblod/ember-rdfa-editor/assets/18174827/3d5f5bea-c97c-4e73-8440-01edb21fb419

brenner-company:
> Working very nicely now imo
> 
> I'm heavily nitpicking now, but in a situation like here:

>  It _looks_ like these are 2 different styles on the same level, but that's not the case, the hierarchical is actually a child of the "a" item there. It's completely correct and also behaves correctly, but just because we're quite deep the hierarchical list marker becomes so long that it starts aligning with the parent marker.
> 
> Honestly not sure if that's solvable or even really a problem, but just mentioning it in case you have ideas. Otherwise lgtm

I think that would indeed fall within the category of 'if solvable or even really a problem'.

abeforgit:
> Johan is giving a thumbs-up for this version, but when demoing I encountered the following issue:
> 
> Starting with a regular (type) list that contains a hierarchical list, when you add or remove items to/from that regular list, the numbering of the items within the hierarchical list don't adjust correctly. But when you reload the page (and thus the editor), it's shown the correct way.
> CleanShot.2024-03-15.at.18.39.14.mp4



good catch! we'll tackle it next week

abeforgit (approved):
can't find any faults anymore, nice job
@brenner-company last call?

abeforgit:
it seems we still have problems on chrome only

brenner-company:
Did a check a few moments ago and, as @abeforgit mentioned, the issue seems to persist within Chrome (Firefox is fine though).

brenner-company:
@abeforgit Ok, I'll double check this with Johan again 🤞

brenner-company:
Just to be sure, does switching to hierarchical list apply to both parent and descendant lists of a non-hierarchical list type? For example if you want a bullet list within a hierarchical list, that would not be possible?

abeforgit:
> I've just had a quick call with Johan. Starting with a non-hierarchical list and going into a hierarchical list on the next level is not a requirement for us but having the functionality to (E.g. my previous example) create a bullet (or other non-hierarchical type) list within a hierarchical list (and ending with that) would be a very big plus for the user. Would that be a possibility? 🤞
> 
> That way it would be possible to neglect the Chrome bug of course, but would include disabling the hierarchical list button in that case (I think?).

Christ what kinds of lists are they coming up with over there :scream: 
@dkozickis you think this is feasible? I can have a look if you won't have time

brenner-company:
> Christ what kinds of lists are they coming up with over there 😱
You wouldn't believe me if I told you 😅

brenner-company:
@dkozickis I'll give this a look when the demo/retro/plan-meeting is finished over here 👍

abeforgit (changes requested):
unfortunately, quite a few issues
- can't turn a bullet list into a hierarchical list even though button is active
- can't turn a sublist of a hierarchical list into a bullet list (explicitly asked by @brenner-company )
- after turning a sublist of a hierarchical list into a flat list, you can't change it back to make the full list hierarchical in any way

see recording for more detail (if the video doesn't work in firefox, try chrome)

https://github.com/lblod/ember-rdfa-editor/assets/16419862/96113691-4462-4b22-818b-7c7672f8f527

brenner-company:
> unfortunately, quite a few issues
> 
> * can't turn a bullet list into a hierarchical list even though button is active
> * can't turn a sublist of a hierarchical list into a bullet list (explicitly asked by @brenner-company )
> * after turning a sublist of a hierarchical list into a flat list, you can't change it back to make the full list hierarchical in any way
> 
> see recording for more detail (if the video doesn't work in firefox, try chrome)

I've just tested this locally and I can confirm the issues @abeforgit mentions.

abeforgit:
superceded by #1184

PR #1160, opened 2024-03-12
Add support for alternating row styles

elpoelma:
### Overview
This PR adds support for configuring alternating row styles of tables. 
Developers making use of the table plugin can configure the background for the odd and even rows of a table.

It is implemented as follows:
- If background colors for odd or even rows are provided, the following variables are set in the inline styles of the table element:
   * `--say-even-row-background`
   * `--say-odd-row-background`
- The following css has been added to `_c-table.scss`:
```
tr:nth-child(even){
  	background: var(--say-even-row-background);
}
tr:nth-child(odd){
  	background: var(--say-odd-row-background);
}
```
- The `serialize` method of the `table_row` nodespec adds inline styles based on if the row is even or odd. This is only possible in `serialize`, not in `toDOM`, as only `serialize` has access to the state.

Additionally, this PR includes the following changes:
- Addition of a `constructInlineStyles` utility function, which takes a css object and converts it to an inline styles string
- Addition of `getParent` and `getPos` utilities. These utilities allow you to resolve a prosemirror-node in a document. (needed to check if a row is even/odd in a table)

Solution to: https://binnenland.atlassian.net/browse/GN-4705?atlOrigin=eyJpIjoiMzRlNzA1NzcwNDdkNGMwOGJmMzlhYzg0Y2YxYWIzNDkiLCJwIjoiaiJ9

### How to test/reproduce
- `npm start`
- Open the index dummy page
- Insert a table with a few rows
- Notice that the rows alternate
- Export to html and observe the inline styles responsible for the alternating row backgrounds.

### Challenges/uncertainties
I was not able to enable the `nth-child` css rules in the table inline css, as inline css does not allow for setting children styles. The approach with css variables seemed the best second option.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/plugins/table/nodes/table.ts
+++ b/addon/plugins/table/nodes/table.ts
@@ -218,7 +218,7 @@ export function tableNodes(options: TableNodeOptions): TableNodes {
       serialize(node, state) {
         const pos = getPos(node, state.doc) as ResolvedPos;
         // table rows are 1-indexed
-        const isEven = (pos.index() + 1) % 2 === 0;
+        const isEven = pos.index() % 2 === 1;
         const style = {
           ...rowStyle,
           ...(isEven ? evenRowStyle : oddRowStyle),
--- a/addon/plugins/table/nodes/table.ts
+++ b/addon/plugins/table/nodes/table.ts
@@ -1,9 +1,11 @@
 // Helper for creating a schema that supports tables.
 
-import { Node as PNode, NodeSpec } from 'prosemirror-model';
+import { Node as PNode, NodeSpec, ResolvedPos } from 'prosemirror-model';
 import { getRdfaAttrs, rdfaAttrs } from '@lblod/ember-rdfa-editor/core/schema';
 import { TableView } from '@lblod/ember-rdfa-editor/plugins/table';
 import SayNodeSpec from '@lblod/ember-rdfa-editor/core/say-node-spec';
+import { getPos } from '@lblod/ember-rdfa-editor/utils/node-utils';
+import { constructInlineStyles } from '@lblod/ember-rdfa-editor/utils/_private/html-utils';
 
 interface ExtraAttribute {
   default: unknown;
@@ -97,13 +99,17 @@ interface TableNodeOptions {
     style?: string;
     color: string;
   };
+  rowBackground?: {
+    even?: string;
+    odd?: string;
+  };
 }
 
 interface TableNodes extends Record<string, NodeSpec> {
   table: SayNodeSpec;
-  table_row: NodeSpec;
-  table_cell: NodeSpec;
-  table_header: NodeSpec;
+  table_row: SayNodeSpec;
+  table_cell: SayNodeSpec;
+  table_header: SayNodeSpec;
 }
 
 export function tableNodes(options: TableNodeOptions): TableNodes {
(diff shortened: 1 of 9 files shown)


addon/plugins/table/nodes/table.ts
```diff
@@ -194,8 +215,26 @@ export function tableNodes(options: TableNodeOptions): TableNodes {
           },
         },
       ],
+      serialize(node, state) {
+        const pos = getPos(node, state.doc) as ResolvedPos;
+        // table rows are 1-indexed
+        const isEven = (pos.index() + 1) % 2 === 0;
+        const style = {
+          ...rowStyle,
+          ...(isEven ? evenRowStyle : oddRowStyle),
```
piemonkey, line 224: When serializing this applies the correct even or odd CSS var, but we can't necessarily rely on the CSS being applied. I think for this to work as expected we'd need to apply `background: isEven ? options.rowBackground.even : options.rowBackground.odd`.

addon/plugins/table/nodes/table.ts
```diff
@@ -194,8 +215,26 @@ export function tableNodes(options: TableNodeOptions): TableNodes {
           },
         },
       ],
+      serialize(node, state) {
+        const pos = getPos(node, state.doc) as ResolvedPos;
+        // table rows are 1-indexed
+        const isEven = (pos.index() + 1) % 2 === 0;
```
piemonkey, line 221: Could be simplified to `pos.index() % 2 === 1`

piemonkey (commented):
It feels like a long way around, but I also can't think of a shorter way within the constraints. I'd also like to be able to use something like CSSStyleDeclaration to construct the styles, but since we don't have one and can't construct one, we can't.

addon/plugins/table/nodes/table.ts
elpoelma: very true haha :)

elpoelma:
> It feels like a long way around, but I also can't think of a shorter way within the constraints. I'd also like to be able to use something like CSSStyleDeclaration to construct the styles, but since we don't have one and can't construct one, we can't.

Yes indeed, I also looked into another way to construct the CSS styles, but could not find any.

addon/plugins/table/nodes/table.ts
piemonkey: I was thinking of e.g. when copy-pasting, we can't rely on where we're pasting to understanding the CSS variables. Actually, I just tried this in Google Docs, and while it doesn't apply the color at all with the variables approach, and if I inline the colors themselves, it does, the background is just applied to the text as a highlight, not the whole cell. Maybe this looks more broken? Then again, I think this is a feature that Google Docs doesn't have, so perhaps it's expected and might work well elsewhere?

piemonkey: Sorry, that was a bit rambling...

I think the inlined version is more portable compared to the var approach when serialising. I have a slight preference for that approach, even if it might lead to odd results in some places.

elpoelma: It makes sense that google docs cannot handle the variables approach, I'll also try with word inline. You're right that it's probably better to refrain from using the variables in the serialization.

elpoelma: Should be changed now

abeforgit (approved):
looking good! inline serialization looks clean as well

abeforgit (changes requested):
I just noticed that the header-row/header-column interact strangely with this feature
I say strangely cause I cant quite figure out if it's wrong or not, but maybe have a look?

elpoelma:
> I just noticed that the header-row/header-column interact strangely with this feature I say strangely cause I cant quite figure out if it's wrong or not, but maybe have a look?

That should be fixed now, it was because `th` elements had their `background-color` set to white in the css.

abeforgit (approved):

PR #1162, opened 2024-03-14
GN-4706: Table cell background

### LINKED ISSUE #1161: Background color cell
<p>Allow the user to set the bg color of a cell. SHould be inline styles, working kind of like the text marking feature we have but specifically for cells. Use a similar color selection menu for consistency. Add to both the toolbar table menu and the floating menu.</p>

dkozickis:

Closes #1161 

### Setup

1. Checkout

### How to test/reproduce

Repeat after me:

https://github.com/lblod/ember-rdfa-editor/assets/769698/50bd46eb-4b5c-41b4-8219-2f059a93e457

### Challenges/uncertainties

Design 😅 

### Checks PR readiness

Uploading CleanShot 2024-03-14 at 12.21.26.mp4…

- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/addon/components/plugins/table/color.hbs
+++ b/addon/components/plugins/table/color.hbs
@@ -1,0 +1,39 @@
+{{#if @controller}}
+  <div class="say-toolbar__button say-toolbar__highlight-button">
+    <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} as |velcro|>
+      <button
+        type='button'
+        class='main {{if @active "is-active"}}'
+        disabled={{not this.canSetColor}}
+        title={{t 'ember-rdfa-editor.color.button-label'}}
+        {{velcro.hook}}
+        {{on 'click' this.toggleDropdown}}
+        {{this.setupDropdownButton}}
+      >
+        <AuIcon
+          @icon='pencil'
+          @ariaHidden={{true}}
+          @size='large'
+          style={{this.htmlSafe (concat "border-bottom-color: " this.currentColor)}}
+        />
+      </button>
+
+      {{#if this.dropdownOpen}}
+        <Utils::ColorSelector
+          class='say-dropdown__menu is-visible'
+          {{velcro.loop}}
+          {{focus-trap
+            shouldSelfFocus=true
+            focusTrapOptions=(hash
+              clickOutsideDeactivates=this.clickOutsideDropdown
+            )
+          }}
+          tabindex="-1"
+          @onChange={{this.setColor}}
+          @color={{this.currentColor}}
+          @presetColors={{@presetColors}}
+        />
+      {{/if}}
+    </this.Velcro>
+  </div>
+{{/if}}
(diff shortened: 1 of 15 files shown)


piemonkey (approved):
Looks good. I feel like having both the toolbar button and the color choice in the (very aggressive in my view) tooltip menu seems like too much, but maybe that's just me.

addon/components/plugins/table/color.hbs
```diff
@@ -0,0 +1,39 @@
+{{#if @controller}}
+  <div class="say-toolbar__button say-toolbar__highlight-button">
+    <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} as |velcro|>
+      <button
+        type='button'
+        class='main {{if @active "is-active"}}'
+        disabled={{not this.canSetColor}}
+        title={{t 'ember-rdfa-editor.color.button-label'}}
```
elpoelma, line 8: We should use another label for this.

addon/components/plugins/table/color.ts
elpoelma, line 1: I'm wondering if we should include a seperate dropdown button, like we use for the highlight toolbar button.
This way the user can use between quickly applying a color and selecting a different color.
Not sure what the best approach would be here.

addon/plugins/table/nodes/table.ts
```diff
@@ -106,24 +110,78 @@ interface TableNodes extends Record<string, NodeSpec> {
...
+  borderLeft: {
+    default: inlineBorderStyle ?? null,
+    getFromDOM(dom) {
+      return dom.style.borderLeft || null;
+    },
+    setDOMAttr(value, attrs) {
+      if (typeof value === 'string') {
+        appendToStyleAttribute(attrs, `border-left: ${value}`);
+      }
+    },
```
elpoelma, line 155: Nice approach to include seperate `fromDOM` and `toDOM` methods for each of the attributes. I was already thinking of including that option in the default `SayNodeSpec` API.

elpoelma (changes requested):
Works well overall, just some remarks.

- We should (as you mentioned) review the design.
- We should probably also use a different icon to avoid confusion.
- I'm wondering if it is possible to include the color picker in the floating tooltip?

abeforgit:
> Looks good. I feel like having both the toolbar button and the color choice in the (very aggressive in my view) tooltip menu seems like too much, but maybe that's just me.

having both is indeed much, I think in apps we'd have either/or though
for the tooltip menu I agree its a bit in your face atm. You have ideas on how to improve it? maybe moving it a bit further away from the cursor would already help?

abeforgit (commented):
Don't really know how to solve this, but if you select multiple cells, the selection color overrides the background color, so it kinda seems like the cell color feature isn't doing anything (until you deselect)

If you know of a quick solution, it would be nice (maybe something with the alpha channel on css `rgba()` colors? or some opacity setting?)

abeforgit (approved):

PR #1165, opened 2024-03-15
Ensure `rdfa-parsing` unit tests run correctly

elpoelma:
### Overview
This PR includes the following changes:
- Addition of a `deepArrayContains` extension to qunit-assert
- Modifications to the `rdfa-parsing` unit tests to ensure these run

None

### How to test/reproduce
- Observe that all tests succeed (especially the `rdfa-parsing` tests)

### Challenges/uncertainties
While (if strictly following their interfaces) the objects/subjects of node properties/backlinks should follow the `RDF.Term` interfaces, they currently do not. It is currently not possible to deep compare NamedNodes constructed by `sayDataFactory.namedNode` and objects present in properties/backlinks.
The objects/subjects of properties/backlinks lack the `equals` method. This is due to `JSON.parse` and `JSON.serialize`.

Possible solution to this problem:
- When calling `JSON.parse` when parsing incoming-triples/outgoing-triples, we could configure it to recreate `RDF.Term` objects (literals, namednodes etc.)
- Find another way to pass the incoming-triples/outgoing-triples than using JSON serialization.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.woodpecker/.e2e.yml
+++ b/.woodpecker/.e2e.yml
@@ -1,7 +1,7 @@
 steps:
   e2e:
     # this image must be the same as in `scripts/e2e.sh`
-    image: mcr.microsoft.com/playwright:v1.42.1-jammy
+    image: mcr.microsoft.com/playwright:v1.41.2-jammy
     commands:
       - corepack enable
       - pnpm i --frozen-lockfile
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -18,5 +18,5 @@ docker run -it --rm \
 -e DISPLAY=$DISPLAY_ENV \
 -w /e2e \
 --platform linux/amd64 \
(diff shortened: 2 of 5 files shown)


abeforgit:
cfr equals: was kinda the reason to remake those interfaces so that we could drop that requirement
could we just make a separate simple equals function?

abeforgit (approved):

elpoelma:
We can also adapt our `factory` to no longer extend from `DataFactory` and make objects without `equals` function. This way they can be easily parsed/serialized.

abeforgit:
ok that doesnt look too bad, go for it

elpoelma:
> ok that doesnt look too bad, go for it

I'm going to put the JSON parsing/serialization reviving in a different PR, mainly for clarity.

PR #1168, opened 2024-03-18
Revive say-datafactory terms when parsing outgoing and incoming props

elpoelma:
### Overview
This PR ensures that the different `SayTerm` objects present in outgoing/incoming properties are correctly revived.
To implement this, the following features/fixes are added:
- A `fromJson` method on the `SayDataFactory` class. This converts JSON objects to full-fledged `SayTerm` objects.
- Updates to the `LiteralNodeTerm`, `ResourceNodeTerm` and `ContentLiteralTerm` interfaces. The `equals` method is no longer included.

When parsing the json-serializations of the outgoing/incoming properties, a `revive` function is added to ensure that the `object` and `subject` fields are correctly revived.

This PR also includes adaptations to the rdfa-parsing tests to ensure they remain running correctly.

None

### How to test/reproduce
This PR does not include any new user-visible features. All the RDFa features should remain the same as before.

### Challenges/uncertainties
**A note on object spreading**
The `Term` objects produce by the rdf-data-factory library (https://www.npmjs.com/package/rdf-data-factory) are structured as follows:
https://github.com/rubensworks/rdf-data-factory.js/blob/96fa3185b141d4abaf0b27ec002b6d5c58276483/lib/NamedNode.ts
As you can see in the above code snippet, they each `value` and `termType`  properties, as well as an `equals` method on their prototype. When spreading an object of this class, the `value` and `termType` properties are correctly preserved, while the `equals` method is not (as it is part of the prototype). This would not be a problem if typescript handled this correctly.

An example of the issue:
```
interface SampleInterface {
  property1: string;
  property2: string;
  aMethod(): string;
}

class SampleClass implements SampleInterface {
  constructor(readonly property1: string, readonly property2: string) {}

  aMethod() {
    return this.property1 + this.property2;
  }
}


const a: SampleInterface = new SampleClass('test1', 'test2');

const b = {...a};

b.aMethod(); //Runtime Error: b.aMethod is not a function
```
The above snippet will not cause typescript to complain, but will result in a runtime error. Based on the `SampleInterface` type, typescript cannot determine if `aMethod` is defined on the prototype or object itself. It always assumes that all properties (including methods/functions) are defined on the object itself.

Note: the following snippet _will_ result in a compile-time error:
```
interface SampleInterface {
  property1: string;
  property2: string;
  aMethod(): string;
}

class SampleClass implements SampleInterface {
  constructor(readonly property1: string, readonly property2: string) {}

  aMethod() {
    return this.property1 + this.property2;
  }
}


const a = new SampleClass('test1', 'test2');

const b = {...a};

b.aMethod(); //Compile-time error: Property 'aMethod' does not exist on type '{ property1: string; property2: string; }
```
In the above snippet, typescript can correctly conclude that `aMethod` is part of the prototype, and not instances of this class.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations


addon/core/say-data-factory.ts
```diff
@@ -53,8 +56,13 @@ export class SayDataFactory extends DataFactory {
   contentLiteral(
     languageOrDataType?: string | SayNamedNode<string>,
   ): ContentLiteralTerm {
-    const literal = this.literal('', languageOrDataType);
-    return { ...literal, value: '', termType: 'ContentLiteral' };
+    // TODO: we need to be careful with using spread operators here:
```
piemonkey, line 59: I wonder if it's better to modify the return to include an `equals` method (ideally on the prototype, so it's as similar to the other Terms as possible). The classes for [Literals](https://github.com/rubensworks/rdf-data-factory.js/blob/master/lib/Literal.ts) and [NamedNodes](https://github.com/rubensworks/rdf-data-factory.js/blob/master/lib/NamedNode.ts) seem almost trivial, so we could implement our own versions which compare the correct `termType`s easily.

piemonkey (commented):
Hmm, interesting problem. I feel like we're going to keep having future problems with this if we don't fix this in the factory at least.

addon/core/say-data-factory.ts
elpoelma: Yes that could be better indeed. We could define our custom terms as classes (not as interfaces). We could even define the `equals` method as an arrow function, so it is preserved when using the spread operator. If we use classes, we are sure we don't accidentally make spreading 'mistakes'.

We could base our implementation on https://www.npmjs.com/package/rdf-data-factory

elpoelma: I created a PR to introduce custom term classes: https://github.com/lblod/ember-rdfa-editor/pull/1171. Not entirely sure if it is really necessary, but it should provide a solution to the use of `equals` methods an object spreading.

addon/core/say-data-factory.ts
```diff
@@ -72,18 +80,58 @@ export class SayDataFactory extends DataFactory {
...
   }
+
+  fromJson(serialized: Omit<SayTerm, 'equals'>): SayTerm {
+    switch (serialized.termType) {
+      case 'BlankNode':
+        return this.blankNode(serialized.value);
+      case 'NamedNode':
+        return this.namedNode(serialized.value);
+      case 'Literal': {
+        const { datatype, language } = <SayLiteral>serialized;
```
abeforgit, line 106: For consistency, I prefer the `as` syntax for casting
Will also be a useful habit for when we move to `.gts` components

abeforgit (changes requested):
minor nitpick, looks good otherwise

addon/core/say-data-factory.ts
elpoelma: Yeah makes sense, it's the first time I've used it (as the original lib also uses it)

abeforgit (approved):

PR #1170, opened 2024-03-18
Content vertical alignment in table cell

### LINKED ISSUE #1169: Content vertical alignment in table cell
<p>can use the “vertical-align” css property on the table cell as an inline style. As usual, make buttons, add to both menus. If there are no suitable svgs, make a ticket for them.</p>

https://binnenland.atlassian.net/browse/GN-4707

dkozickis:
### Overview

Content vertical alignment in table cell #1170

#### Connected issues and PRs:

Closes #1169 

### Setup

1. Checkout
2. `npm run start`

### How to test/reproduce

Repeat after me.

https://github.com/lblod/ember-rdfa-editor/assets/769698/38ee4065-6ad1-429f-b1d5-c434f3fc8708

### Challenges/uncertainties

Because there are no Appuniversum icons available have to user "self-made" components, instead of using convenient dropdown component.

### Checks PR readiness
- [X] changelog
- [X] npm lint
- [X] no new deprecations

### DIFF
--- a/addon/components/plugins/table/vertical-align.hbs
+++ b/addon/components/plugins/table/vertical-align.hbs
@@ -1,48 +1,46 @@
 {{#if @controller}}
-  <div class="say-toolbar__button">
-    <div class="say-dropdown options">
-      <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} as |velcro|>
-        <button
-          type="button"
-          class='say-dropdown__button {{if this.dropdownOpen "is-active" ""}}'
-          title={{t 'ember-rdfa-editor.table.vertical-align'}}
-          disabled={{not this.canSetVerticalAlign}}
-          {{on 'click' this.toggleDropdown}}
-          {{velcro.hook}}
-          {{this.setupDropdownButton}}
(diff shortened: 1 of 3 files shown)


abeforgit (changes requested):
works well, but this button:
image
is a separate click target but doesn't do anything

there should only be one click target, the whole icon + arrow, which should open the dropdown, similar to the horizontal alignment controls

addon/components/plugins/table/vertical-align.hbs
```diff
@@ -0,0 +1,48 @@
+{{#if @controller}}
+  <div class="say-toolbar__button">
+    <div class="say-dropdown options">
+      <this.Velcro @placement="bottom" @offsetOptions={{hash mainAxis=6}} as |velcro|>
+        <button
```
abeforgit, line 5: why not just merge these two buttons into one?

abeforgit (changes requested):
still not quite right:
- if you click on the arrow, it works ok
- if you click on the icon next to the arrow, you need to click twice to open the dropdown

PR #1178, opened 2024-03-27
Update to latest ember-appuniversum v3.4.0

piemonkey:
### Overview
Passing icons as components instead of strings, as this uses inline SVGs rather than making an extra request for an svgiconset which fails on embeddable.

This allows and requires the latest ember-appuniversum, 3.4.0, as without that, the passing of components to `@icon` arguments will fail. Any usage of icons using a `string` argument will result in the svgiconset version being used, which will fail on embeddable.

In order to support importing icons in template only components, added support for `.gts` single file templates and updated the tooling to add better type checking supported by these.

Part of https://github.com/lblod/frontend-embeddable-notule-editor/issues/243
Uses the work on https://github.com/appuniversum/ember-appuniversum/issues/482

### Setup
You might need to do some set-up in your editor if you haven't set it up to support the new template syntax, details [in the ember docs](https://guides.emberjs.com/release/components/template-tag-format/#toc_editor-integrations) for highlighting and [the glint docs](https://typed-ember.gitbook.io/glint/getting-started#glint-editor-extensions) for type checking.

### How to test/reproduce
All the icons should work. If you inspect an icon, you should no longer see a `<use>` element inside the `<svg>`, you should just see the svg inline.

### Challenges/uncertainties
While refactoring it was easier to write macros that relied on `'` quotes being used instead of optionally supporting `"`, so I converted them. This results in more consistency between files anyway.

There's a bug with some of the new icon components, noted [here](https://github.com/appuniversum/ember-appuniversum/pull/488) which I have worked around rather than waiting for a new AU release.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


package.json
```diff
@@ -206,7 +211,7 @@
     "webpack": "^5.78.0"
   },
   "peerDependencies": {
-    "@appuniversum/ember-appuniversum": "^2.15.0",
+    "@appuniversum/ember-appuniversum": "^2.17.0 || ^3.0.0",
```
elpoelma, line 214: We should support at least version 2.15.0 to be non-breaking.

addon/components/_private/attribute-editor/index.ts
```diff
@@ -7,13 +7,48 @@ import TransformUtils from '@lblod/ember-rdfa-editor/utils/_private/transform-ut
 import type { ResolvedPNode } from '@lblod/ember-rdfa-editor/utils/_private/types';
 import { Changeset, EmberChangeset } from 'ember-changeset';
 import { trackedReset } from 'tracked-toolbox';
+import { dependencySatisfies, macroCondition } from '@embroider/macros';
+import { importSync } from '@embroider/macros';
+const CheckIcon = macroCondition(
```
elpoelma, line 12: Nice approach to supporting the older versions too :) I wonder if it is useful to move these imports to a seperate general file to avoid code duplication.

elpoelma (changes requested):
Looks good overall, depending on which version I install in the dummy app, the addon correctly switches between inlining the svg's and `use`.
I was also able to easily setup the glint language server in vscode.

package.json
piemonkey: For that we would need to switch the path that the scss is imported from. `2.17.0` is the first version that supports the newer path. I don't know if we can use embroider macros in scss. I'll check.

addon/components/_private/attribute-editor/index.ts
piemonkey: I thought about that, but it would mean we maintain a separate file or files with all of the icon imports we use, as they are all in separate files. Perhaps it would be cleaner though. I'm not sure.

package.json
elpoelma: hmm okay interesting. This would technically mean that this causes a breaking release then :/

addon/components/_private/attribute-editor/index.ts
elpoelma: Yeah it's not super necessary indeed. The icon compatibility layer will not be for the long term anyway.

package.json
elpoelma: But I guess this is not a problem, as we only need to import the ember-appuniversum styles in the consuming app, not in the addon. So depending on which version the host app has installed, they can choose between the two type of import paths. So I think we should just be able to set the minimum version to 2.15.0.

addon/components/plugins/image/insert-menu.ts
```diff
@@ -1,9 +1,17 @@
...
 import { tracked } from 'tracked-built-ins';
+import { dependencySatisfies, macroCondition } from '@embroider/macros';
+import { importSync } from '@embroider/macros';
+const ImageIcon = macroCondition(
+  dependencySatisfies('@appuniversum/ember-appuniversum', '>=3.4.1'),
+)
+  ? // @ts-expect-error TS/glint doesn't seem to treat this as an import
+    importSync('@appuniversum/ember-appuniversum/components/icons/image')
+      .ImageIcon
+  : 'image';
```
elpoelma, line 14: In order to support version 2.15.x, we also need to add an additional macrocondition.
```suggestion
import { dependencySatisfies, macroCondition } from '@embroider/macros';
import { importSync } from '@embroider/macros';
const ImageIcon = macroCondition(
  dependencySatisfies('@appuniversum/ember-appuniversum', '>=3.4.1'),
)
  ? // @ts-expect-error TS/glint doesn't seem to treat this as an import
    importSync('@appuniversum/ember-appuniversum/components/icons/image')
      .ImageIcon
  : macroCondition(
        dependencySatisfies('@appuniversum/ember-appuniversum', '>=2.16.0'),
      )
    ? 'image'
    : 'website';
```

package.json
piemonkey: Hmm, good point, I'll double-check that and set the minimum to 2.15.0 assuming it works.

piemonkey: Minimum is now 2.15.0 again. I've kept the test app on 2.17.0 as there shouldn't be many differences and it means we can stick on the new import paths.

elpoelma: Ok perfect!

elpoelma (approved):

PR #1179, opened 2024-03-28
Drop RDFa support from tables

elpoelma:
### Overview
This PR drops RDFa support from tables, as it is currently not reasonably possible to support them with the `rdfaAware` system.
We currently cannot support `rdfaAware` tables as tables (and table-rows) do not allow div children (or any other children aside from table-specific children). This means we do not have a proper way to store their RDFa.

### How to test/reproduce
- Start up the dummy app
- Visit the 'editable nodes' page
- Insert a table.
- Notice that the table is rendered correctly and that all resizing features work as expected.
- Notice that the table does not contain any RDFa related properties/features

### Challenges/uncertainties
This PR also support classic RDFa support from tables. Unsure if we need to keep that. It makes this PR technically breaking.

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/plugins/table/nodes/table.ts
+++ b/addon/plugins/table/nodes/table.ts
@@ -1,8 +1,14 @@
-import { Node as PNode, type NodeSpec, ResolvedPos } from 'prosemirror-model';
+// Helper for creating a schema that supports tables.
+
+import { Node as PNode, ResolvedPos, type NodeSpec } from 'prosemirror-model';
+import {
+  getRdfaAttrs,
+  rdfaAttrSpec,
+} from '@lblod/ember-rdfa-editor/core/schema';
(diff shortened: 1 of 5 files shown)


piemonkey (commented):
I don't understand the reason to remove the 'classic rdfa' support. I don't know whether it is used anywhere but it feels wrong to make a hidden breaking change.

elpoelma:
> I don't understand the reason to remove the 'classic rdfa' support. I don't know whether it is used anywhere but it feels wrong to make a hidden breaking change.

Yeah that's fair. I'll just add the classic RDFa support back for now, and remove support for the `rdfaAware` API.

piemonkey (approved):

PR #1182, opened 2024-03-29
ci(woodpecker): improved detection of prerelease tags in release pipeline

elpoelma:
### Overview
This PR contains improved detection of prerelease tags in the release pipeline:
- Only perform a stable release if git tag does not contain a pre-release suffix (e.g. `-alpha`, `-dev`, `-next`)
- Perform an npm release with tag `next` if qualifier part of pre-release suffix is equal to `next`

### Challenges/uncertainties
I tried to generically extract the qualifier from the pre-release suffix based on the `CI_COMMIT_TAG` env var and string substitutions (https://woodpecker-ci.org/docs/usage/environment#string-substitution), but was unable to make it work. Apparently woodpecker does not support nested substitutions (remove both prefix and suffix from strings). Ideas to make this work are very welcome :)
I now limited it to the `next` qualifier.

### DIFF
--- a/.changeset/fair-stingrays-float.md
+++ b/.changeset/fair-stingrays-float.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Release to npm with tag `next` if the semantic version qualifier is equal to `next`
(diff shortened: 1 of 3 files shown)


piemonkey (approved):
No ideas for how to remove both prefix and suffix without knowing the length, which would fail if anything goes to double digits :/

PR #1184, opened 2024-04-03
GN-4754: hierarchical lists

abeforgit:
### Overview

Supports all manner of creative list combinations. No matter what styles are
chosen, each list item now keeps track of its path in the list. At each level of the path,
both the style and hierarchicality are recorded, allowing for each list-item to fully render
its marker without any other context.

This also reduces the css rule to simply read the value from the html attribute, and avoids any
bugs related to counters.

To keep this state up to date while editing, a new plugin is added that corrects the paths of every
changed list.

Hierarchicality and style are separated, allowing for lists like 1.a.I.2. Also, the count is kept up
even through bullet-lists.

In unstyled contexts, the html degrades gracefully to non-hierarchical lists. It is impossible to do
hierarchy lists in html only without creating extra elements.

TODO:

- [x] make it backwards compatible (aka make the hierarchy feature optional)

TO ENABLE:

- add `enableHierarchicalList` to list node configs:
e.g.
```ts
      list_item: listItemWithConfig({ enableHierarchicalList: true }),
      ordered_list: orderedListWithConfig({ enableHierarchicalList: true }),
      bullet_list: bulletListWithConfig({ enableHierarchicalList: true }),
```
- add the listTracking plugin to your plugin array
```ts
import { listTrackingPlugin } from '@lblod/ember-rdfa-editor/plugins/list';

  @tracked plugins: PluginConfig = [
    listTrackingPlugin(),
// other plugins
]
```
- add the `@enableHierarchicalList` argument to the list menu component or the sample toolbar
```hbs

      <Plugins::List::Unordered @controller={{@controller}}/>
      <Plugins::List::Ordered @controller={{@controller}} @enableHierarchicalList={{@enableHierarchicalList}}/>
```

supercedes https://github.com/lblod/ember-rdfa-editor/pull/1156

### Setup

just run it

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [ ] no new deprecations


abeforgit:
for testing in your app: you need to add the plugin to your plugin array

brenner-company:
I've only tested these changes within the dummy app, but I've noticed the following already: when switching from a bulleted list to a hierarchical (numbered) list and vice versa, there seems to be an issue regarding consistency. See the screen recording below:


https://github.com/lblod/ember-rdfa-editor/assets/18174827/de9480bf-a0f4-4e33-8555-00678ca00015


https://github.com/lblod/ember-rdfa-editor/assets/18174827/31b084a8-1d5f-45c5-8a2e-f88e1c562678

brenner-company:
I'll try and give it a more thorough look next week 🤞

addon/components/plugins/list/ordered.hbs
```diff
@@ -2,22 +2,37 @@
 {{#if @controller}}
   <Toolbar::Button
     @active={{this.isActive}}
-    @title={{t 'ember-rdfa-editor.ordered-list.button-label'}}
-    @optionsLabel={{t 'ember-rdfa-editor.ordered-list.options-label'}}
+    @title={{t "ember-rdfa-editor.ordered-list.button-label"}}
```
piemonkey, line 5: Is there a preference to use `"` in place of `'`? Our template lint config doesn't give a preference (at least it doesn't error for me if I run it on a file with a mixture) but for the `gts` components, prettier starts looking at them and judges us if we use `"` (other than if there are e.g. nested quotes). We can configure this differently, but we probably shouldn't each convert files in the opposite direction...

piemonkey (commented):
Looks good to me, just a question around template style changes.

I didn't test much around switching between hierarchical styles and unordered lists, as it seems that changes will be needed from @brenner-company 's feedback. One thing that would make sense to me from a user point of view, is a way to nest unordered lists within an ordered hierarchical list (not necessarily with the hierarchical numbering intact, but the formatting as a single list. I'm thinking something like:
```
1. one
  1.1. one-one
    * some sub-part of one-one
    * another
  1.2. one-two

abeforgit:
> I've only tested these changes within the dummy app, but I've noticed the following already: when switching from a bulleted list to a hierarchical (numbered) list and vice versa, there seems to be an issue regarding consistency. See the screen recording below:
> CleanShot.2024-04-05.at.18.21.19.mp4
> CleanShot.2024-04-05.at.18.23.48.mp4

thanks, seems I didn't test enough with uncollapsed selections

abeforgit:
> Looks good to me, just a question around template style changes.
> 
> I didn't test much around switching between hierarchical styles and unordered lists, as it seems that changes will be needed from @brenner-company 's feedback. One thing that would make sense to me from a user point of view, is a way to nest unordered lists within an ordered hierarchical list (not necessarily with the hierarchical numbering intact, but the formatting as a single list. I'm thinking something like:
> 
> ```
> 1. one
>   1.1. one-one
>     * some sub-part of one-one
>     * another
>   1.2. one-two
> ```

But this is possible, is it not?

abeforgit:
> Looks good to me, just a question around template style changes.
> 
> I didn't test much around switching between hierarchical styles and unordered lists, as it seems that changes will be needed from @brenner-company 's feedback. One thing that would make sense to me from a user point of view, is a way to nest unordered lists within an ordered hierarchical list (not necessarily with the hierarchical numbering intact, but the formatting as a single list. I'm thinking something like:
> 
> ```
> 1. one
>   1.1. one-one
>     * some sub-part of one-one
>     * another
>   1.2. one-two
> ```

piemonkey:
> > ```
> > 1. one
> >   1.1. one-one
> >     * some sub-part of one-one
> >     * another
> >   1.2. one-two
> > ```
> 
> But this is possible, is it not?

Since 'unordered' or 'bullets' isn't an option in the menu, I can find a way to continue the hierarchical list around the bullet points. I can get this:
image
but the bullet points are clearly not part of the rest of the list.

addon/components/plugins/list/ordered.hbs
abeforgit: I'm more used to seeing double quotes in html-like languages, but I really don't have any preferences
You're right that we should configure everything to match though. Lets look at what the ember default setting is and go with that

abeforgit:
> > > ```
> > > 1. one
> > >   1.1. one-one
> > >     * some sub-part of one-one
> > >     * another
> > >   1.2. one-two
> > > ```
> > 
> > 
> > But this is possible, is it not?
> 
> Since 'unordered' or 'bullets' isn't an option in the menu, I can find a way to continue the hierarchical list around the bullet points. I can get this: image but the bullet points are clearly not part of the rest of the list.

seems I broke it, looking into it

abeforgit:
> I've only tested these changes within the dummy app, but I've noticed the following already: when switching from a bulleted list to a hierarchical (numbered) list and vice versa, there seems to be an issue regarding consistency. See the screen recording below:
> CleanShot.2024-04-05.at.18.21.19.mp4
> CleanShot.2024-04-05.at.18.23.48.mp4

Funnily enough this is actually not a regression. If you try this with your current version, without hierarchy, but with mixed ordered/unordered lists, you'll see the same behavior. Its due to the logic of the toggle buttons: they take the first "common ancestor" ul or ol and operate on that node, leaving the others alone. I agree this is not very intuitive, but it's at least not caused by this PR so I'm inclined to open it as a separate ticket.

brenner-company (approved):

brenner-company:
Good to go ✅ Regarding the 'unintuitive' behaviour mentioned above: an additional ticket to tackle this would be ideal (checked with Johan).

abeforgit:
@brenner-company see #1187 for followup, note it's not currently on our short-term roadmap, so let us know if it's urgent for you

PR #1188, opened 2024-04-26
Add selectNode helper to ember-node and improve types

piemonkey:
### Overview
I noticed that many of the variable plugin nodes as well as image nodes have a copy-pasted version of this that I wanted to copy-paste into the snippet placeholder, so moved it here instead. I then noticed that there was a modifier to do this that wasn't being used. Since the node args are already a kind of public api for ember nodes, I thought it was better to keep that than switch to the modifier instead.

Done as part of the snippet placeholder work: https://binnenland.atlassian.net/browse/GN-4782

### Setup
N/A

### How to test/reproduce
Image nodes still work as expected. Can also be tested as part of the snippet placeholder PR when that lands.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (approved):
nice! indeed makes sense to pass it in as its something almost every node needs

PR #1192, opened 2024-05-16
Improve wrapping of list nodes with RDFa

piemonkey:
### Overview
A slightly weird collection of tweaks and fixes. Makes list nodespecs no longer RDFa aware, prevents finding of RDFa child nodes from including the parent and adds a command which immitates `wrapIn` but goes up by parent nodes to find something that can be wrapped.

https://binnenland.atlassian.net/browse/GN-4797

### Setup
N/A

### How to test/reproduce
Create arbitrary list nodes and tables and verify that the 'wrap with block literal' and 'wrap with block resource' buttons work in a way that at least makes sense to Prosemirror.

### Challenges/uncertainties
It seems that this kind of behaviour isn't well defined within prosemirror itself and there are many functions with documentation that says things like 'find a set of wrapping nodes that would allow a node of the given type to appear' but that don't handle nodes with very specific children (such as bullet_list) well as they don't include anything that necessarily already exists in the document.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/hip-parrots-shout.md
+++ b/.changeset/hip-parrots-shout.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+Better handle wrapping with RDFa blocks for nested node structures
--- a/.changeset/orange-icons-search.md
+++ b/.changeset/orange-icons-search.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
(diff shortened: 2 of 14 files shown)


abeforgit (approved):
nice work! slightly awkward behavior when unindenting a wrapped nested list, but I'm filing that under "who cares" for now :stuck_out_tongue:

Issue #1194, opened 2024-05-29
Feature request to add upper-alpha list style


brenner-company:
For the Kaleidos application, we're getting ready to migrate our forked version of `ember-rdfa-editor` to the official package, but as to date we're missing the following small bit functionality: allowing the user to add a ordered list of style type `upper-alpha`.

![CleanShot 2024-05-29 at 09 54 41@2x](https://github.com/lblod/ember-rdfa-editor/assets/18174827/b6547539-c67c-4b40-9c70-18257bc1dc23)

Is this something that may be added? If yes, a PR can be provided by me with the necessary changes.

abeforgit:
sorry for not answering sooner, yes this can definitely be added, if you have a pr ready feel free to post it (you never need to ask for permission for that :) )
if not we can probably pick it up sometime this sprint

PR #1196, opened 2024-06-05
Addition of ordered list style type upper-alpha

brenner-company:
### Overview
Concerning https://github.com/lblod/ember-rdfa-editor/issues/1194: the addition of ordered list style type `upper-alpha`.

### Challenges/uncertainties
I'm not sure of this change also has to be covered by the e2e test, but I've made a small addition of this (style) type within `e2e/files/ordered-list.html`. Though I can't seem to get the `list` tests running (getting a process related error from Playwright). If yes, can someone double check this for me?

### Checks PR readiness
- [x] npm lint
- [x] test (general)
- [x] e2e tests


addon/plugins/list/nodes/list-nodes.ts
```diff
@@ -319,6 +323,8 @@ function renderEntry(entry: ListPathEntry) {
   const { style, pos } = entry;
   if (style === 'lower-alpha') {
     return indexToAlpha(pos);
+  } else if (style === 'upper-alpha') {
+    return indexToAlpha(pos).toUpperCase();
```
brenner-company, line 327: This seems the most simple way to accomplish this, but if extending `indexToAlpha` a bit to support upper case fits better here, just let me know.

elpoelma: this solution looks good to me!

elpoelma (commented):
This solution looks good to me! I'm investigating the issues with playwright when using docker. Locally it runs without issues, but when running it in a docker container, I also get `process` errors.

elpoelma:
@brenner-company I updated the playwright snapshot, so the e2e tests should run correctly on both CI and when running them locally. There are still issues with the playwright-docker setup, but that does not impact local/CI runs.

brenner-company:
@elpoelma Thanks for the follow-up! I've added the missing changeset ✅

PR #1197, opened 2024-06-11
Add transaction utils and expose the rdfa-aware addProperty function

abeforgit:
### Overview

Add a few util functions to avoid duplicating logic in plugins.

Main new thing: transactionCombinator

It can chain functions of the signature state => TransactionResult, which I've called transaction monads here, not entirely sure it's a "true" monad but it feels like one.

It can optionally start from an existing transaction + state, or it makes its own. The result it produces is a single transaction that combines all the steps in the given transactions. It also collects the result states of its chain, allowing the caller to inspect potential failure of specific transactions in the chain.

I added this cause I needed a more flexible system than our current way of combining commands. I mainly wanted to be able to use it outside commands, and to avoid the history issues that come with using multiple transactions to do one logical thing.

I also added the initial state to the controller's `withTransaction` callback argument so that the combinator can be used there.

And lastly I turned the rdfa-aware addProperty command into a transactionmonad, so that it can be used everywhere in the plugins without generating extra transactions.

https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/440
https://github.com/lblod/frontend-reglementaire-bijlage/pull/236
### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/utils/_private/rdfa-utils.ts
```diff
@@ -315,3 +326,90 @@ export function isLinkToNode(triple: OutgoingTriple): triple is LinkTriple {
...
   );
 }
+
+export type AddPropertyArgs = {
+  /** The resource to which to add a property */
+  resource: string;
+  /** Property to add */
+  property: OutgoingTriple;
+  /**
+    A transaction to use in place of getting a new one from state.tr
```
piemonkey, line 336: This comment refers to a transaction but it's on a member of type `EditorState`. In fact, I can't see this type being used anywhere?

addon/utils/transaction-utils.ts
```diff
@@ -0,0 +1,41 @@
+import { EditorState, Transaction } from '@lblod/ember-rdfa-editor';
+export interface TransactionResult<R> {
+  initialState: EditorState;
+  transaction: Transaction;
+  result: R;
+}
+
+export type TransactionMonad<R> = (state: EditorState) => TransactionResult<R>;
+
+export function transactionCombinator<R>(
```
piemonkey, line 10: I feel like this could do with some docs to explain usage.

piemonkey (commented):
Looks good to me. Just some minor comments and I'll actually test it in the review of the plugins.

abeforgit:
Also renamed the result type, as TransactionResult wasn't really correct

piemonkey (approved):

piemonkey (approved):

PR #1200, opened 2024-06-17
Fix issues related to `equals` method defined on RDF term instances

elpoelma:
### Overview
This PR includes some changes in how the `equals` method/property is defined on RDF term instances.
It defines `equals` as a prototype method on RDF term instances instead of as arrow functions.
This has two implications:
- The `equals` function is no longer defined on RDF term instances themselves, but rather on their prototype.
- Comparison functions (such as the one used by prosemirror-devtools) no longer fail due to them not supporting function properties.
- Spreading an RDF term results in losing the `equals` method.

None

### How to test/reproduce
- Open up the dummy app
- Insert a sample document containing RDFa
- Ensure that `functions are not supported` error message no longer shows up.
- There should be no issues with missing/undefined `equals` properties.

### Challenges/uncertainties
**The change/rollback included in this PR does have some (possibly unwanted) side-effects:**
- As mentioned before , applying the spread operator on an RDF term instances results in losing the `equals` method. Typescript tries to handle this in the best way possible, but may sometimes make mistakes. An example can be found on: 
[Typescript playground] (https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgCrQLbIN4FgBQyyGEYAFgPYAmAggBQCUAXMgG4XBUDcBRJ51AEKMW7Tj3wBfAgQQAbOAGdFyAMpwMABzkR0ULMC06S4FXqx5CxUpVrIAvMkYOAfDl5FkCCiEUUdAHRyFADmdADkALI21Mg0TnBQUBQA7sgwAK4gCGDAPgzhDB7SVvy2wgyWnl4+foHBYVExVMiCTprJYBRgAJ6aKGXUBQwSRCUlBAD0AFTTyABiwFCKYMiKSCCJecjTkzL43r6rkPoOyCAQaepGupiMEpOTyABKWSopwAIZq8DKGRCKAgnDABQa0e5TJ6vXzID5fH5-AFAzCg5oVCSyWqrRQdCBwKgQKjmM7YAJk4GSDH4R7IAAycBAIQycBCKEU0FY0BqSQgOTkPWQmmACAA1ioMppkOQ4KsAAZgmiy5C-ZAM1VJVLpLI5PIgVUgFrkDIqXHsqCclopMgQPU4qB4qigEIBAh2h2E8yogTgkb7GnytFKlVwQWdbp9AbNfUtEDdUMAjmEgJoMgq7w8vkCkLATkqEO9frrKDATSraDJKAu6lPN34j2YACMXvKEPwU1mal5Pha6xtWwoOz2bYOWKlmAATCxiY4LlcNNpbvpW4cVmtcXWiROSWSAsDx5TISmVWmKBmwK71wTN-px83qPRfcOafTGczWWtE1ApamQGLkAHvUEIMVBWYA5DkZACRgUBCQAGmQAAjb5lQRc5umTVBUxUYNkDkQxPhlXVkAoGApQjIsS1WKh-ilAdPjQ1YRVjD5GWVUiAPKYDIIgaCLhaHxvxQDpunDfpiK-cgUAoBCACteR+MB2TkGAq1rK9zFvMF0SAA)

**Other possible approaches:**
- Using simple data objects to represent the RDF terms and using separately defined `equals` functions. This does have the drawback that we cannot leverage the RDFJS interfaces.
- Ensuring that the comparison library used by the prosemirror devtools adds support for function properties. This can be a long-term solution which is possibly the best approach, as it may occur in other situations that prosemirror node-attributes contain arrow function properties.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/lucky-experts-kneel.md
+++ b/.changeset/lucky-experts-kneel.md
@@ -1,0 +1,11 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Define `equals` as a prototype method on RDF term instances instead of as arrow functions.
+
+**Why is this necessary?**
+Defining `equals` as a prototype method ensures that is not part of the RDF term objects themselves. This makes it easier/less error-prone to compare objects containing these terms. (e.g. the object-comparison library used by prosemirror-tools does not support function properties)
+
+**Possible drawbacks of this approach**
+This approach does have the drawback that spreading an RDF term results in losing the `equals` method.
--- a/addon/core/say-data-factory/blank-node.ts
+++ b/addon/core/say-data-factory/blank-node.ts
@@ -13,9 +13,9 @@ export class SayBlankNode implements RDF.BlankNode {
     this.value = value;
   }
 
-  equals = (other?: Option<SayTerm>) => {
+  equals(other?: Option<SayTerm>) {
     return (
       !!other && other.termType === 'BlankNode' && other.value === this.value
     );
-  };
+  }
 }
--- a/addon/core/say-data-factory/default-graph.ts
+++ b/addon/core/say-data-factory/default-graph.ts
@@ -16,7 +16,7 @@ export class SayDefaultGraph implements RDF.DefaultGraph {
     // Private constructor
   }
 
-  equals = (other?: Option<SayTerm>) => {
+  equals(other?: Option<SayTerm>) {
     return !!other && other.termType === 'DefaultGraph';
-  };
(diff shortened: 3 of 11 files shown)


lagartoverde (approved):
Looks good, maybe you should wait for another review that knows about the implications of this change (cough cough Arne) :)

PR #1201, opened 2024-06-20
Convert horizontal alignment menu into icons

lagartoverde:
### Overview
Converted the horizontal alignment menu into icons in order to comply with the new menu proposed by Saska

GN-4854

### Setup
None

### How to test/reproduce
Go to the editor, click on a paragraph and the horizontal alignment menu should now be icons instead of text, it should work the same way as before

### Challenges/uncertainties
I changed the color of the icons given by Saska to match our current colors, this should change when we implement the new menu styles

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/cold-maps-agree.md
+++ b/.changeset/cold-maps-agree.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": minor
+---
+
+Convert horizontal alignment menu into icons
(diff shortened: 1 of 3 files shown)


elpoelma (commented):
Looks good to me! I would just add a `title` attribute to each of the menu entries.

PR #1202, opened 2024-06-24
Upgrade to ember-template-imports 4.1.1

piemonkey:
### Overview
This fixes a problem on dependencies using ember-appuniversum v3, which was being hidden by a working combination of versions existing in the lockfile. This doesn't fix these versions for the test-app in this repo.

Plugins PR tbc

### Setup
N/A

### How to test/reproduce
Test that everything works... Should also be tested in plugins PR.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/nasty-taxis-pull.md
+++ b/.changeset/nasty-taxis-pull.md
@@ -1,0 +1,5 @@
+---
+"@lblod/ember-rdfa-editor": patch
+---
+
+Update ember-template-imports to 4.1.1
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "ember-cli-babel": "^8.2.0",
     "ember-cli-htmlbars": "^6.2.0",
     "ember-focus-trap": "^1.1.0",
-    "ember-template-imports": "^3.4.2",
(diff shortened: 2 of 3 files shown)


abeforgit:
maybe we should regen the lockfile with pnpm 9 also?

abeforgit (approved):

piemonkey:
@abeforgit The reason I never asked for any reviews on this one (I meant to turn it into a draft), is that this is actually a breaking change in some cases. While a user of the library does not need to depend on ember-template-imports, it seems that any user of it that *does* depend on it, needs to use v4 also (I haven't tested which exact versions, but I tested with 4.1.1). I don't know a way to encode this in the dependencies. I'm also not sure how AU gets away with this, but I suspect that it's to do with how ember-nodes are rendered by code here, but provided by plugins, so it's possible that it's a unique problem for those two libraries.

I'll put in another PR with more breaking updates. I'll also see if just updating ember-template-imports on the plugins, without the other updates works, and if so, put in two PRs there.

PR #1204, opened 2024-07-03
Upgrade many dependencies - breaking

piemonkey:
### Overview
Since it seems like upgrading ember-template-imports was a breaking change*, we might as well include many updates into a single major release. I updated all of the peer dependencies, with the only version kept for backwards compatibility being ember-intl v6 as that's very new and includes a breaking api change, so this allows clients on v6 to not have to make too many changes.

I also ran `pnpm up`, which updates dependencies to their latest versions supported by their version specifiers. Unfortunately there were some related changes to the types in these updates. It's possible that further updates could be made, for example, major versions of dependencies.

*The [ember-template-imports upgrade](https://github.com/lblod/ember-rdfa-editor/pull/1202), while possible to get working on plugins by upgrading the same libraries, I wasn't able to find a way to get it to work on GN. 

Supersedes https://github.com/lblod/ember-rdfa-editor/pull/1202
Plugins PR and GN PR tbc.

### Setup
N/A

### How to test/reproduce
Best tested using the built version (or linking) to GN, but can also be tested in the test app. Everything should work as before.

### Challenges/uncertainties
It seems non-trivial to manage transitive dependencies, even in pnpm and it's unclear what should be in peerdeps, for example plugins seems to need some specific dependency versions, but GN does not need them...

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


elpoelma (commented):
The test-app seems to work without issue, just wondering if we could mark `ember-template-imports` as an optional peerdep, to ensure if the host-app has it installed, it uses the same version range?

piemonkey:
@elpoelma You're talking about the crash in plugins? That should be fixed now.

elpoelma:
> @elpoelma You're talking about the crash in plugins? That should be fixed now.

Sorry, that edit was meant for my comment on the plugins.

piemonkey:
@elpoelma It indeed seems like we can at least get a peerdep warning by using optional peerdeps. I've done that and included a couple of other libraries that can be problematic but aren't true peer deps. For at least floating-ui/dom though it's still possible to get a build error when the client has a conflicting version as a transitive dependency, but I don't know whether we can solve that.

abeforgit:
I'm gonna yolo-merge and release these, I keep running in circles between the prs and it's not giving me any more or less confidence

PR #1206, opened 2024-07-18
Add option to configure editor as non-editable

elpoelma:
### Overview
This PR adds an option to configure the editor as non-editable.
It can be configured in two ways:
- Using the ember `Editor` component, by passing the `editable` argument. The editor will also correctly react to value changes of this argument.
- Using an instance of the `SayController` class, by executing the `toggleEditable` method

### How to test/reproduce
- Start the test-app
- Open the index page
- Test out the `toggle editable` button. This button uses the `Editor` component argument approach
- The `toggleEditable` method of the editor controller should yield the same results.

### Challenges/uncertainties
- The toolbar/sidebar do not react to the editability of the document. At the moment, they should be configured to be enabled/disabled by the user of this addon.
- Unsure if it is useful to keep both pathways to configure the editability of a document (controller vs component argument)

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


addon/components/editor.ts
```diff
@@ -111,6 +116,9 @@ export default class RdfaEditor extends Component<RdfaEditorArgs> {
       nodeViews: this.args.nodeViews,
       defaultAttrGenerators: this.args.defaultAttrGenerators,
       keyMapOptions: this.args.keyMapOptions,
+      editable: () => {
+        return !(this.editable === false);
```
lagartoverde, line 120: you can do !!this.editable I think

lagartoverde (approved):
Works correctly, seems like it takes some time to make it editable again, but I guess the use of this is not to make editable/not editable quickly

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lblod/ember-rdfa-editor",
-  "version": "9.14.1",
+  "version": "10.0.2",
```
abeforgit, line 3: given we're merging into a 9.x branch, I feel this should be here, same with all the changelog mentions of 10.x

abeforgit: if you developed this on the master branch and then pr-d to 9.x, you'd have to cherry-pick I think

abeforgit (changes requested):

abeforgit:
I'm missing also the visual changes we do in GN (placeholders from yellow to grey, etc)

I think we should try to make it as ootb as possible, requiring no extra css from embeddable users until they specifically ask for customization

abeforgit:
With that said, we might even want to swap out the toolbar and sidebar as well with this toggle

addon/core/say-controller.ts
```diff
@@ -63,6 +63,19 @@ export default class SayController {
     return this.mainEditorView.domParser;
   }
 
+  get editable() {
+    return this.mainEditorView.editable;
+  }
+
+  toggleEditable() {
```
abeforgit, line 70: I tend to prefer explicit `setEditable(editable: boolean)` interfaces, they're more flexible than toggles, and toggle behavior is trivial to implement in the consumer

abeforgit (changes requested):

package.json
elpoelma: Yeah, it seems I forgot to rebase this PR on the 9.x branch, will do that now

elpoelma: should be fixed now.

elpoelma:
> With that said, we might even want to hide the toolbar and sidebar as well with this toggle

I'm not sure if that is the responsibility of this addon? As the user can configure themselves whether to show the sidebar and toolbar, essentially this addon does not have control over their visibility. We can however, control that visibility in the embeddable package.

addon/components/editor.ts
elpoelma: Not in this case I think, as we do not want to trigger it when `editable` is `undefined` or `null`, this is why we strictly compare it to `false`.

elpoelma:
> I'm missing also the visual changes we do in GN (placeholders from yellow to grey, etc)
> 
> I think we should try to make it as ootb as possible, requiring no extra css from embeddable users until they specifically ask for customization

Agreed, will look into adapting the css. We might then also have to do some adaptions of the plugins css.

abeforgit:
> > With that said, we might even want to hide the toolbar and sidebar as well with this toggle
> 
> I'm not sure if that is the responsibility of this addon? As the user can configure themselves whether to show the sidebar and toolbar, essentially this addon does not have control over their visibility. We can however, control that visibility in the embeddable package.

yeah ok fair enough

abeforgit:
still "just works" after all this time :)

PR #1209, opened 2024-08-08
Adjust approach to parse tables

elpoelma:
### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


lagartoverde (approved):
Works good, as you said we should create different PRs to parse from the most popular editors, I think most of them are wrong, for example in the following screenshot I pasted from google docs, our table and libreoffice 
image

elpoelma:
> Works good, as you said we should create different PRs to parse from the most popular editors, I think most of them are wrong, for example in the following screenshot I pasted from google docs, our table and libreoffice image
> 
> Maybe we should create Jira issues for this?

I added a Jira ticket for this :)

PR #1212, opened 2024-08-09
Add concept of imported resources to editor

piemonkey:
### Overview
This is covers  2 parts of the 3-part story of snippet rdfa linking:
1. Snippet creation: Defining 'placeholder' resource URIs that *will* exist in the final document, making them accessible to the RDFa tools when creating a snippet and allowing us to create links *from* those resources to nodes in the snippet. Links *to* those resources is more troublesome and will come in a later PR.
2. Template creation: Mapping placeholder resource URIs to real resource URIs that exist in the document.

The UI for linking placeholder resources to real ones is very rough, but should work for now.

##### connected issues and PRs:
Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/462
 and RB PR: https://github.com/lblod/frontend-reglementaire-bijlage/pull/263.

### Setup
Needs to be tested by being used in the plugins repo.

### How to test/reproduce
See plugins PR for test steps.

### Challenges/uncertainties
There were some problems with types after converting a component to gts and adding the necessary types packages. I think these were just being hidden by being assigned to `any`, so I had to convert some template only components to use gts to correct these.

### Checks PR readiness
- [ ] UI: works on smaller screen sizes
- [ ] UI: feedback for any loading/error states
- [ ] Check cancel/go-back flows
- [ ] Check database state correct when deleting/updating (especially regarding relationships)
- [x] changelog
- [x] npm lint
- [x] no new deprecations

PR #1213, opened 2024-08-09
Messy version of allowing resources to link to imported resources

piemonkey:
### Overview
To allow for linking from resources in a snippet to those imported, adds imported resources to those listed for creating relationships, but actually the linking only works if we create a NamedNode relationship instead of a ResourceNode. This is just a draft as we need to figure out a better way to do this.

Made to https://github.com/lblod/ember-rdfa-editor/pull/1212 as that is the more finalised part of the design.

### Setup
Link to plugins PR.

### How to test/reproduce
See plugins PR for most details. When creating a link to an imported resource from a resource in the snippet, click 'add relationship' as normal, but create a NamedNode link with a manually entered URI.

### Challenges/uncertainties
Need to figure out the right way to do this.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/addon/components/_private/rdfa-editor/relationship-editor/modal.gts
+++ b/addon/components/_private/rdfa-editor/relationship-editor/modal.gts
@@ -46,7 +46,7 @@ export default class RelationshipEditorModal extends Component<Sig> {
             id={{formId}}
             @onSubmit={{this.save}}
             @controller={{@controller}}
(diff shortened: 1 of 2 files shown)


elpoelma:
@piemonkey do you think this PR is still necessary? I don't think this is really a usecase right?

PR #1214, opened 2024-08-16
Add the ability to update an attribute without registering into the hstory

lagartoverde:
### Overview
Now you can update attribute without registering the change in the history. Needed for loading the regulatory statements in GN.
I also added some missing types

GN-4961

### Setup
None

### How to test/reproduce
You can test with my GN pr I will post below when creating it 

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/chilly-experts-confess.md
+++ b/.changeset/chilly-experts-confess.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+updateAttribute now allows a third argument to avoid adding the change to the history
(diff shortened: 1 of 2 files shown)


lagartoverde:
https://github.com/lblod/frontend-gelinkt-notuleren/pull/700

elpoelma (changes requested):
Seems to work as expected, just some comments:
- The type annotations should be include in the `EmberNodeArgs` type, this way you don't also need to include type annotations in the header of the method implementation itself.
- Ideally, `ignoreHistory` should be optional
- This approach works well, especially for situations where we want to execute non-undoable events just after opening a document. Aside from that, I think it could be interesting to propose (to Marijn Haverbeke) an extension to the prosemirror-history which would allow us to append a transaction (using a meta key) to the last event recorded by the history plugin, instead of creating a new event => this would allow us to combine multiple transactions into one event, which could come in handy in other situations. I know this is already being done with transactions that happen quickly after each other (like when typing text), but there doesn't seem to be a public facing API for it.

elpoelma (approved):

PR #1216, opened 2024-09-04
Replace vm-browserify with a stub to avoid breaking embeddable

piemonkey:
### Overview
This was only a dependency through a chain from graphy dataset -> stream-browserify -> crypto-browserify, it is not actually used. It does throw an error on build if nothing is included though, so use a noop stub. This allows us to avoid the build error while also allowing embeddable builds to succeed as they were failing with vm-browserify included as a dependency. It turned out that overriding the ember-auto-import version used by ember-data-table wasn't actually necessary.

N/A

### Setup
N/A

### How to test/reproduce
Link to embeddable and verify both that things work but also that `pnpm pack` in the embeddable-say-editor dir passes.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


elpoelma:
As I recall correctly this issue (the requirement of `vm-browserify`) happened due to a downgrade of `asn1.js` in one of the transitive dependencies right? The downgraded version of `asn1.js` (4.10.1) relies on the existence of the `vm` package. Are we sure this isn't used anywhere?

piemonkey:
> If I recall correctly this issue (the requirement of `vm-browserify`) happened due to a downgrade of `asn1.js` in one of the transitive dependencies right? The downgraded version of `asn1.js` (4.10.1) relies on the existence of the `vm` package. Are we sure this isn't used anywhere?

There seem to be 2 potential places this could be needed. Crypto-browserify depends indirectly on parse-asn1. Crypto-browserify is required by webpack v4 and by the use of `crypto` in @graphy.dataset. The webpack 4 dep is what lead me to override the ember-auto-import dependency, but I don't think that's actually necessary. The error we were seeing during the build process is from the `vm` module not being found, which means that if we actually needed `vm` then we would already be seeing problems.

For more certainty on the webpack 4 side, we could re-add the override of ember-auto-import, which then means we only use webpack 5, which does not have this dependency.

I followed the rabbit hole on the @graphy side for more certainty, the only `crypto` call it makes is `crypto.createHash('sha256')`, this in turn uses create-hash -> sha.js -> safe-buffer -> buffer. buffer looks to use vm for testing only.

PR #1217, opened 2024-09-17
Add doNotClean option to initialize and setHtmlContent

piemonkey:
### Overview
Allows for HTML generated by the editor to be passed in (e.g. from a saved document) without it being passed through the processing steps intended for handling clipboard contents. This leaves empty elements intact, for instance.

Fixes the issue around empty lines being removed on save.

Jira ticket: https://binnenland.atlassian.net/browse/GN-5051
RB PR: https://github.com/lblod/frontend-reglementaire-bijlage/pull/278
GN PR: https://github.com/lblod/frontend-gelinkt-notuleren/pull/718

### Setup
N/A

### How to test/reproduce
Create a document with newlines with no content. Refresh the page, which stores then re-initializes the HTML from localstorage. Previously this was removing empty lines, but now they remain intact.

### Challenges/uncertainties
I was unsure of which steps we should keep in the processing. I kept the sanitization as it is still user input, but removed the other processing. We could re-enable some parts of it, I think it's the docx handling that's causing the problem.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/strong-foxes-remain.md
+++ b/.changeset/strong-foxes-remain.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add option to SayController.initialize() and SayView.setHtmlContent() to prevent cleaning input HTML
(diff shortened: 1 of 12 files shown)


elpoelma (approved):
Good addition! :)
- Empty lines/elements no longer get removed
- Copying text with empty lines from google docs still works as expected (as google docs puts `br` elements inside the 'empty' `p` elements)
- Is the `cleanEmptyElements` operation executed by `cleanDocx` necessary?

PR #1219, opened 2024-10-01
Add util function to strip specially marked elements when publishing

piemonkey:
### Overview
This is the simplest approach to this problem, so I've gone for that due to the lack of clarity on how this would ideally be used. This most likely could be pulled out/copied to be used in the pre-publish flow if needed.

Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/483
PR for GN: https://github.com/lblod/frontend-gelinkt-notuleren/pull/723.
Jira ticket: https://binnenland.atlassian.net/browse/GN-4999

### Setup
If not testing linked to the plugins repo (to have access to the template comment plugin) or within GN, will need to modify a nodespec or markspec to set the `HIDE_FOR_PUBLISH_ATTR` property to be `true`.

### How to test/reproduce
Create a document with the nodespec you added the attribute to and click 'Show Export For Publish' and verify the elements are removed. The other exports should work as before.

### Challenges/uncertainties
Understanding how this is actually intended to be used as it is not clear from the ticket. It would make more sense to me if the publishing flow involved creating a new entity with a 'frozen' document, exported from the editor and stored in the triplestore; allowing for this processing to be more nuanced and avoiding the need for the publish flow to process the content of the document.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/slow-games-cry.md
+++ b/.changeset/slow-games-cry.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add util for filtering out html elements marked with a special attribute, to allow for filtering these on publish
--- a/addon/components/debug-tools.hbs
+++ b/addon/components/debug-tools.hbs
@@ -1,10 +1,13 @@
 {{! @glint-nocheck: not typesafe yet }}
 <div {{this.setUpListeners}}>
-  <button type='button' {{on 'click' this.showStyledExportPreview}}>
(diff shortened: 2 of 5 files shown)


lagartoverde (approved):
It works, but I can't see how are we going to use this on the publication flow, unless we use a headless editor at some point

abeforgit (approved):

PR #1229, opened 2024-11-19
Add ability to manage document-external triples

abeforgit:
### Overview

Add ability to add external (= loose) triples to a resource node

External triples are data not connected to an in-document node. They can be used
to define knowledge about URIs which are not mentioned in the document. 

It's expected this will be rather rare, but crucial for some embedded usecases.

External triples are serialized inside a new container inside the already
exising rdfa-container, annotated with the `data-external-triple-container` 
attribute, which is why I don't consider this a breaking change. 

Also adds a new transactionMonad for manipulating these:
`transformExternalTriples`, and 2 shorthand wrappers: `addExternalTriple` and
`setExternalTriples`.

It also expands the rdfa-editor sidebar tools with an external triple section.

https://binnenland.atlassian.net/browse/GN-5170
[plugin pr](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/510)
[embeddable pr](https://github.com/lblod/frontend-embeddable-notule-editor/pull/269)

### How to test/reproduce

There's a few automated tests that hopefully document a bit how it works. But should be fairly straightforward, you add triples and they stick around when you reload the doc, and otherwise don't get in the way of anything

I've done my best to keep the commits reasonably atomic, but everything is called metaTriples before the rename commit, keep that in mind if you're browsing the history.
### Challenges/uncertainties

- It relies too much on a sort of quirk of the rdfa-processing logic for my liking, that whole thing needs a more robust implementation, but that'd likely involve writing our own parser.

- has somewhat weird but also kind of logical behavior when you mention the same resource in 2 different nodes: it collects all the knowledge about each subject onto all nodes that talk about it

kind of nice in a way, but also kind of not, cause if you edit in one location, you now have conflicting data.
But I don't have time to make document-wide editing logic. It's starting to feel a lot like a document-agnostic way of adding/editing information is in the cards though, which would be really cool.

- I know external-triple-form duplicates a lot of boilerplate from outgoing-triple-form, but I couldn't find a nice extraction point in a reasonable amount of time, as it ultimately comes down to building a form library

### Checks PR readiness

- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/components/_private/rdfa-editor/external-triple-editor.gts
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return null;
+  }
+  closeModal = () => {
+    this.editModalOpen = false;
+  };
+  updateExternalTriples = (trip: FullTriple) => {
+    let tr;
+    if (isSome(this.indexBeingEdited)) {
+      console.log('tripe', trip);
```
piemonkey, line 154: console.log

addon/components/_private/rdfa-editor/external-triple-form.gts
```diff
@@ -0,0 +1,337 @@
...
+class SelectField<T> extends Component<SelectFieldSig<T>> {
+  <template>
+    <FormField @name={{@name}} @errors={{@errors}} @required={{@required}}>
+      <:label>
+        {{yield}}
+      </:label>
+
+      <:default as |id|>
+
+        {{! @glint-expect-error our version of powerselect doesn't have good types}}
```
piemonkey, line 294: We can steal the types that I added to plugins if it helps...

addon/components/_private/rdfa-editor/property-editor/modal.gts
```diff
@@ -0,0 +1,65 @@
...
+    property?: PlainTriple | ContentTriple;
+    onCancel: () => void;
+    onSave: (property: PlainTriple | ContentTriple) => void;
+    modalOpen: boolean;
+    title?: string;
+    termTypes: SupportedTermType[];
+  };
+}
+
+export default class PropertyEditorModal extends Component<Sig> {
```
piemonkey, line 26: Looks like the `.hbs` is still there for this one.

addon/core/schema.ts
```diff
@@ -250,57 +310,60 @@ export function renderInvisibleRdfa(
...
-        }
+        propElements.push(literalSpan(predicate, object));
         break;
       }
     }
   }
+  const externalTriples = nodeOrMark.attrs['externalTriples'] as FullTriple[];
+  if (externalTriples.length) {
+    const externalElements = [];
+    for (const fullTriple of nodeOrMark.attrs[
```
piemonkey, line 333: nit: `for (const fullTriple of externalTriples) {`?

addon/nodes/doc.ts
```diff
@@ -117,10 +105,12 @@ export const docWithConfig = ({
       },
     ],
     toDOM(node) {
+      console.log('rendering doc', node);
```
piemonkey, line 108: console.log

addon/nodes/doc.ts
```diff
@@ -117,10 +105,12 @@ export const docWithConfig = ({
...
       const attrs: Record<string, unknown> = {
         lang: node.attrs['lang'] as string,
         'data-say-document': true,
         properties: node.attrs['properties'],
+        about: node.attrs['subject'],
```
piemonkey, line 113: Doc nodes can define resources now? This seems like a change that is separate from the external triples thing. (just a feeling, I've maybe missed why it's here).

piemonkey (commented):
Some interesting behaviours around subject URIs existing in the document, whether in other external triples or resource nodes, but while confusing, I think there's only one thing that does not make logical sense. This is the fact that the external triples include all of the triples for the subject URI and can therefore editing one causes a duplication, including the old and the new.

I wonder if we should restrict this to document nodes for now to avoid this problem, or whether we can rely on template authors not causing problems with this?

I'll do more playing around but wanted to get my review so far here first.

addon/components/_private/rdfa-editor/external-triple-editor.gts
```diff
@@ -0,0 +1,229 @@
...
+
+interface EditModalSig {
+  Args: {
+    modalOpen: boolean;
+    onCancel: () => void;
+    onSubmit: (trip: FullTriple) => void;
+    triple?: Option<FullTriple>;
+  };
+}
+// TODO: fix as soon as we can import the TOC type
```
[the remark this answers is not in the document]
abeforgit: huh well it seems then a mismatch between the docs and our ember version, they explicitly say it's called `TOC`, but that's good news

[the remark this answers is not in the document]
piemonkey: I think that's what got me when I first tried to use it, but discovered this one later.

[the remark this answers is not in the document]
abeforgit: fixed in [d45d359](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/d45d3593021f2c8ff8e37a28debb170276ddcfae)

addon/components/_private/rdfa-editor/external-triple-editor.gts
abeforgit: fixed in [1c528ec](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/1c528ec9dc93bbeee04271788d37ed3c9dffeab1)

addon/nodes/doc.ts
abeforgit: fixed in [1c528ec](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/1c528ec9dc93bbeee04271788d37ed3c9dffeab1)

addon/components/_private/rdfa-editor/property-editor/modal.gts
abeforgit: fixed in [ce3731c](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/ce3731ca849d2941d4b185e6254781e9fb71b442)

addon/core/schema.ts
abeforgit: fixed in [590ec1a](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/590ec1a8da9429738752371660a9693116faab9e)

addon/nodes/doc.ts
abeforgit: fixed in [251b09a](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/251b09abe5e77eaf7605ea2627dfd2fe2ad98385)

addon/components/_private/rdfa-editor/external-triple-form.gts
```diff
@@ -0,0 +1,337 @@
...
+}
+interface StringFieldSig {
+  Blocks: {
+    default: [];
+  };
+  Args: FieldArgs & { value: string };
+}
+
+// eslint-disable-next-line ember/no-empty-glimmer-component-classes
+class StringField extends Component<StringFieldSig> {
```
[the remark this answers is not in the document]
abeforgit: fixed in [9db555c](https://github.com/lblod/ember-rdfa-editor/pull/1229/commits/9db555c3911a73d4e9c575c6b238eee97b8a5e67)

abeforgit:
> I wonder if we should restrict this to document nodes for now to avoid this problem, or whether we can rely on template authors not causing problems with this?
that's not a bad stopgap indeed
I'll continue for now with the dev build to get a version into the hands of cipal, and look at restricting maybe tomorrow

addon/components/_private/rdfa-editor/external-triple-form.gts
```diff
@@ -0,0 +1,365 @@
...
+interface SelectFieldSig<T> {
+  Args: SelectFieldArgs<T>;
+  Blocks: { default: [] };
+}
+
+// TOCs can't have a generic argument, so in this case we have to make a backing
+// class
+// ref: https://typed-ember.gitbook.io/glint/environments/ember/template-only-components
+class SelectField<T> extends Component<SelectFieldSig<T>> {
+  onChange(selection: unknown, select: Select, event?: Event) {
```
piemonkey, line 312: This needs to either be an arrow function or an `@action`, otherwise `this.args` is undefined.

piemonkey (changes requested):
It seems that adding external triples for more than one subject, while displaying correctly in the 'edit HTML' view, after refresh, only the triples for one of the subjects remain.

abeforgit:
> It seems that adding external triples for more than one subject, while displaying correctly in the 'edit HTML' view, after refresh, only the triples for one of the subjects remain.

great catch once again. I've fixed this, and in doing so also fixed the duplication issue. Now, all external triples about the same subject get collected onto the first node (in document order) that defined that subject. 

Sort of a happy accident, and maybe still not ideal, but I think it's much better than duplicating

piemonkey (approved):
Looks good to me now.

PR #1230, opened 2024-11-20
Feat/recreate uuids rework

lagartoverde:
### Overview
Rework the recreate uuids function so it works better with the actual use case

### Setup
None

### How to test/reproduce
Test on https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/511

### Challenges/uncertainties
As I broke the retrocompatibility for the plugin I decided to do a major release

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/hungry-zoos-hide.md
+++ b/.changeset/hungry-zoos-hide.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Recreate uuids plugin rework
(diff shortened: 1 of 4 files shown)


abeforgit (changes requested):
its maybe silly, but it does seem fairly straightforward to maintain compat and just deprecate the old attrs?
it'll just give us a bit of peace of mind

PR #1231, opened 2024-11-21
Internal/standarization and classes rework

lagartoverde:
### Overview
First part of the standarization ticket, add a class to every node that can be changed by the user and document how.

GN-4644

### Setup
None

### How to test/reproduce
Mostly a review of API, but you can add classes to the different nodes and check that they get applied, also you can add styles to the existing classes

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


README.md
```diff
@@ -331,6 +331,59 @@ This addon uses CSS variables to customise the styling. You can override these v
...
+#### Override node classes
+
+You can also customize the look of your editor by either adding styles to the class of each node, or even changing that class to your custom one.
+The classes asigned to each node are the following:
+| Node | Css class |
+|---|---|
+| Heading | say-heading |
+| Paragraph | say-paragraph |
+| RepairedBlock | say-repaired-block |
+| ListItem | say-li |
```
elpoelma, line 343: As all others classnames are written in full, maybe we could also use `say-list-item` here? Really pedantic I know.

app/styles/ember-rdfa-editor/_c-content.scss
```diff
@@ -123,17 +135,17 @@ $say-editor-highlight-selected-color: var(--au-gray-300) !default;
...
   // Setup some margin rules
-  p,
-  ul,
-  ol,
-  table,
+  p.say-paragraph,
+  ul.say-bullet-list,
+  ol.say-ordered-list,
+  table.say-table,
   div,
```
elpoelma, line 142: Would it be possible to replace this `div` by e.g. `.say-block-rdfa`. This way that rule would be less general, and wouldn't target unwanted elements.

elpoelma (commented):
Really nice work, I do have some (pedantic) comments.

On the `classNames` attribute/setting:
- Maybe add it to the `SayNodeSpec` type
- Maybe we could extend its type to `string[] | (node: PNode) => string[]`. This allows for dynamic configurations.

lagartoverde:
> * string[] | (node: PNode) => string[]

pedantic review is just what I was looking for, this is an API change so it should be as good as possible :)
What would the function approach solve that we don't solve with the array?

lagartoverde:
And about adding it to the SayNodeSpec @abeforgit said something against it, don't remember what it was, I think something about not forcing the plugin creators to add it

elpoelma:
> > * string[] | (node: PNode) => string[]
> 
> pedantic review is just what I was looking for, this is an API change so it should be as good as possible :) What would the function approach solve that we don't solve with the array?

So for historical reference: it would allow node-specs to define dynamic classes which depend on e.g. attribute values.

elpoelma:
> And about adding it to the SayNodeSpec @abeforgit said something against it, don't remember what it was, I think something about not forcing the plugin creators to add it. But not sure why it can't be an optional attribute

Yeah I'd make it optional indeed.

elpoelma (changes requested):
Looks sane code-wise.
- There are still some failing tests
- I would maybe wait with merging this until we see the effects on the editor-plugins and GN

PR #1233, opened 2024-12-03
Fix document RDFa extractor to get subject attributes

piemonkey:
### Overview
Currently causes RB to crash when attempting to save new regulatory statement templates as they have properties but no subject.

N/A

### Setup
I tested linked to RB.

### How to test/reproduce
Create a new blank regulatory statement template in RB and then save it. It should save and not crash.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/rude-pillows-confess.md
+++ b/.changeset/rude-pillows-confess.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Fix crash when saving documents with subjects
--- a/addon/core/schema.ts
+++ b/addon/core/schema.ts
@@ -174,7 +174,9 @@ export function getRdfaAwareDocAttrs(
   }
 
   let properties: OutgoingTriple[] = [];
+  let subject: string | undefined = undefined;
   if (node.dataset['outgoingProps']) {
+    subject = node.dataset['subject'];
     properties = JSON.parse(
       node.dataset['outgoingProps'],
       jsonToTerm,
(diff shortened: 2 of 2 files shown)


elpoelma (changes requested):
This makes sense, but shouldn't we do the `node.dataset['subject']` check seperate from the `node.dataset['outgoingProps']`. This way we also support nodes with a subject but without outgoing props. Similar to the original `getRdfaAwareAttrs`.
Wouldn't it also be possible to check on the `node.dataset['rdfaNodeType']` attribute, similar to `getRdfaAwareAttrs`?

piemonkey:
> This makes sense, but shouldn't we do the `node.dataset['subject']` check seperate from the `node.dataset['outgoingProps']`. This way we also support nodes with a subject but without outgoing props. Similar to the original `getRdfaAwareAttrs`. Wouldn't it also be possible to check on the `node.dataset['rdfaNodeType']` attribute, similar to `getRdfaAwareAttrs`?

You're right about making the subject check separate. I've changed that now. For the rdfaNodeType, I don't think we currently set a node type for documents, so I don't think this works here. Did you have something else in mind?

abeforgit:
yeah we dont set the rdfaNodeType for documents
I tried to set that always to resource, but it caused other issues
so the doc node is being special cased in all the parsing logic atm

abeforgit (approved):
lgtm

PR #1235, opened 2024-12-11
Tweak style used by active ember-nodes to have rounded corners

piemonkey:
### Overview
Fixes snippet-placeholders. This style is normally overridden for most ember-nodes, e.g. by anything including a pill. Also affects mandatee tables and inserted snippets.

There will be a plugins PR to fix mandatee tables and snippets.

### Setup
Needs to be tested where these nodes are available, e.g. linked to the plugins

### How to test/reproduce
Insert a snippet-placeholder, select it, note that the blue border matches the border-radius of the node.

### Challenges/uncertainties
There's a chance that this causes visual issues somewhere I didn't spot.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/popular-starfishes-mix.md
+++ b/.changeset/popular-starfishes-mix.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Tweak style used by active ember nodes such as snippet placeholders to have rounded corners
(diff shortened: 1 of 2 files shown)


lagartoverde (changes requested):
Wouldn't it be better to just add this to the say-snippet class? :)

piemonkey:
> Wouldn't it be better to just add this to the say-snippet class? :)

To me the most common style for ember-nodes is having the corners rounded, so it makes sense to have this as the default. If anything, we should actually put this default into the CSS for an ember-node.

PR #1237, opened 2025-01-09
Ember 5

piemonkey:
### Overview
Adds support but doesn't increase the version used for testing.

PRs for plugins and GN incoming.

### Setup
Update to ember-source 5.12 to test in this repo.

### How to test/reproduce
Test that everything works! Best tested in GN.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/fresh-seals-glow.md
+++ b/.changeset/fresh-seals-glow.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add support for Ember 5.4+
--- a/addon/components/editor-container.hbs
+++ b/addon/components/editor-container.hbs
@@ -11,7 +11,9 @@
   {{yield to='top'}}
   <div class='say-container__main'>
     {{#if @loading}}
-      <AuLoader />
+      <AuLoader @hideMessage={{true}}>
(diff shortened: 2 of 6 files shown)


abeforgit (changes requested):
we should move to ember 5.12 for the testing app as well
Having the test-app on the latest version and then providing best-effort backwards-compatibility support is the strategy ember-appuniversum uses, and it seems to be the most feasible for our situation of a small team and and small userbase

piemonkey:
> we should move to ember 5.12 for the testing app as well Having the test-app on the latest version and then providing best-effort backwards-compatibility support is the strategy ember-appuniversum uses, and it seems to be the most feasible for our situation of a small team and and small userbase

Hmm. A key difference I can see there is that they have tests running against multiple versions of ember, so issues are more likely to be detected in advance. I guess it depends how long we intend to keep support for. Given that 4.12 has been 'end of life' for nearly 8 months now, maybe we can take the risk.

abeforgit:
I'll see if we can carve out some time for sam's help on getting a good ember-try setup
it's slow and often problematic, but it seems to remain the only way to test on multiple ember versions

abeforgit (approved):

PR #1248, opened 2025-02-03
allow inline images with config and checkPasteSize plugin

lagartoverde:
### Overview
Add configuration so the inline images can be enabled and added a plugin to check and avoid big paste sizes

GN-5423

### Setup
None

### How to test/reproduce
I enabled the base64 pasting in the index route, so you can try pasting the image there, if it's over 100kb (configurable by the host app) you will receive a console error (the host app will be able to hook a notification system there so the final user knows what is happening)

### Challenges/uncertainties
I added the configuration needed in the index route so the PR testing is easier, but is still unknown to me if I should make some configuration changes to any of the testing environments, because in my head I will delete that config before merging the PR

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/rude-eagles-help.md
+++ b/.changeset/rude-eagles-help.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Allow inline images with a new config and added checkPasteSize plugin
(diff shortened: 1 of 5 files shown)


elpoelma (approved):
Looks good!
- Performance is not an issue if I disable the prosemirror devtools. Resizing etc. works well.
- When the `checkPasteSize` is enabled, it correctly limits the size of the pasted content.

lagartoverde:
For future reviews, maybe 100kb is too low of a default? 1MB is maybe better

PR #1249, opened 2025-02-04
Tweaks to support article structure changes

piemonkey:
### Overview
Pull some command logic into utils that handle transactions, to allow them to be used outside of commands.

Also fix a bug in table inserting, that required the nodes to exist, which isn't true, e.g. in an embedded editor.

Part of https://binnenland.atlassian.net/browse/GN-4866
Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/535

### Setup
Not needed to test, but it makes sense to test this in the article structures PR in the plugins.

### How to test/reproduce
In the editor repo, these are used in the RDFa wrapping tools in the sidebar, which should work as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/grumpy-games-itch.md
+++ b/.changeset/grumpy-games-itch.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add util for wrapping including parents without using a command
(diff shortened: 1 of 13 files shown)


abeforgit (approved):
I added a test case for the bug you found in #1251 
all looks good

PR #1250, opened 2025-02-10
Fix weird placeholder margin

lagartoverde:
### Overview
There's a style that sets the right margin of mark-highlight-manual (the old class for placeholder) so it seems like after that node there are 2 spaces instead of 1, I fixed this using our new class for placeholder to minimize the impact on other places, as this class is used a lot in our apps

GN-5445 (maybe?)

### Setup
None

### How to test/reproduce
Insert a placeholder and notice the margin right is reduced to a normal amount

### Challenges/uncertainties
Maybe is just better to remove it from the class

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/fast-bats-jam.md
+++ b/.changeset/fast-bats-jam.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Reduce margin right property of placeholders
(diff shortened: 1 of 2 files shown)


elpoelma:
We should probably look at removing the `mark-highlight-manual` class altogether, and include its relevant styles on the `say-placeholder` class.

lagartoverde:
Yeah but that would be a breaking release, and I wanted to do a patch :P.
Maybe something to include in a future issue

elpoelma:
> Yeah but that would be a breaking release, and I wanted to do a patch :P. Maybe something to include in a future issue

That's fair, you'll need to target and older release then however, as the current slated release _will_ be breaking.

elpoelma (approved):

PR #1255, opened 2025-02-18
Feat/notification system

lagartoverde:
### Overview
Developed a notification system for the editor and made the pasteSize plugin deliver notifications to this system. 
The user will be able to add 2 more arguments to the editor compontent, `notificationCallback` which will be called when there's a notification to show with a notification argument:
```
export interface Notification {
  title?: string;
  message?: string;
  options: {
    type?: 'info' | 'success' | 'warning' | 'error'; // Default depends on the used display method
    icon?: string; // Any valid Appuniversum icon name, default depends on the used display method
    timeOut?: number; // delay in milliseconds after which the toast auto-closes
    closable?: boolean; // Can the toast be closed by users, defaults to `true`
  };
}
```
This by default will show the notification in a appuniversum toaster.
And `notificationToaster` which will determine if the AuToaster component gets added to the app, defaults to true.

GN-5448

### Setup
None

### How to test/reproduce
Paste an image that's too big (more than 1MB on the editor) and notice a notification pops out, you can play around with the arguments, changing the notification for a console log with a custom callback or noticing that if you don't show the toaster the notification doesn't appear

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


addon/plugins/image/plugins/checkPasteSize.ts
```diff
@@ -27,7 +28,13 @@ export function checkPasteSize({
           if (onLimitReached) {
             onLimitReached();
           } else {
-            console.error('Paste size is bigger than expected');
+            // Show a notification via the notification plugin
+            notificationPluginKey.getState(view.state).notificationCallback({
+              title: 'Paste size limit reached',
```
abeforgit, line 33: This hardcoded message isn't translated, and it also doesn't explain to the user what they can do about it

abeforgit (changes requested):

PR #1258, opened 2025-02-19
fix(tooling): adjust `package.json` `imports` field to not include `declarations` folder

elpoelma:
### Overview
This PR slightly adjusts the `package.json` `imports` field to not include the `declarations` folder.
As the `imports` field is only used internally while developing, it does not seem to be necessary to include the `declarations` output folder.
This also fixes an issue with auto-imports not importing with the correct `.ts` extension.

### Setup
- I tested this in vscode with the glint extension (https://marketplace.visualstudio.com/items?itemName=typed-ember.glint-vscode)

### How to test/reproduce
- Try to import from an internal module in the editor package
- Ensure the auto-imports it proposes have `.ts` extensions

### Challenges/uncertainties
Not sure if I'm missing something about the `declarations` folder being needed here?

### DIFF
--- a/packages/ember-rdfa-editor/package.json
+++ b/packages/ember-rdfa-editor/package.json
@@ -168,14 +168,8 @@
   "imports": {
     "#root": null,
     "#root/index.ts": null,
-    "#root/*.ts": {
-      "types": "./declarations/*.d.ts",
-      "default": "./src/*.ts"
-    },
-    "#root/*": {
-      "types": "./declarations/*.d.ts",
-      "default": "./src/*"
-    }
+    "#root/*.ts": "./src/*.ts",
+    "#root/*": "./src/*"
   },
   "typesVersions": {
     "*": {


piemonkey (commented):
For me, fixes the file extensions for `#root` imports to be `.ts` instead of `.js` but for relative imports I still get no extension. It's very possible that's on my editor config side though.

I'm not going to hit 'approve' as I can't say whether the declarations were needed for something else...

elpoelma:
> For me, fixes the file extensions for `#root` imports to be `.ts` instead of `.js` but for relative imports I still get no extension. It's very possible that's on my editor config side though.
> 
> I'm not going to hit 'approve' as I can't say whether the declarations were needed for something else...

Ok, so for some reason this change completely breaks the type-checks in the `test-app`, which I still don't fully understand why. I guess this is also not super urgent...

abeforgit:
The reason it breaks the types is that while yes, for a consuming package importing a module, only the `exports` field matters, for imports one level deeper, the `imports` field is still driving the resolution. 

And we need to make sure the declaration files are importing from the declaration files, and not from the ts sources. This matters because private fields are not included in declaration files, so classes with private fields are a different type in the declaration compared to the source code. 

ref: https://github.com/microsoft/TypeScript/issues/60230

PR #1262, opened 2025-02-20
Feat/notification system v2

lagartoverde:
Copy of https://github.com/lblod/ember-rdfa-editor/pull/1255 with the v2 addon integration
### Overview
Developed a notification system for the editor and made the pasteSize plugin deliver notifications to this system. 
The user will be able to add 2 more arguments to the editor compontent, `notificationCallback` which will be called when there's a notification to show with a notification argument:
```
export interface Notification {
  title?: string;
  message?: string;
  options: {
    type?: 'info' | 'success' | 'warning' | 'error'; // Default depends on the used display method
    icon?: string; // Any valid Appuniversum icon name, default depends on the used display method
    timeOut?: number; // delay in milliseconds after which the toast auto-closes
    closable?: boolean; // Can the toast be closed by users, defaults to `true`
  };
}
```
This by default will show the notification in a appuniversum toaster.
And `notificationToaster` which will determine if the AuToaster component gets added to the app, defaults to true.

GN-5448

### Setup
None

### How to test/reproduce
Paste an image that's too big (more than 1MB on the editor) and notice a notification pops out, you can play around with the arguments, changing the notification for a console log with a custom callback or noticing that if you don't show the toaster the notification doesn't appear

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (changes requested):
The notification system works well, but I'm now getting a `node is undefined` error when trying to paste an image. This also happens on master.

packages/ember-rdfa-editor/src/plugins/notification/index.ts
```diff
@@ -0,0 +1,36 @@
+import type IntlService from 'ember-intl/services/intl';
+import { Plugin, PluginKey } from 'prosemirror-state';
+
+export interface Notification {
+  title?: string;
+  message?: string;
+  options: {
+    type?: 'info' | 'success' | 'warning' | 'error'; // Default depends on the used display method
+    icon?: string; // Any valid Appuniversum icon name, default depends on the used display method
```
piemonkey, line 9: Does using a string for icons work with embeddable?

piemonkey (commented):
It's not really related to the notification system, but the image paste logic seems to not work well. The only way I seem to be able to paste an image into the editor is by right-click -> copy image, in a browser, but this pastes into the document as an image node (i.e. referencing an external url for the image). If I paste a small enough image, I see nothing, if I paste a too large image, I get the notification...

packages/ember-rdfa-editor/src/components/editor.ts
```diff
@@ -100,12 +127,42 @@ export default class RdfaEditor extends Component<RdfaEditorArgs> {
       );
     }
 
+    const notificationCallback =
+      this.args.notificationCallback ??
```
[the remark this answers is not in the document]
lagartoverde: so the 2 arguments serve a different purpose, will see how can I clarify them.
Basically the notificationCallback overrides all the behaviour in the plugin, and lets you do whatever you want with the notification object.
But the toaster argument is useful for cases like GN, where we want to keep the behaviour but we already have a toaster present in order to show our own notifications, so we just need to avoid the editor adding another one

packages/ember-rdfa-editor/src/plugins/notification/index.ts
lagartoverde: it should work, but we are not using it

lagartoverde: yet

packages/ember-rdfa-editor/src/components/editor.ts
```diff
@@ -56,6 +62,22 @@ export interface RdfaEditorArgs {
  */
 export default class RdfaEditor extends Component<RdfaEditorArgs> {
   @tracked controller: SayController | null = null;
+  @service declare toaster: {
+    notify: (
```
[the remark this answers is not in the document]
piemonkey: My bad, they don't actually publish types (as far as I can see) or this.

packages/ember-rdfa-editor/src/plugins/notification/index.ts
piemonkey: Even though we had to move to passing Icon components?

piemonkey:
It seems to mostly work well, but a couple of issues I spotted:

- On Chromium it seems to block pasting an image that I can paste on FF dev edition (a 550KB image).
- Drag-dropping an image seems to circumvent the size check, so I can e.g. drag a 1.3MB image in without error.

elpoelma (commented):
Seems to work well overall, just added a small naming comment.

When trying to drag-and-drop a large image, a notification is now correctly shown.
In chromium browsers, paste works as expected. Drag-and-drop sometimes works, but also sometimes simply opens the image in a new tab (in the chromium image viewer). This doesn't seems like a big issue to me however.

packages/ember-rdfa-editor/src/plugins/notification/index.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface NotificationOptions {
+  type?: 'info' | 'success' | 'warning' | 'error'; // Default depends on the used display method
+  icon?: TemplateOnlyComponent; // Any valid Appuniversum icon imported, default doesn't work on embeddable
+  timeOut?: number; // delay in milliseconds after which the toast auto-closes
+  closable?: boolean; // Can the toast be closed by users, defaults to `true`
+}
+
+export const notificationPluginKey = new PluginKey('NOTIFICATION');
+
+interface notificationOptions {
```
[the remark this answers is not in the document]
abeforgit: https://typescript-eslint.io/rules/naming-convention/ christ almighty that's probably why lol, what a beast of a lint

elpoelma (approved):

PR #1263, opened 2025-02-24
fix(html-input-parser): return correct htmlelement

elpoelma:
### Overview
This PR ensures that the `HTMLInputParser.prepareHTML` method returns an `HTMLElement` object instead of a `Document` object.
A change was introduced in the `v2-addon` branch related to the usage of return values, this caused a bug in the `prepareHTML` method. See https://github.com/lblod/ember-rdfa-editor/commit/24bdc2046f3530b63beee08046823b6cb840152d for more details.

### How to test/reproduce
- Start the dummy app
- Paste any html (e.g. an inline image)
- Ensure this pasting works as expected
- Notice there's no longer an error in the console

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/tricky-glasses-fail.md
+++ b/.changeset/tricky-glasses-fail.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+adjust `HTMLInputParser.prepareHTML` to return `HTMLElement` instead of `Document`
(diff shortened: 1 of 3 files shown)


lagartoverde (approved):
fixes the bug :)

PR #1264, opened 2025-02-26
Fix hot reloading in the test app

abeforgit:
### Overview

Sets up https://github.com/NullVoxPopuli/pnpm-sync-dependencies-meta-injected
We have to use `injected = true` for the addon package, because we have
a peerdep (ember-appuniversum) which _also_ has to be a devdep because we need it at build-time (for the sass).

In this scenario, because of the symlinking happening in the monorepo setup, the app would incorrectly
resolve the peerdep from the addon's node_modules, instead of the app's, leading
to a peerdep clash (and pnpm refusing to run the build scripts)

That's what injected is for, it packages the dep as if it would be properly
installed. But that interferes with the automatic reloading. So nullvox made
a tool that fixes _that_

I originally didn't set this up as the autoreload seemed to be working, and
I was also having issues with the setup. Luckily, at the same time the
autoreload stopped working, the issues with the setup also went away, so now we
can use this.

### How to test/reproduce

`pnpm start` from the root, and make sure that edits to the addon's sources
correctly reload the test-app and use the new code.


elpoelma (approved):
Don't really understand why the hot reload stopped working, but this does solve it :)

elpoelma:
It's weird as I now get errors while trying to run the test-app with this setup, but without this setup the hot-reload seems to work again...

abeforgit:
yeah I was struggling yesterday as well, don't know what's going on anymore

PR #1273, opened 2025-03-07
Support external triples and backlinks for literalNodes

abeforgit:
### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations

### DIFF
--- a/.changeset/fifty-ghosts-drop.md
+++ b/.changeset/fifty-ghosts-drop.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Support external triples and loose backlinks on literal nodes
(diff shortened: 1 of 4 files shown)


elpoelma:
So just to be sure, there is no UI support for this right?

abeforgit:
for the external triples there is, but not for the backlinks, no

elpoelma:
> for the external triples there is, but not for the backlinks, no

ok, it's because I can't seem to add external triples to literal nodes through the UI...

abeforgit:
> > for the external triples there is, but not for the backlinks, no
> 
> ok, it's because I can't seem to add external triples to literal nodes through the UI...

ah yes, cause they're backlinks now. I'll add a way to make them

PR #1281, opened 2025-04-01
Support for adding backlinks (and other improvements to the rdfa-tools)

elpoelma:
### Overview
This PR contains a bunch of changes/small improvements to the rdfa-tooling components, with the main one being the support for adding/removing/editing backlinks.

So specifically this PR includes:
- Converting the rdfa-tooling components to the `gts` format
- Merging the `relationship-editor` and `property-editor` into a single `property-editor` component
  * Note: this component no longer manages backlinks, this is now done in a seperate component
- Adjusting the folder structure of the rdfa-tooling components to ensure consistency
- Introduction of a `backlink-editor` component, which can be used to add/edit/delete backlinks.
- Introduction of an `addBacklinkToNode` transaction-monad.

### How to test/reproduce
- Start the test-app
- Open the `editable-node` page
**To test the new `property-editor`**
- Ensure you can add properties with objects of any type + edit and remove them
- If having multiple nodes with the same subject, the changes to properties should be applied to all these nodes
- Imported resources should still work as expected with the `document` node

**To test the new `backlink-editor`**
- Insert a literal or resource node
- Ensure you may now add/manage/delete backlinks through the new component
- For literal nodes, it is also possible to add multiple backlinks
- When adding a backlink, you may also provide a custom subject

### Challenges/uncertainties
- Hopefully did not break anything related to imported resources
- When adding a property, you may now select all object types. This may not make complete sense/is not fully supported by imported resources.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/form.gts
```diff
@@ -123,6 +134,7 @@ export default class OutgoingTripleFormComponent extends Component<Sig> {
   }
 
   get literals(): string[] {
+    console.log('Get literals!');
```
piemonkey, line 137: Console.log

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/form.gts
```diff
@@ -138,6 +150,7 @@ export default class OutgoingTripleFormComponent extends Component<Sig> {
         result.push(rdfaId);
       }
     });
+    console.log('Literals: ', result);
```
piemonkey, line 153: Console.log

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/index.gts
```diff
@@ -163,30 +103,50 @@ export default class RdfaRelationshipEditor extends Component<Args> {
...
+  get documentImportedResources(): string[] | false {
+    return (
+      this.type === 'document' &&
+      !!this.controller?.schema.nodes['doc']?.spec.attrs?.[
+        IMPORTED_RESOURCES_ATTR
+      ] &&
+      ((this.node.attrs[IMPORTED_RESOURCES_ATTR] as string[]) || [])
+    );
+  }
+  get allImportedResources(): string[] {
```
piemonkey, line 133: The return of `string[] | false` was to signal when there is no imported resource attr set, which indicates that there is no support for imported resources in this document. As a result, the subject field is always shown (and is required).

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/index.gts
```diff
@@ -236,84 +196,33 @@ export default class RdfaRelationshipEditor extends Component<Args> {
...
-  addBacklink = (_backlink: IncomingTriple) => {
-    throw new NotImplementedError();
-  };
-
   addProperty = (property: OutgoingTriple, subject?: string) => {
+    console.log('Add prop to subject: ', subject);
     // This function can only be called when the selected node defines a resource or the selected
     // node is a document that imports resources (e.g. a snippet)
     const resource = this.currentResource || subject;
+    console.log('Resource: ', resource);
```
piemonkey, line 225: Console.log

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/index.gts
```diff
@@ -236,84 +196,33 @@ export default class RdfaRelationshipEditor extends Component<Args> {
...
-  cancel = () => {
-    this.status = undefined;
-  };
-
-  addBacklink = (_backlink: IncomingTriple) => {
-    throw new NotImplementedError();
-  };
-
   addProperty = (property: OutgoingTriple, subject?: string) => {
+    console.log('Add prop to subject: ', subject);
```
piemonkey, line 221: Console.log

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/wrapping-utils/index.gts
```diff
@@ -5,24 +5,27 @@ import {
   wrapLiteral,
   wrapResource,
 } from '#root/commands/_private/rdfa-commands/index.ts';
-import WrappingModal from './modal.ts';
+import WrappingModal from './modal.gts';
 import type { ResolvedPNode } from '#root/utils/_private/types.ts';
 import { wrapInlineLiteral } from '#root/commands/_private/rdfa-commands/wrap-inline-literal.ts';
 import { PlusIcon } from '@appuniversum/ember-appuniversum/components/icons/plus';
 import type SayController from '#root/core/say-controller.ts';
+import AuContent from '@appuniversum/ember-appuniversum/components/au-content.js';
```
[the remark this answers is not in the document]
elpoelma: should be fixed now

packages/ember-rdfa-editor/src/components/_private/rdfa-editor/property-editor/index.gts
elpoelma: Should be fixed now!

piemonkey (commented):
Looks good. There's one thing that I spotted while testing, though it doesn't actually come from these changes. The `unwrap()` in `property-editor/form` (L218) means that if a node can't be found for the RdfaId, the app crashes and needs to be reloaded. I'm not sure of a good way to handle this, but we definitely should do it more gracefully...

elpoelma:
> Looks good. There's one thing that I spotted while testing, though it doesn't actually come from these changes. The `unwrap()` in `property-editor/form` (L218) means that if a node can't be found for the RdfaId, the app crashes and needs to be reloaded. I'm not sure of a good way to handle this, but we definitely should do it more gracefully...

Good point, does this cause a bug in practice?

piemonkey:
> > Looks good. There's one thing that I spotted while testing, though it doesn't actually come from these changes. The `unwrap()` in `property-editor/form` (L218) means that if a node can't be found for the RdfaId, the app crashes and needs to be reloaded. I'm not sure of a good way to handle this, but we definitely should do it more gracefully...
> 
> Good point, does this cause a bug in practice?

I saw this while looking at some templates on dev.RB. These templates had some resources with relationships to rdfa ids which don't seem to exist, so this indeed crashes when trying to edit these.

elpoelma:
> > > Looks good. There's one thing that I spotted while testing, though it doesn't actually come from these changes. The `unwrap()` in `property-editor/form` (L218) means that if a node can't be found for the RdfaId, the app crashes and needs to be reloaded. I'm not sure of a good way to handle this, but we definitely should do it more gracefully...
> > 
> > 
> > Good point, does this cause a bug in practice?
> 
> I saw this while looking at some templates on dev.RB. These templates had some resources with relationships to rdfa ids which don't seem to exist, so this indeed crashes when trying to edit these.

Huh ok, I didn't think it was possible to get in such a state...

abeforgit:
> Looks good. There's one thing that I spotted while testing, though it doesn't actually come from these changes. The `unwrap()` in `property-editor/form` (L218) means that if a node can't be found for the RdfaId, the app crashes and needs to be reloaded. I'm not sure of a good way to handle this, but we definitely should do it more gracefully...

made ticket so we don't lose sight of this: https://binnenland.atlassian.net/browse/GN-5559

abeforgit (changes requested):
- "add property" on the document is disabled now, it wasn't before


- Adding a backlink on the document doesn't work. 
Most elegant would be if it did, but I can see how that might be tricky to implement. If so, disabling the button is enough for now, we don't have a compelling usecase for it atm.

- adding a backlink on an inline resourcenode to itself doesn't work (throws an error in the console)
Doing this in the other direction does work, so it would be nice if it did. 
This only seems to occur on inline nodes. If it turns out to be super difficult cause of the nested PMs, disable the button and make a jira ticket.

- [only if easy]
would be nice if it would give a visual indication if a backlink is to a non-existent subject:
image

elpoelma:
> * "add property" on the document is disabled now, it wasn't before
> 
> * Adding a backlink on the document doesn't work.
>   Most elegant would be if it did, but I can see how that might be tricky to implement. If so, disabling the button is enough for now, we don't have a compelling usecase for it atm.

Most (if not all) document nodes we are interacting with are not actually resource or literal nodes (they don't have an `rdfaNodeType`). So I don't think it really makes sense to enable the 'Add property' or 'Add backlink' buttons for these.
If `hasResourceImports` is enabled, the 'Add property' button is enabled on document nodes, otherwise it is only enabled if the document node is actually also a resource node. The 'Add backlink' button is only enabled on document nodes if they have either 'literal' or 'resource' as their `rdfaNodeType`.

> * adding a backlink on an inline resourcenode to itself doesn't work (throws an error in the console)
>       Doing this in the other direction does work, so it would be nice if it did.
>       This only seems to occur on inline nodes. If it turns out to be super difficult cause of the nested PMs, disable the button and make a jira ticket.
We should indeed make this work, I'll look into it.




> * [only if easy]
>       would be nice if it would give a visual indication if a backlink is to a non-existent subject:

This is not very hard to do (we only need to check if the subject occurs in the document), but might be/is computationally expensive. It would mean that we do this check/search after every transaction, thus every keystroke, which might not be ideal. Of course, it would only occur when the RDFa dev-tools are enabled.

abeforgit:
> Most (if not all) document nodes we are interacting with are not actually resource or literal nodes (they don't have an `rdfaNodeType`). So I don't think it really makes sense to enable the 'Add property' or 'Add backlink' buttons for these. If `hasResourceImports` is enabled, the 'Add property' button is enabled on document nodes, otherwise it is only enabled if the document node is actually also a resource node. The 'Add backlink' button is only enabled on document nodes if they have either 'literal' or 'resource' as their `rdfaNodeType`.

I see, yes. Not sure how I ended up with a state where my doc node was a resource node :thinking: I can't reproduce it atm
> This is not very hard to do (we only need to check if the subject occurs in the document), but might be/is computationally expensive. It would mean that we do this check/search after every transaction, thus every keystroke, which might not be ideal. Of course, it would only occur when the RDFa dev-tools are enabled.

ok, let's not do it then

abeforgit (approved):
good to go for me!

PR #1284, opened 2025-04-03
Improvements to imported resources RDFA tools

piemonkey:
### Overview
Re-uses some of the new UI elements from [the add backlinks PR](https://github.com/lblod/ember-rdfa-editor/pull/1281) and uses them to make imported resources more intuitive. The old behaviour is still there, but we could disable it if we want, since the more intuitive version is more likely to be used.

There seems to be the issue seen elsewhere of needing 2 clicks to use sidebar buttons. I wasn't sure what the fix for this was, so didn't look into it yet, assuming that there's a quick fix.

This doesn't handle literals correctly, but that was already an issue with imported resources.

Jira ticket: https://binnenland.atlassian.net/browse/CIT-26

### Setup
To test in the test app, you need to set `hasResourceImports: true` in the `docWithConfig` options for the editable-nodes page. Otherwise, can be tested in RB.

### How to test/reproduce
Select the document node and see that the 'Imported Resources' section in the rdfa tools appears. The process for adding, updating and removing both imported resource URIs and properties for these, should be intuitive and all work as expected.

### Challenges/uncertainties
A bit of mess with rebasing, but worth it to get the ready-made UI elements.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
@piemonkey wrt the double clicks: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/463
prob not the exact same cause here but maybe similar (modal state flickering on and off)?

piemonkey:
@abeforgit The double-click problem is now fixed along with a couple of other things I noticed that I had missed, such as 'properly'* supporting literals for imported resource relationships.

*'properly' as they still disappear from the tooling on refresh...

elpoelma (commented):
Overall looks good, adding/removing/editing properties to imported resources seems to work as expected.

- Properties targeting literal nodes indeed disappear after a reload, unless there is also a property present which targets a resource node.
- After properties targeting literal nodes have disappeared, adding a property targeting a resource node and reloading the document cause the properties targeting the literal nodes to appear again...

abeforgit (approved):
much nicer than it was, good work
if you see a solution for elena's comment, have a look, but otherwise like you said this never worked before either, so it doesn't block the merge imo

piemonkey:
> much nicer than it was, good work if you see a solution for elena's comment, have a look, but otherwise like you said this never worked before either, so it doesn't block the merge imo

I've figured out that by putting in an extra node in the invisible RDFa for the document, this allows the outgoingProps to be put onto it as a data attribute, and therefore, to be parsed again. This doesn't seem to have any impact on the RDFa that is generated, but could do with a second pair of eyes. Changes can be seen here: https://github.com/lblod/ember-rdfa-editor/pull/1284/commits/69eccbca504be536eaa2f9245503c1edca70977c

abeforgit (approved):

abeforgit:
ticket to remove ember-changeset: https://binnenland.atlassian.net/browse/GN-5564

PR #1286, opened 2025-04-09
CIT-31: rdfa tools tab behaviour/submit shortcuts

elpoelma:
### Overview
This PR improves the behaviour of the forms included in the rdfa-tool modals:
- Addition of `Ctrl+Enter`/`Cmd+Enter` handlers to submit creation/edit forms inside modals of the rdfa-tool components.
- Set up modifiers to initially focus first inputs of the forms (in combination with `focus-trap`)

[CIT-31](https://binnenland.atlassian.net/browse/CIT-31)

### How to test/reproduce
- Start the test-app and open the 'editable node' page
- Open any of the rdfa-tool modals containing a form
- The first focuseable input of the form should be focused by default
- Ensure you can traverse between the different form fields. There should be a focus trap on the modal
- Pressing `Ctrl+Enter`/`Cmd+Enter` should submit the form as expected. Note: 

### Challenges/uncertainties
When a power-select is focused, the `Ctrl+Enter` key is not propagated correctly. I am still looking into this.

I opted for modifiers (in combination with `focus-trap.initialFocus`) to implement the autofocussing of the first form element. Another option would be to use the id's of the elements to focus initially (instead of the elements themselves).
Another option (maybe to be used in the future) is to use the native `dialog` element, in combination with the `autofocus` attribute, check-out https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#usage_notes.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/legal-fans-push.md
+++ b/.changeset/legal-fans-push.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Rdfa-tool components: add `Ctrl+Enter`/`Cmd+Enter` handler to submit creation/edit forms inside modals.
--- a/.changeset/slow-rules-teach.md
+++ b/.changeset/slow-rules-teach.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+RDFa tool forms/modals: initially focus relevant form inputs
--- a/packages/ember-rdfa-editor/src/components/_private/rdfa-editor/backlink-editor/form.gts
+++ b/packages/ember-rdfa-editor/src/components/_private/rdfa-editor/backlink-editor/form.gts
@@ -13,10 +13,12 @@ import { uniqueId } from '@ember/helper';
 import { action } from '@ember/object';
 import type SayController from '#root/core/say-controller.ts';
 import { getSubjects } from '#root/plugins/rdfa-info/index.ts';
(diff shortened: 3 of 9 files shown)


lagartoverde (approved):
Code looks good, thanks for the help, but hte shortcut of ctrl+enter is not working for me, I think because the select is not bubbling the events

elpoelma:
> Code looks good, thanks for the help, but hte shortcut of ctrl+enter is not working for me, I think because the select is not bubbling the events

Yes indeed, that's what I noticed too... I'm looking into it.

elpoelma:
> > Code looks good, thanks for the help, but hte shortcut of ctrl+enter is not working for me, I think because the select is not bubbling the events
> 
> Yes indeed, that's what I noticed too... I'm looking into it.
> 
> The main problem is that this is hard to circumvent: https://github.com/cibernox/ember-power-select/blob/3f948b90c4bb1fe14bc7dc025f14fc48efc90fb6/ember-power-select/src/components/power-select.ts#L462-L464

I do see one option: to extend the `onKeyDown` handler to also submit the form when ctrl+enter is pressed.

elpoelma:
@lagartoverde I tested the approach of also using the keydownhandler for the power-select in the backlink-editor, does this approach make sense to you? I'll adjust the other forms afterwards.

lagartoverde (approved):

PR #1287, opened 2025-04-10
Update the compatibility information in the readme

Windvis:
v3.28 support was dropped already so the readme was a bit out of date.

It now matches the [`@embroider/addon-blueprint` version](https://github.com/embroider-build/addon-blueprint/blob/be089ee2ee9c7a5c72540cfd004226d3bff0ca4d/files/README.md?plain=1#L7-L8).


elpoelma (approved):
Thanks for the update! Our readmes are indeed not always very up-to-date :grimacing:

PR #1288, opened 2025-04-14
Move test-app to vite

elpoelma:
### Overview
Moves the `test-app` to the vite setup.

### How to test/reproduce
- Ensure both the dev builds and production builds of the test-app work as expected
- The tests should run in headless and browser mode
- The E2E tests should run as expected
- Sourcemaps should be perfect, and should also work in your code editor :smile:

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
I feel like there are still too many caveats to the vite build (including the slower speed and frequent crashes), will convert this to draft for now...

abeforgit:
@elpoelma do you have any more info on the crashes? is it mostly when hot-reloading? is it related to type-errors crashing the build?

abeforgit:
@elpoelma another idea is to move to the newest addon blueprint, which is no longer a monorepo.
I'd still keep the monorepo structure, cause I would like to move some packages into this repo, but for the test-app it may be better?

my preference is still to get the separate test-app to work, but if we cant get good DX out of it (yet), it may be worth (temporarily) switching

elpoelma:
> @elpoelma do you have any more info on the crashes? is it mostly when hot-reloading? is it related to type-errors crashing the build?

So a few things:
- The package build sometimes crashes, but this is of course not due to vite, we'll have to fix it in a seperate PR
- Sometimes the vite build would simply stay on a blank page, only showing the application after a manual reload
- It often takes quite a while for the application to show, during the initial build (you do see feedback/progress in the browser console though)

PR #1290, opened 2025-04-16
feat: lighter styling for block rdfa nodes

abeforgit:

related to https://binnenland.atlassian.net/browse/CIT-30 but not really blocking

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


abeforgit:
No idea wth is happening in the compat tests, I only changed some styles :sob:

elpoelma (commented):
The smaller margins are indeed a big improvement, I'm just wondering if a tiny margin on the left would not be better to distinguish the different blocks.

On the change of font thickness and size: it indeed feels lighter, but is also a bit less readable IMO.

On using the dotted border, I was maybe thinking of distinguishing 'loose' and 'connected' nodes that way:
- Connected nodes would have a solid border
- Loose nodes would have a dotted/dashed border

piemonkey (commented):
It probably depends on the exact browser configuration, but for me the dotted borders on Firefox are thin enough that they look broken. I think it looks better with all solid borders.
image

piemonkey (approved):
I moved back to solid inactive borders and changed back the font-size for the labels (though kept the font-weight lower).

PR #1292, opened 2025-04-17
CIT-32: linking UI POC

elpoelma:
### Overview
This (draft) PR contains an initial implementation of a user-friendly node-linking UI.

What exactly does this PR entail:
- A sidebar button + modal with form which allows you to link/add a backlink to the currently selected node
- The form does not contain free input fields, but rather selectors which users can use to select a predicate and subject
- The component can be configured with:
  * a function to generate a list of predicate ('relatie-type') options
  * a function to generate a list of subject ('onderwerp') options
- Integration of ember-headless-form and yup validation

Some important points:
- Users do not want to see technical URI's but rather a label/title + description for each of the options
- Users should first select the predicate ('relatie-type') before selecting a subject ('onderwerp')

The component may be configured with:
a `predicateOptionGenerator`, which conforms to the following types:
```ts
type PredicateOptionGeneratorArgs = {
  selectedObject?: SayTerm;
  searchString?: string;
};
type PredicateOptionGenerator = (
  args?: PredicateOptionGeneratorArgs,
) => TermOptionGeneratorResult<SayNamedNode>
```

and a `subjectOptionGenerator`, which conforms to the following types:
```ts
type SubjectOptionGeneratorArgs = {
  selectedObject?: SayTerm;
  selectedPredicate?: SayTerm;
  searchString?: string;
};

type SubjectOptionGenerator = (
  args?: SubjectOptionGeneratorArgs,
) => TermOptionGeneratorResult<ResourceNodeTerm>;
```

The result of these generator functions can either be a list of options, or a promise of a list of options.
This way the consumer of this component can configure how the options are generated, e.g.:
- Retrieve options based on content in the document
- Retrieve options from a SPARQL endpoint or other external datasource

Possible future work:
- Move the controls to add relationships to a node to the node itself, similar to the controls of snippet nodes or link nodes.
- Distinguish 'connected' nodes from 'loose' nodes by a different border style
- I've thought about adding a border animation to the selected node when adding/editing a relationship, but this seems a bit too distracting to me.
- We might want to merge this component with the 'backlink-editor', and provide an opt-in dev-mode switch.

[CIT-32](https://binnenland.atlassian.net/browse/CIT-32)

### How to test/reproduce
- Start the test-app
- Open the editable-node page
- Click on the 'Link node' button
- You get presented with a UI in which you may select a predicate from a list, and then select a subject
- When not providing a predicate and/or subject, a warning message is shown
- The 'submit' button should add the correct backlink

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


piemonkey (commented):
Nice PoC. Good to see a non-ember-changeset form validation lib in use too :)

Some thoughts about the feature:
- In terms of differentiating 'connected' and 'loose' nodes, this is related to something @abeforgit and I discussed yesterday, about using imported-resources (or something like it) to indicate that a snippet (or something like it) has something that still needs to be linked in the document.
- I like the idea of putting controls on the node itself as I'm not a fan of the sidebar overuse we have at the moment.
- One bigger question is, should this be only backlinks? It strikes me that we might want to be able to insert some RDFa which needs an *outward* link to something in the document. To the user, they just need to be related, the direction of that relationship might not be important to them. For this, we would need to know, for a given selected node, what backlinks *or* properties are possible. This is in fact easier for properties if the node has a `rdf:type` as we can know what relationships that type has. I get that isn't what's needed right now, but maybe it should be included in the API somehow, e.g. by including the direction in the `SubjectOptionGenerator` return.

PR #1297, opened 2025-04-24
Add RDFA visualiser

piemonkey:
### Overview
Displays a tree of resources found in the current document.

The display of each element is configurable, by passing the required generator to the component. The possible keys are the different `SayTermTypes` as well as 'predicate' for predicates. There are 2 demo functions in the editable nodes test page. If a termtype doesn't have a generator configured, it falls back to using the component used in the property viewer.

The generator function for display returns an array, each element representing a part of the display for that entity, from a choice of different elements, such as plaintext or an `AuPill`.

The requirements in the ticket, such as no URIs shown to users have not been met. This is because this is because this configuration should sit in the plugins repo, as it would need to know details such as a list of prefixes and how to handle them for display. The example functions go some of the way to this, but don't cover many cases.

Jira: https://binnenland.atlassian.net/browse/CIT-34

### Setup
Can be tested as is, or passing the option `keepOpen` to the visualizer component to test the impact of this mode on performance.

### How to test/reproduce
The structure in the sidebar should stay up to date with the document, including all relationships. Notes:
- The top level should be the tops of any 'trees' in the RDFa, including those that are only tree-like at this highest level.
- If there are non-tree structures, i.e. those containing loops, with no 'top', then the resource defined at the highest level of the document should be shown at the top level.
- Imported resources should always be at the top level.
- The way elements are displayed is configurable

### Challenges/uncertainties
Debouncing... Also British spelling (visualiser) vs American (visualizer). I defaulted to British but tried to keep US in the interface as I guess that's the default... I may have missed something though.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit:
Works very well, and even if no further development occurs will be very valuable as a demo/debugging tool

follow-up work has been defined here:
https://binnenland.atlassian.net/browse/CIT-43
https://binnenland.atlassian.net/browse/CIT-33

PR #1302, opened 2025-04-28
Example: integrate LOV api

elpoelma:
### Overview
This PR contains an example on how to integrate the LOV (https://lov.linkeddata.es/dataset/lov) API into the node-linker-poc component.
It is quite fast and I can already see the advantages of integrating it.

Using the LOV api we can also query for labels and descriptions of different terms, so that might also be interesting.

### How to test/reproduce
- Start the test-app
- Open the editable-node page and click on the 'Add relationship' button
- Notice that when typing in the 'relationship-type' field, it now autocompletes


abeforgit:
very cool, but maybe we should do https://binnenland.atlassian.net/browse/CIT-41 first?
then it can be a default fallback I think

elpoelma:
> very cool, but maybe we should do https://binnenland.atlassian.net/browse/CIT-41 first? then it can be a default fallback I think

Agreed, I just stumbled upon the API and was excited to try it out :)

elpoelma:
Replaced by https://github.com/lblod/ember-rdfa-editor/pull/1317

PR #1309, opened 2025-05-12
intl(test-app): dynamically import intl files

elpoelma:
### Overview
This PR adjusts the setup of the test-app to dynamically load the translation files included in the editor package. This ensures that live-reloading the translation files works with vite.
What does this PR specifically include?
- Addition of `@modyfi/vite-plugin-yaml`, which allows us to import yaml files
- Dynamic imports of the editor translation yaml files

### How to test/reproduce
- Start the test-app
- Make a change to one of the translation files of the editor
- Ensure that the app correctly live-reloads

### Challenges/uncertainties
As we only have 2 translation files to load, this setup is quite simple. If we want to extend this to other apps with multiple translation files, we'll have to apply some translation file merging.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey (approved):
Seems to work well :) I was worried the rename of translation files would cause a problem for consuming apps, but linking to plugins seems to work fine.

PR #1315, opened 2025-05-15
Add click handler to block rdfa labels to allow for easy selection

piemonkey:
### Overview
Just adding the click handler helps a lot with the reported problems. Due to type problems, I didn't get much time to look in to keeping RDFa relationships valid or drag handling.

Part of https://binnenland.atlassian.net/browse/GN-5579

### Setup
N/A

### How to test/reproduce
When mousing over block labels you should now see a 'cell' cursor (think what excel shows when you select cells). Clicking the label selects the node. You can now e.g. copy/cut and paste somewhere else.

### Challenges/uncertainties
Weird type issues.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


piemonkey:
> Drag-and-drop already seems to work quite well tbh.

Huh, how did I not test that again? I must have tried it before the rest was fully working or something and assumed that it would keep not working...

> Not fully sure on which cursor to use though, maybe `pointer` or `move` is more applicable here?

I went for the 'cell' one as I thought it was just to select the node and it feels familiar to me from spreadsheets, but maybe it should be a pointer, and once selected, a 'grab'?

PR #1317, opened 2025-05-20
CIT-47: relationship-editor `lov` config

elpoelma:
### Overview
This PR adds a configurable `lovConfig` function which can be used as a config for the new relationship-editor.
Additionally, this PR also adds a `combineConfigs` functions which allows you to combine a series of configs.

[CIT-47](https://binnenland.atlassian.net/browse/CIT-47?atlOrigin=eyJpIjoiYTU2YjgxOTUyODYzNGIxOThlZmVmMDA3NGI3ZTE2ZjMiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the test-app and open the editable-node page
- Open the relationship-editor modal
- The subjects/predicates/objects should autocomplete using terms already used in the document
- The predicates/objects should additionally autocomplete using terms consumed from the LOV api. Note: you only get suggestions from the LOV api when your search string has length >= 1.
- The search should debounce, and irrelevant requests are not sent/cancelled

### Challenges/uncertainties
In an ideal scenario, I also wanted to expose a `taskify` function which taskifies (using ember-concurrency) a given config for a certain context. I did however not get this to work.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
> Looks good. I don't like how searching now runs at the speed of the slowest config being combined, so in this case the API lookup. I can't think of a good way to prevent that though as it would require returning intermediate results from a task...

Yeah it's not ideal indeed, but I also couldn't find a better way.

PR #1318, opened 2025-05-20
Feat/toolbar redesign

lagartoverde:
### Overview
Sorry this took longer than expected I was mentally blocked with a few things that made me very slow on this PR but it's done now, the toolbar proposed is almost the same as Saska design, we talked about the ordered list and decided to leave it like it's now, and for the alignment we decided the neutral icon should be the current alignment.
Everything else should match so if you find something weird let me know this also means we will have to redesign the toolbars on GN, RB... but that should be easy

GN-4737

### Setup
None

### How to test/reproduce
Just play around with the toolbar to check everything works as expected and that the design matches the figma

### Challenges/uncertainties
None

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma (changes requested):
This looks really nice! :) The colors and hover effects alone make a big difference.
I have some remarks though:
- In the design Saska made, the toolbar also has a bottom border, which I don't see in this implementation
- The entries in the alignment, list, color and table submenus do not have the same hover-effect as the other buttons. It would be nice to align these
- The arrow indicator of the list submenu is not correctly positioned

elpoelma (changes requested):
Looks good!
One nitpicky remark: the color dropdown menus still have an offset, while the other don't. When they are active, they also don't have a thick bottom-border.

elpoelma (approved):

elpoelma:
Something I still noticed, but does not need to be implemented in this PR. We probably want to align the design of the table-popup with the new toolbar design.

elpoelma (approved):

PR #1319, opened 2025-05-23
GN-5604: fix for literal relationship bug

elpoelma:
### Overview
This PR contains an approach to fixing the literal relationship bug.

**Approach contained in this PR**
We basically need to ensure that literal-nodes are completely uniquely identified by the `contentNodeMap`. This can be done by identifying it using:
- Subject which points to the literal node
- Predicate
- Object (value + language/datatype)

We can determine the object based on either the `content` attribute (if present), or the by using the `activeTag.text` attribute.
To ensure the `activeTag.text` attribute is correctly populated, the `postProcessTagAsRdfaNode` procedure needed to move to the `onTagClose` function.

I also needed to comment out https://github.com/lblod/ember-rdfa-editor/blob/0858cfcf31d0b0b40227269f919bac24aa3304b7/packages/ember-rdfa-editor/src/utils/_private/rdfa-parser/rdfa-parser.ts#L1121 ,not sure if this is a problem?

**Possible alternative approach?**
Do not use the `contentNodeMap`, but rather directly store `literal` vs `literalNode` vs `contentLiteral` terms when emitting quads/triples. This ensures that we already know what term we're dealing with, and makes the map obsolete (I think at least).

[GN-5604](https://binnenland.atlassian.net/browse/GN-5604?atlOrigin=eyJpIjoiOWRkNmQzZWY4OTkxNDRiNGFkNDFkNWU5NjQ0MGRmYjQiLCJwIjoiaiJ9)

### How to test/reproduce
- Add a `block_rdfa` resource node
- Add a `block_rdfa` literal node
- Add a relationship between the resource node and literal node
- Add a relationship with the same predicate pointing to a simple literal
- Reload the doc
- Ensure both relationships are kept intact

### Checks PR readiness
- [x] npm lint
- [x] no new deprecations
- [x] changelog

### DIFF
--- a/.changeset/curvy-ravens-type.md
+++ b/.changeset/curvy-ravens-type.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Toolbar redesign
--- a/.changeset/four-donuts-sell.md
+++ b/.changeset/four-donuts-sell.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Add support for predicate autocompletion in the content-predicate form
--- a/.changeset/free-eagles-happen.md
+++ b/.changeset/free-eagles-happen.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': patch
----
-
-Improve UX when entering a custom predicate/target option in the relationship-editor
--- a/.changeset/loose-aliens-work.md
+++ b/.changeset/loose-aliens-work.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Align table popup menu design with new toolbar design
--- a/.changeset/puny-stars-join.md
+++ b/.changeset/puny-stars-join.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Add section to relationship-editor to edit content predicate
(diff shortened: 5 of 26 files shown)


abeforgit:
@elpoelma I also don't know... this is from the original parser code I copied from Taelman
I could see it causing issues here maybe
https://github.com/lblod/ember-rdfa-editor/blob/292716f003589618dbcf6fceff06ab806c217fa2/packages/ember-rdfa-editor/src/utils/_private/rdfa-parser/rdfa-parser.ts#L1190

I'm not exactly sure what the collectChildTags boolean means though.
Based on not much more than gut feeling I think the edge case may be in the realm of XMLLiterals
I guess the safest approach would be to instead save the text to a property, and then when you need to access activeTag.text and its null, check if theres anything in that property?

One day we'll write our own parser -- if only to actually understand it fully :grimacing:

elpoelma:
> @elpoelma I also don't know... this is from the original parser code I copied from Taelman I could see it causing issues here maybe
> 
> https://github.com/lblod/ember-rdfa-editor/blob/292716f003589618dbcf6fceff06ab806c217fa2/packages/ember-rdfa-editor/src/utils/_private/rdfa-parser/rdfa-parser.ts#L1190
> 
> I'm not exactly sure what the collectChildTags boolean means though. Based on not much more than gut feeling I think the edge case may be in the realm of XMLLiterals I guess the safest approach would be to instead save the text to a property, and then when you need to access activeTag.text and its null, check if theres anything in that property?
> 
> One day we'll write our own parser -- if only to actually understand it fully 😬

I indeed think it is turned on when the datatype is `XMLLiteral`, as it then also takes the xml/html tags into account. I'll add an additional `content` property to `activeTag` (which either contains the text concatenated, or the `content` attribute. We can then use that property.

PR #1320, opened 2025-05-23
GN-5592: fix firefox cursor bug

elpoelma:
### Overview
This PR contains two adjustments to improve cursor behaviour around inline, non-editable nodes:

**The addition of two new ember-node options:**
- `domClassNames`: classnames to apply to the nodeview which is created
- `contentDomClassNames`: classnames to apply to the content node within the nodeview
These are needed/useful to workaround the `stopEvent` problem: we want prosemirror to ignore most events occuring inside the custom ember nodeview itself (e.g. button presses). This does mean that e.g. clicks before the content of such a node are not  handled by prosemirror => cursor is not correctly set. Events inside the `contentDOM` _are_ handled correctly, so the `contentDomClassNames` can allow custom styling on that node (e.g. spacing/padding which makes it easier to click before the content).

Aside from that, I still think it's a sensible default for prosemirror to ignore most events on the nodeview itself (except for maybe drag events).

**Removal of the obsolete firefox fake-cursor decoration**
The custom click handler is still needed, but it seems firefox now correctly handles displaying a native cursor (where before we needed a fake cursor).

[GN-5592](https://binnenland.atlassian.net/browse/GN-5592?atlOrigin=eyJpIjoiOWQwNjUyNjVhMzVmNDQ5MWI4NmQxNTZmMzhhZTNkMTEiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the test-app and open the 'Dummy plugins' page
- Insert the following two dummy nodes: `Card` and `Sample Block`
  * `Card` has no styling/padding on its content dom
  * `Sample Block` has padding on its content dom
- Insert a simple hyperlink at the start of each dummy node
- Observe the difference in click/cursor handling:
  * In `Card`, you can seemingly set a cursor before the link by clicking before it, but it will not work (the click handler of the `firefoxCursorFix` plugin is not called as we are clicking inside the nodeview)
  * In `Sample Block`, you can correctly set a cursor (which works) when clicking before the link (the click handler of the `firefoxCursorFix` plugin is correctly called as we are clicking inside the contentDOM)

### Challenges/uncertainties
- You'll notice some small visual bugs (like the cursor jumping around a second before setting the correct/fixed one). I have not yet found a solution for this, but I already consider this a large improvement upon the current situation.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/shy-days-wash.md
+++ b/.changeset/shy-days-wash.md
@@ -1,0 +1,7 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Introduction of two new ember-node options:
+- `domClassNames`: classnames to apply to the nodeview which is created
+- `contentDomClassNames`: classnames to apply to the content node within the nodeview
--- a/.changeset/whole-bats-scream.md
+++ b/.changeset/whole-bats-scream.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Remove obsolete firefox fake cursor styling. The fake cursor is no longer needed, the native browser cursor can now be used without issues.
--- a/packages/ember-rdfa-editor/ember-rdfa-editor.scss
+++ b/packages/ember-rdfa-editor/ember-rdfa-editor.scss
@@ -26,12 +26,6 @@
   width: 20px;
 }
 
-.ProseMirror-firefox-fake-cursor {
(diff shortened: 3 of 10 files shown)


piemonkey (approved):
Looks good and I think we can live with the jumping around a little.

PR #1321, opened 2025-05-26
Handle weird drag-drop cases by adding on-changed plugin

piemonkey:
### Overview
Most of the changes are to add the on-changed plugin, which generalises the process of generating transactions and appending them, when nodes of a certain spec are modified. This allows nodes like structures to keep their numbering and RDFa up to date.

Also adds a generic way to add transaction metadata when using transaction monads, including the special case of 'check' transactions, which only exist to verify that they could be applied. This re-enables some optimisations when we know a process won't fail as well as allowing us to prevent recursion in the on-changed plugin.

Part of https://binnenland.atlassian.net/browse/GN-5579
Plugins PR: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/577

### Setup
Can be tested in the test app, but makes sense to also test it in [the plugins PR](https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/577) to see the functionality used.

### How to test/reproduce
See plugins PR.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/eight-memes-occur.md
+++ b/.changeset/eight-memes-occur.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-rdfa-tools: disable `Add property` button when doc node is selected, as this functionality is now covered by the imported-resource editor
--- a/.changeset/metal-bobcats-drum.md
+++ b/.changeset/metal-bobcats-drum.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Add click handler for rdfa block labels that selects the block
--- a/.changeset/spicy-planets-leave.md
+++ b/.changeset/spicy-planets-leave.md
@@ -1,5 +1,0 @@
----
-'@lblod/ember-rdfa-editor': minor
----
-
-Add support for hiding elements from the rdfa-visualizer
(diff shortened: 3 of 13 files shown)


elpoelma:
> > The `onChanged` plugin makes sense, but I'm not completely following why dry-run transactions are necessary. In my view, if you're creating a transaction, you should already know they work (by checking the initial state), so a transaction would never be a dry run.
> 
> The problem with this is that transaction monads don't know whether they're being run just to test the initial state or in order to actually apply the resulting transaction. So either we need to verify that a transaction monad will succeed before running it (which would duplicate some logic), or we have some way of passing in that it is a dry run.
> 
> An example of where this would have been useful (except that the on-changed plugin solves the problem in a better way), is for recalculating structure numbers when inserting (or checking inserting) structures. Knowing that it's a 'dry run' allows us to avoid calculating new numbers, as this will never fail.

Right, so if a transaction is a dry-run, plugins can ignore them if they want?

piemonkey:
> Right, so if a transaction is a dry-run, plugins can ignore them if they want?

If it won't change whether or not the dry run succeeds, then yes.

PR #1324, opened 2025-06-06
relationship-editor: improve UX of entering custom predicate/target option

elpoelma:
### Overview
This PR adds the `ember-power-select-with-create` (https://github.com/cibernox/ember-power-select-with-create) dependency to improve the UX of entering custom predicate/target option in the dev-mode version of the relationship editor.

### How to test/reproduce
- Start the test-app and open the editable nodes page
- Open the relationship-editor modal in dev-mode
- Enter a valid term in the search field
- Ensure an additional option in the list of options (on top) now appears corresponding with your current search query

### Challenges/uncertainties
- The main caveat here is that we need to wait for the `options` promise to resolve before showing the custom option.
- We might also want to apply the same UX to the user-facing relationship-editor modal.

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/free-eagles-happen.md
+++ b/.changeset/free-eagles-happen.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Improve UX when entering a custom predicate/target option in the relationship-editor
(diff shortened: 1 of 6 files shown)


piemonkey (approved):
Looks good. Glad to get rid of the hacky power-select add logic.

PR #1326, opened 2025-06-12
Add controller block param to editor-container to avoid undefined checks

piemonkey:
### Overview
This should allow editor components such as toolbar buttons and cards to rely on the controller being present, instead of needing to accept `undefined` as a possibility and handle it gracefully.

I re-named the blocks so that the change can be backwards compatible, otherwise it would be a breaking change, as failing to pass a `@controller` argument would prevent the sidebar and toolbar from being rendered. There are some components that could have their `if (this.controller)` checks removed, but I decided not to do that, also for backwards compatibility.

N/A

### Setup
To test the types it's best to test in GN as the `rdfa-editor-container.gts` file is type checked, unlike the usages of EditorContainer in the test app.

### How to test/reproduce
If you pass an undefined controller to EditorContainer, the sidebar and toolbar should not be shown. Otherwise, everything should work as before.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/wild-humans-cry.md
+++ b/.changeset/wild-humans-cry.md
@@ -1,6 +1,7 @@
 ---
-'@lblod/ember-rdfa-editor': major
+'@lblod/ember-rdfa-editor': minor
 ---
 
-Add always defined controller block param to EditorContainer to allow cutting down on boilerplate is-not-undefined checks in editor components
-BREAKING: it is now necessary to pass a SayController to the EditorContainer component
+Add always defined controller block param to EditorContainer to allow cutting down on boilerplate is-not-undefined checks in editor components.
+This requires a SayController be passed to EditorContainer.
+For backwards compatibility, the blocks that are only shown when the controller exists are named differently to the previous blocks.
--- a/.changeset/wild-humans-cry.md
+++ b/.changeset/wild-humans-cry.md
@@ -1,0 +1,6 @@
+---
+'@lblod/ember-rdfa-editor': major
+---
+
+Add always defined controller block param to EditorContainer to allow cutting down on boilerplate is-not-undefined checks in editor components
+BREAKING: it is now necessary to pass a SayController to the EditorContainer component
--- a/packages/ember-rdfa-editor/src/components/editor-container.gts
+++ b/packages/ember-rdfa-editor/src/components/editor-container.gts
@@ -1,6 +1,6 @@
 import Component from '@glimmer/component';
 import { hash } from '@ember/helper';
-import { and } from 'ember-truth-helpers';
+import { and, or } from 'ember-truth-helpers';
 import AuLoader from '@appuniversum/ember-appuniversum/components/au-loader';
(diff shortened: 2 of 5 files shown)


piemonkey:
> It feels a bit weird/unnatural to me to pass an @controller argument to the parent container of the editor, which in itself does not really require the @controller

But it does require the @controller if it wants to only show it's contents if it exists. We could change it to `hasController: boolean`, but that makes the API more awkward given the limitations of logic in templates (we can't simply pass `!!this.controller`).

I couldn't think of another API that doesn't significantly change how the editor is used, but I'm happy to consider others if you have any ideas.

> Instead of deprecating the top and aside yieldables, we could also introduce some kind of deprecation flag

You mean to add a console deprecation warning if `aside` or `top` are passed instead of jsdoc deprecating it? I kinda see those kinds of logs as most useful for cases where a jsdoc deprecation is not possible, as the developer experience to me feels much worse (you find there is a problem if/when you look at the logs and it's not hidden by other deprecations instead of just seeing it in the code). I guess we could have both.

elpoelma:
> But it does require the @controller if it wants to only show it's contents if it exists. We could change it to `hasController: boolean`, but that makes the API more awkward given the limitations of logic in templates (we can't simply pass `!!this.controller`).

Yes that's indeed fair.



> You mean to add a console deprecation warning if `aside` or `top` are passed instead of jsdoc deprecating it? I kinda see those kinds of logs as most useful for cases where a jsdoc deprecation is not possible, as the developer experience to me feels much worse (you find there is a problem if/when you look at the logs and it's not hidden by other deprecations instead of just seeing it in the code). I guess we could have both.

Instead of deprecating `aside` and `top` and introducing new yieldables, I was thinking more of introducing some kind of flags, e.g. like in https://github.com/appuniversum/ember-appuniversum/blob/249d323a20b6434e3ebd2cb6efb005b2162f7487/ember-cli-build.js#L26-L27 . This way we can keep the yieldables, and enforce the new strategy when the flag is opted into. With a new major, we could then remove the flag.

piemonkey:
> Instead of deprecating `aside` and `top` and introducing new yieldables, I was thinking more of introducing some kind of flags, e.g. like in [appuniversum/ember-appuniversum@`249d323`/ember-cli-build.js#L26-L27](https://github.com/appuniversum/ember-appuniversum/blob/249d323a20b6434e3ebd2cb6efb005b2162f7487/ember-cli-build.js#L26-L27) . This way we can keep the yieldables, and enforce the new strategy when the flag is opted into.

Ah, I see. I'm not sure I agree for this case though as I don't think the yieldable names are good ones. `top` is not at the top if you pass `showToolbarBottom` and while `aside` isn't outright wrong, it doesn't give you much hint as to what it is. I think it makes more sense to change the names to match the terminology used in the other arguments, as I have tried to do with `toolbar` and `sidebarRight`.

elpoelma (approved):

PR #1328, opened 2025-06-13
Fix dropdown bug

lagartoverde:
### Overview
Added a fallback focus so the focus trap doesn't fail anymore

GN-4737

### Setup
None

### How to test/reproduce
Try to get in a gap cursor position where none of the text formatting is enabled, then check that the dropdown still works

### Challenges/uncertainties
Non a keyboard user, the focus seems good to me, but not sure if I broke something weird

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/rich-sheep-teach.md
+++ b/.changeset/rich-sheep-teach.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Fix bug with toolbar dropdown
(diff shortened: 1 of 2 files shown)


lagartoverde:
> Is the additional `div` necessary? I think we should also be able to add the modifier to the parent div?

I tried, it doesn't work, don't ask why xd

elpoelma:
> I think is because the fallbackFocus should be inside the focustrap

The example of the underlying focus-trap lib also shows that you can pass the focus-trap container as the fallbackFocus: https://github.com/focus-trap/focus-trap/blob/master/docs/js/tricky-initial-focus.js

elpoelma (approved):

PR #1329, opened 2025-06-13
WIP: parse headings containing RDFa as `block_rdfa`

elpoelma:
### Overview
While we technically support rdfa on heading nodes, we don't really support it in a user-friendly/good way.
To solve this issue, we could parse headings containing RDFa as `block_rdfa` nodes containing a simple heading node (without RDFa). An example of such a solution is included in this PR.

### How to test/reproduce
- Start the test-app
- Insert html containing a heading with RDFa
- Notice that it is parsed as a `block_rdfa` node (containing the RDFa) with a simple `heading` node inside.

### Challenges/uncertainties
- Currently, the changes in this PR are breaking (as the rdfa-mode of headings are no longer supported). But we could introduce a deprecation-flag or something for this change.
- We might also want to apply this to other nodes than `heading`.

### Checks PR readiness
- [ ] changelog
- [ ] npm lint
- [ ] no new deprecations


piemonkey (commented):
For existing headings with links (made using the rdfaAware mode), this seems to work as expected, so I think just keeping the config option, but deprecating it, would work well to make this backwards compatible.

It seems that now there is a difference between headings that are loaded and headings that are inserted, as in order to insert a heading with an RDFa block, it's necessary to both insert the block and the heading. I think the most expected behaviour would be to parse headings created with `RdfaAware: true` as just a heading if they do not have any backlinks, and only with a wrapping block if they do. Alternatively, we could always insert a block_rdfa with every heading, but I think this would be more confusing...

PR #1331, opened 2025-06-17
test-app: upgrade to ember 6.5.0 + move routes to `gts` format

elpoelma:
### Overview
This PR upgrades the test-app to ember 6.5.0 and converts the route templates to the `gts` format. This also means that the controllers are no longer necessary.

This will make it easier to test/check the types of the editor package.

### How to test/reproduce
- Check if the typecheck still succeeds
- Start the test-app. It should work as before

### Checks PR readiness
- [ ] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/curly-fans-run.md
+++ b/.changeset/curly-fans-run.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add explicit support for ember-source version 6.x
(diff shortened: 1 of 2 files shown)


piemonkey (approved):
Looks good and happy to see gts in routes :) We should probably up the peer-dep of the editor to include ember-source 6.x too.

PR #1333, opened 2025-06-24
rdfa-visualizer: hide resource node when necessary

elpoelma:
### Overview
This PR ensures that resource nodes entries (including the 'toggle' button) are fully hidden from the rdfa-visualizer when configured this way (using `hidden: true`).

[CIT-57](https://binnenland.atlassian.net/browse/CIT-57?atlOrigin=eyJpIjoiMjA0YTIzZjkwZWEzNDE3M2E2M2EwZWRiYTMwODU4OTQiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the test-app
- In the visualizer config of the editable-nodes page, adjust the resource-node config to return `[{ hidden: true }]` in some/all cases
- Ensure these cases are now fully hidden from the visualizer panel

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


elpoelma:
> hmmm there's still the issue that empty list items are now rendered...

should be solved now with an extra level of wrapping :grimacing:

elpoelma:
> Hmm, this has the odd side-effect that if the ResourceName config returns `{ hidden: true }`, then a resource which links to that Resource will have the predicate appear in it's list in the visualiser, but with no value: image

Yes true, I think this is also the case with e.g. the `NamedNode`, `Literal` or `LiteralNode` config options. It depends on whether we want to put that responsibility at the user side or not...

piemonkey:
> > Hmm, this has the odd side-effect that if the ResourceName config returns `{ hidden: true }`, then a resource which links to that Resource will have the predicate appear in it's list in the visualiser, but with no value
> 
> Yes true, I think this is also the case with e.g. the `NamedNode`, `Literal` or `LiteralNode` config options. It depends on whether we want to put that responsibility at the user side or not...

I was thinking that perhaps it would make sense to add a field such as `isTopLevel` to the `GeneratorContext`. This way we could treat top level (unlinked) resources differently. So we could hide snippets only if they're not linked from somewhere else, for example. As it is, the 'user' has the option to either hide them everywhere or not at all.

piemonkey:
> Technically, I guess the user could check if the node passed in the generatorcontext has backlinks or not, it would mean the same as being top-level right?

Hmm, true, except for the handling of loops. The visualiser should display the node with the lowest depth in the document as the top level, but they will all have backlinks. I started on this for the Citerra PoC before getting distracted on other things, I'll put it into a PR.

elpoelma:
> > Technically, I guess the user could check if the node passed in the generatorcontext has backlinks or not, it would mean the same as being top-level right?
> 
> Hmm, true, except for the handling of loops. The visualiser should display the node with the lowest depth in the document as the top level, but they will all have backlinks. I started on this for the Citerra PoC before getting distracted on other things, I'll put it into a PR.

Alright, is it ok if I merge this PR then?

piemonkey (approved):
Here's the PR: https://github.com/lblod/ember-rdfa-editor/pull/1334

PR #1334, opened 2025-07-01
Add context to rdfa-visualiser to say if we're at the top level

piemonkey:
### Overview
Allows for hiding elements only when they are not part of a hierarchy for example, snippets.

Related to https://github.com/lblod/ember-rdfa-editor/pull/1333

### Setup
N/A

### How to test/reproduce
In the editable-nodes sample, Snippets should be hidden from the top level, so pasting in HTML with snippets should not display them in the top level (even if the nodespecs don't work for obvious reasons). Linking any resource to the snippet should allow it to appear in the visualiser in the correct place.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/mean-waves-begin.md
+++ b/.changeset/mean-waves-begin.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add context to RDFa visualiser components that says if a resource is at the top level so the display can be changed accordingly
--- a/packages/ember-rdfa-editor/src/components/_private/common/configurable-rdfa-display.gts
+++ b/packages/ember-rdfa-editor/src/components/_private/common/configurable-rdfa-display.gts
@@ -7,11 +7,11 @@ import { trackedFunction } from 'reactiveweb/function';
 import AuLoader from '@appuniversum/ember-appuniversum/components/au-loader';
 import AuPill from '@appuniversum/ember-appuniversum/components/au-pill';
 import { type OutgoingTriple } from '#root/core/rdfa-processor.ts';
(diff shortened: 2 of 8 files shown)


elpoelma (changes requested):
https://github.com/lblod/ember-rdfa-editor/blob/09451449a9bcde555ea2c331c5672f2e7c552948/packages/ember-rdfa-editor/src/components/_private/relationship-editor/card.gts#L444 gives type errors now, as the `@isTopLevel` argument is not passed. 
Alternatively we could also not pass the `isTopLevel` arg to the `configurable-rdfa-display` component, and already encode it in the generator function we pass to that component. (as the component itself doesn't really need that argument, and it is not super relevant for every usage of it)

piemonkey:
> https://github.com/lblod/ember-rdfa-editor/blob/09451449a9bcde555ea2c331c5672f2e7c552948/packages/ember-rdfa-editor/src/components/_private/relationship-editor/card.gts#L444
> gives type errors now, as the `@isTopLevel` argument is not passed.
> Alternatively we could also not pass the `isTopLevel` arg to the `configurable-rdfa-display` component, and already encode it in the generator function we pass to that component. (as the component itself doesn't really need that argument, and it is not super relevant for every usage of it)

Oops, was sure I changed that one too...

I thought about it and actually, I think it makes more sense to have `isTopLevel` passed as part of the whole context, as otherwise, you're right that we're passing things to the component that it doesn't directly care about. I've changed that over and fixed the type error too.

PR #1366, opened 2026-03-27
Context actions

kobemertens:
### Overview
Adds a floating context menu with actions that are relevant to the current cursor position.

https://binnenland.atlassian.net/browse/GN-6016

### How to test/reproduce
The floating plus should show  actions based on the callback functions passed to the `ContextActionsContainer` component.

### Challenges/uncertainties
See inline comments

### Checks PR readiness
- [x] changelog
- [x] npm lint


packages/ember-rdfa-editor/scss/_a-custom-components.scss
```diff
@@ -19,6 +19,7 @@
 @forward "./c-color-selector.scss";
 @forward "./c-image.scss";
 @forward "./c-attribute-editor.scss";
+@forward "./c-contextual-actions";
```
kobemertens, line 22: Is this needed? Do I need to add `packages/ember-rdfa-editor/scss/_c-floating.scss‎` here as well? Seems to work correctly without it as well?

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
```diff
@@ -0,0 +1,210 @@
...
+      }))
+      .filter((group) => group.actions.length > 0)
+      .toSorted(sortByPriority);
+  }
+
+  get controller() {
+    return this.args.controller;
+  }
+
+  get referenceElement() {
```
kobemertens, line 62: This function is very similiar to other components using `floatingUI`, might be worth it to encapsulate this in a reusable function? Any opinions on this?

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
```diff
@@ -0,0 +1,210 @@
...
+            <div
+              class="say-contextual-actions-menu-group-header au-u-muted"
+              {{this.observeSticky}}
+            >
+              {{group.label}}
+            </div>
+            <div class="au-u-padding-left-tiny au-u-padding-right-tiny">
+              {{#each group.actions as |actionItem|}}
+                <button
+                  disabled={{not this.canExecuteAction actionItem}}
```
kobemertens: Do we want this check here? We could also move this responsibility to the plugin that defines the action?

packages/ember-rdfa-editor/src/plugins/contextual-actions/index.ts
```diff
@@ -0,0 +1,18 @@
+import { type Command } from 'prosemirror-state';
+
+export type ContextualAction = {
+  id: string;
```
kobemertens, line 4: Do we want each action to have an ID? Might be useful in the future but is not used now.

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
```diff
@@ -0,0 +1,209 @@
...
+        const coordsFrom = this.controller.mainEditorView.coordsAtPos(
+          selection.from,
+          -1,
+        );
+        const coordsTo = this.controller.mainEditorView.coordsAtPos(
+          selection.to,
+          -1,
+        );
+        const left = (coordsFrom.left + coordsTo.left) / 2;
+        const right = (coordsFrom.right + coordsTo.right) / 2;
```
piemonkey: I find the positioning of the floating menu based on this quite surprising. It feels strange that the menu appears where the selection is, not where the button is.

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
piemonkey: Seems reasonable to me. Is there a reason not to?

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
piemonkey: I think it would be more generic in the plugin, but we'd still have to supply the controller. There's also the question of if we want to include actions we can't perform.

packages/ember-rdfa-editor/src/components/_private/common/selection-tooltip.gts
```diff
@@ -65,4 +70,22 @@ export default class SelectionTooltip extends Component<Args> {
       hide({ strategy: 'escaped' }),
     ];
   }
+  <template>
+    {{! @glint-nocheck: not typesafe yet }}
```
piemonkey: Why not make this typesafe now that it's in a `gts` component?

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @action
+  selectAction(action: ContextualAction) {
+    this.executeAction(action);
+    this.showActions = false;
+  }
+
+  <template>
+    {{! @glint-nocheck: not typesafe yet }}
```
piemonkey: This could be typesafe too.

test-app/app/templates/plugins.gts
```diff
@@ -232,6 +240,14 @@ export default class extends Component {
               @nodeViews={{this.nodeViews}}
               @rdfaEditorInit={{this.rdfaEditorInit}}
             />
+
+            {{#if this.rdfaEditor}}
+              <ContextualActionsContainer
```
piemonkey, line 245: Hmm, it feels like this should be inside `<Editor>` to avoid having to add this boilerplate in every consuming app

piemonkey (commented):
The behaviour when selecting a node within a block seems unintuitive from a user point of view. I get that the actions are related to the current selection, but since the plus button is on the left, it looks like they should be related to the block. For example, when selecting an oslo-location which is inside a plaatsbepaling and pressing the plus:

Image

test-app/app/dummy-plugins/expose-contextual-actions.ts
```diff
@@ -0,0 +1,138 @@
...
+        if (dispatch) {
+          const tr = state.tr;
+          tr.replaceSelectionWith(
+            state.schema.text(action.insert ?? action.label),
+          );
+          if (tr.selection.$anchor.nodeBefore) {
+            const resolvedPos = tr.doc.resolve(
+              tr.selection.anchor - tr.selection.$anchor.nodeBefore?.nodeSize,
+            );
+            tr.setSelection(new NodeSelection(resolvedPos));
```
piemonkey: This is just the dummy app, so it's not a big deal if this has strange behaviour, but it seems to always select the thing that was just inserted, but in a way that isn't visible in the editor as a selection. This means if you insert something, then type, it removes what you just inserted.

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
kobemertens: Yeah, I get the confusion here. My reasoning was that the text is inserted at the cursor position, so it makes sense in that way to display the menu there. And placing the button near the cursor seemed weird as well since it would be moving while the user is typing. So not entirely sure what is the best way to go here.

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
kobemertens: No not really, will fix!

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
kobemertens: Yeah, don't think it makes sense to show actions a user can't perform, will leave the checking to the implementing plugins

test-app/app/dummy-plugins/expose-contextual-actions.ts
kobemertens: fixed in [425d00b](https://github.com/lblod/ember-rdfa-editor/pull/1366/commits/425d00b7592973aba2a51d185466772eb2fdb015)

test-app/app/templates/plugins.gts
kobemertens: Kind of depends on how we want to treat core plugins like this? Since there will be an accompanying proseplugin in a follow-up PR, I'm not sure if it makes sense to always include this component in the `Editor` component while leaving the ProsePlugin as 'optional'? Maybe @abeforgit has opinions on this as well?

packages/ember-rdfa-editor/src/components/_private/common/contextual-actions-menu.gts
kobemertens: Maybe we move the cursor to a new empty line first before showing the options?

test-app/app/templates/plugins.gts
abeforgit: I'm ok with the boilerplate for now

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
```diff
@@ -0,0 +1,125 @@
...
+  get visible() {
+    return this.groups.length > 0 && !this.showActions;
+  }
+
+  loadAndShowActions = task(async () => {
+    const getActions = this.args.getActions ?? [];
+    const editorState = this.controller.mainEditorState;
+
+    const actions = (
+      await Promise.all(getActions.map((cb) => cb(editorState)))
```
abeforgit: I think we need to at least consider some error handling of the actiongetter promises here, endpoints could be down and such, and it'd be good if that was handled somewhat gracefully (showin a user-facing error with as much info/granularity as is feasible)

packages/ember-rdfa-editor/src/commands/replace-selection-and-select-node.ts
```diff
@@ -0,0 +1,23 @@
+import { type Command, NodeSelection } from 'prosemirror-state';
+import { Node } from 'prosemirror-model';
+
+export const replaceSelectionWithAndSelectNode = (node: Node): Command => {
```
abeforgit: this function doesn't appear to be used anywhere

packages/ember-rdfa-editor/src/helpers/set.ts
```diff
@@ -0,0 +1,15 @@
+export default <T extends string, U>(
```
abeforgit: this helper doesn't appear to be used anywhere

abeforgit (changes requested):
some small nitpicks on the code, once those are resolved + the linter errors I think this is good to go

packages/ember-rdfa-editor/src/commands/replace-selection-and-select-node.ts
kobemertens: This is a function that I copied from the plugins repo, but I think should live here. I needed it at some point for this feature, but not anymore. Still made sense to keep it here since it might be useful in the future, and we might want to import this one in the plugins repo instead.

packages/ember-rdfa-editor/src/helpers/set.ts
kobemertens: I'll remove until it is used

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
kobemertens: [0315622](https://github.com/lblod/ember-rdfa-editor/pull/1366/commits/0315622967ab7e772b72a65d48f5b81d52fa9ad1)

Is this enough? Maybe we should improve when specific cases arise?

packages/ember-rdfa-editor/src/commands/replace-selection-and-select-node.ts
abeforgit: I'd rather see a separate pr then where the function is actually moved from the plugins to here

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
abeforgit: yeah seems ok

abeforgit (changes requested):
@kobemertens you need to still fix the CI errors and the merge conflicts

packages/ember-rdfa-editor/src/commands/replace-selection-and-select-node.ts
kobemertens: removed in [9c7831e](https://github.com/lblod/ember-rdfa-editor/pull/1366/commits/9c7831e7185ec1c07786e7471206540bce11ed7c)

PR #1367, opened 2026-04-01
Change literal nodes to work as pointers

piemonkey:
### Overview
Literal nodes, while still having the same name and `rdfaNodeType`, are now ~~pointer~~rdfa nodes. The name has not been changed to avoid a huge number of breaking changes.

Rdfa nodes can either be pointed to by their content (for backwards compatibility, this is the default and can be overridden by specifying the `hasNonLiteralContent` attr), in which case they behave as literal nodes. Any rdfa nodes (those pointed by their content or not) can be the object of relationships. These relationships are serialised to data attributes so they are maintained, if they can't make a complete triple, so can't be encoded to RDFa.

A pointer chain can be 'completed' by adding an RDFa aware node (self-pointing rdfa node (aka literal) or resource node) with a `pointsToNode` attr. This attr holds the rdfaid of the pointed node. When the document is serialised, any of these nodes are serialised with RDFa encoding triples where subject and predicate are from any relationships to the pointed node and the object is the node with the `pointsToNode` attr.

Jira ticket: https://binnenland.atlassian.net/browse/GN-6037
Plugins PR using this logic: https://github.com/lblod/ember-rdfa-editor-lblod-plugins/pull/636

### Setup
~~Since the RDFa tools have not yet been modified to set up `pointed` attrs, you need to either create them manually, or use the plugins PR that makes use of this logic.~~ The devtools are not great for seeing these relationships but at least you can create them.

### How to test/reproduce
- To set up a 'pointer' relationship, first create an RDFa node (using the new sidebar button)
- Then create any RDFa links to this node in the usual way. These will not appear in RDFa, as there's no pointer to give the object, but they should be maintained.
- To create a pointer, create a resource node or literal node, and create a `pointsToNode` attr that contains the rdfaid of the pointer. This can also be done using the 'link pointer' sidebar button.
- You should now see any links in RDFa, pointing to the node with the pointer.
- (the plugins PR contains a way to test this for marcodes)

### Challenges/uncertainties
This seems like a very roundabout way to solve this problem

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations


abeforgit (changes requested):
code looks sensible, and all things considered much cleaner than I expected it to have to be 👍 
Haven't ran it yet, let me know if you have some pointers on how to test

packages/ember-rdfa-editor/src/core/say-serializer.ts
```diff
@@ -37,40 +37,56 @@ import {
   Schema,
 } from 'prosemirror-model';
 import type { EditorState } from 'prosemirror-state';
+// Used for jsdocs
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { MarkSpec, NodeSpec } from 'prosemirror-model';
```
abeforgit: alternatively, you can use `@import` https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#import

packages/ember-rdfa-editor/src/core/schema.ts
```diff
@@ -438,55 +496,78 @@ export function renderRdfaAttrs(
...
       'data-say-id': rdfaAttrs.__rdfaId,
+      'data-literal-node': 'true',
       ...datatypeAndLanguage,
     };
+
+    if (!backlink) {
+      return baseAttrs;
+    }
+
+    if (datatypeAndLanguage['datatype'] || datatypeAndLanguage['lang']) {
```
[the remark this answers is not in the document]
abeforgit: yeah separate types sound the best, but indeed not really the moment to introduce that I think

packages/ember-rdfa-editor/src/core/say-serializer.ts
piemonkey: Hmm, I tried that but my editor didn't seem to pick it up. Typedoc does though, but mangles the comments where we use `@see`, as they intend it to be used for a list of references. I'll figure a way to change these docs that works with typedoc.

packages/ember-rdfa-editor/src/core/schema.ts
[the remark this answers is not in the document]
piemonkey: Hmm, so I'll add another data attribute to pointer nodes to say that they are a pointer, not a literal. We can maybe introduce another node type later.

packages/ember-rdfa-editor/src/core/rdfa-types.ts
```diff
@@ -9,11 +9,14 @@ import { type SayNamedNode } from './say-data-factory/index.ts';
...
 export const rdfaNodeTypes = ['resource', 'literal'] as const;
+export type RdfaNodeType = (typeof rdfaNodeTypes)[number];
 export interface RdfaAwareAttrs {
   __rdfaId: string;
-  rdfaNodeType: (typeof rdfaNodeTypes)[number];
+  rdfaNodeType: RdfaNodeType;
   backlinks: IncomingTriple[];
   externalTriples?: FullTriple[];
+  /** This node is being pointed to by a pointer node */
+  pointed?: string;
```
kobemertens: Continuing our conversation from our call here, I think this variable naming could be more descriptive, although it is not something super easy to name. I was thinking maybe something like `pointedAtBy` or `holdsDataFor`? But other suggestions are definitely welcome

packages/ember-rdfa-editor/src/core/schema.ts
```diff
@@ -390,15 +428,40 @@ export function renderInvisibleRdfa(
       ...externalElements,
     ]);
   }
-  if (nodeOrMark.attrs['rdfaNodeType'] === 'resource') {
-    const backlinks = nodeOrMark.attrs['backlinks'] as IncomingTriple[];
+  let pointsToAttrs: Attrs | undefined = renderable.attrs;
+  if (renderable.attrs['pointsToNode'] && state) {
+    // When serializing, follow the references back to the furthest RDFa node
+    while (pointsToAttrs?.['pointsToNode']) {
```
abeforgit: could we do some cycle detection here or set an upper bound of how long the link is allowed to be? seems easy to accidentally make this go infinite

packages/ember-rdfa-editor/src/core/schema.ts
```diff
@@ -390,15 +428,40 @@ export function renderInvisibleRdfa(
...
     }
-  } else if (nodeOrMark.attrs['rdfaNodeType'] === 'literal') {
-    const backlinks = nodeOrMark.attrs['backlinks'] as IncomingTriple[];
-    if (backlinks.length > 1 && nodeOrMark instanceof PNode) {
-      const literalNodeId = nodeOrMark.attrs['__rdfaId'] as string | null;
+    if (pointsToAttrs) {
+      // pointsToAttrs is only non-null when serializing and we have a 'pointsToNode' value
+      const pointerBacklink = (
+        pointsToAttrs['backlinks'] as RdfaAttrs['backlinks']
+      )[0];
```
abeforgit, line 456: Just a sanity check, but are we sure here the first backlink will be the right one?

packages/ember-rdfa-editor/src/core/schema.ts
```diff
@@ -423,70 +486,122 @@ export function renderInvisibleRdfa(
...
+        : 'false',
+      'data-points-to-node': rdfaAttrs.pointsToNode ?? null,
       ...datatypeAndLanguage,
     };
+
+    // Handle potentially complete chain of pointers
+    let pointsToAttrs: Attrs | undefined = rdfaAttrs;
+    if (rdfaAttrs['pointsToNode'] && state) {
+      // When serializing, follow the pointers back to the furthest RDFa node
+      while (pointsToAttrs?.['pointsToNode']) {
```
abeforgit: same comment as the other while loop

abeforgit (changes requested):
It's not a focus of the pr, but I was surprised by the "direction" choice when pointing a node to an rdfaNode "intermediary" (using the new "link pointer" button)
I couldn't really visualize which one I should choose, so if theres a small comment you could add to make that a little more clear that would be nice

packages/ember-rdfa-editor/src/core/schema.ts
piemonkey: Much like literal nodes, 'rdfa nodes' can only (for now) have one backlink. We could extend it to handle more, but we'd need to figure out how we want to encode that in RDFa.

packages/ember-rdfa-editor/src/core/schema.ts
piemonkey: I added loop detection to these

packages/ember-rdfa-editor/src/core/schema.ts
piemonkey: I added loop detection to these

abeforgit (approved):

PR #1369, opened 2026-04-14
feature/slash commands

kobemertens:
https://binnenland.atlassian.net/browse/GN-6051

### How to test/reproduce
Typing a slash on a line where the placeholder instruction is shown should open the context menu. Navigating away or clicking outside of the menu should close it. 

### Checks PR readiness
- [x] changelog
- [ ] npm lint


abeforgit (changes requested):
- Drag selections that start on top of the placeholder text don't work correctly:

https://github.com/user-attachments/assets/c7fbf7a2-1d3b-4605-bcf6-045ccd1216ee

- the textsize of the placeholder text is affected by the textstyle of the surrounding paragraph:

Image

- especially weird with outlining/indentation:

Image


- you can keep typing slashes and it will keep showing the menu:

Image


- Ideally the tab key would also cycle through the options like the arrow keys do

kobemertens:
> Drag selections that start on top of the placeholder text don't work correctly

This one might be difficult to solve, since Notion seems to have the same issue... I feel like it might be a html/browser restriction but I will have a look if something can be improved

**Edit:** Now it just works for me, I have not changed anything... Maybe it is also dependent on the content of the document?

PR #1374, opened 2026-05-08
PNode context actions

kobemertens:
### Overview
Implements functionality to automatically open the context actions menu when a `PNode` is selected.

https://binnenland.atlassian.net/browse/GN-6003

### How to test/reproduce
In `test-app/app/dummy-plugins/expose-contextual-actions` you can add to `getContextualGroups` and `getContextualActions` to test. The presence of `ContextualActionGroup`s for a given `EditorState` determine if the contextual menu is opened.
This PR adds the selection of a `PNode` as a potential trigger to open the context menu. If any group is returned when a `PNode` is selected, the contextual actions menu will open. Pressing Escape should close the menu.

### Challenges/uncertainties
Which nodes to open the menu for is left to the implementing plugins (based on if it returns groups given the selection state), the implementing plugins thus need to be strict enough, otherwise the menu might show up in places where they do not expect it. We could consider adding a `openOnNodeSelection` parameter to the `ContextualActionGroup` spec but this also feels redundant to me since they have the ability to decide this by returning actions or not.

### Checks PR readiness
- [x] changelog
- [x] npm lint


abeforgit (changes requested):
@kobemertens you need to address CI failures and merge conflicts without waiting for a review

packages/ember-rdfa-editor/package.json
```diff
@@ -250,7 +250,9 @@
     "ember-focus-trap": "catalog:",
     "ember-headless-form": "catalog:",
     "ember-headless-form-yup": "catalog:",
+    "ember-lifeline": "^7.0.0",
```
abeforgit, line 253: can you explain why you've chosen to add this dependency?

abeforgit (changes requested):
It's at the moment unclear how to test this, I see the menu popup when selecting placeholder nodes, but I'm not sure what the intention is behind the setup. Can you add testing instructions, or if the PR is not ready, mark it as draft please?

packages/ember-rdfa-editor/package.json
kobemertens: This is used here: https://github.com/lblod/ember-rdfa-editor/pull/1374/changes#diff-437b8b0f232d5e31226e4833cab1ffbb4b112fe6b783aa31b484a25d912961c2R275-R277

I had some iterations to get this behavior, I started using `ember-focus-trap` which does way too much and captured the focus (not very unexpected ;) ), but this is not what I wanted. I just want to focus once on render and leave the rest of the focusing behavior unchanged. So I started using a ad-hoc `focus` modifier which just did `element.focus()`, but this did not work (not really sure why?). So I had an iteration which used `@ember/runloop` but I got a linter error discouraging the use of this, and recommending to use the `ember-lifeline` package instead.

abeforgit: afaict it works fine with `element.focus({preventScroll: true})` in the modifier (without the runTask)
did you try that?

abeforgit:
@kobemertens also, this pr includes #1375 
as a tip: a cleaner way to structure this would've been to rebase this branch onto the appuniversum pr branch, and then make this pr target the AU branch instead

We don't really have a rule for this, so it's ok, but just something you can do with git you may not have thought of

kobemertens:
Did not think of that indeed, but for review that is a nicer option, will keep it in mind!

packages/ember-rdfa-editor/package.json
kobemertens: Hah, seems to work with `preventScroll` did not expect that to be the key to make this work, thanks!

kobemertens:
The removal of the lifeline dep introduces a new issue
```
Uncaught Error: Assertion Failed: You attempted to update `state` on `SayView`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
```
I think this is because changing the focus causes a blur transaction to be dispatched which updates the editor state, which causes ember to complain.

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
```diff
@@ -19,18 +19,31 @@ import {
...
 } from '#root/plugins/slash-commands/index.ts';
-import { trackedFunction } from 'reactiveweb/function';
+import { localCopy } from 'tracked-toolbox';
+import { restartableTask, timeout } from 'ember-concurrency';
+
+// We don't use a trackedfunction because it causes the menu to reload
+// after the first load (after the debounce time)
+// See https://discord.com/channels/480462759797063690/1501197288603910266
+// eslint-disable-next-line ember/no-at-ember-render-modifiers
+import didInsert from '@ember/render-modifiers/modifiers/did-insert';
```
abeforgit: Afaict using the did-insert here has nothing to do with the trackedFunction debouncing, you solve that by manually rerunning the task in the setSearchQuery function
I think that's the right way to do it here, the fewer tracking magic the better.

The didInsert is just here to launch it the first time, which you can (and should) instead do with a custom modifier

packages/ember-rdfa-editor/src/components/plugins/contextual-actions/container.gts
```diff
@@ -42,6 +55,8 @@ export default class ContextualActionsContainer extends Component<Args> {
   @tracked localEditorState: EditorState | null = null;
   @tracked plusButtonClicked = false;
 
+  @tracked actions: ContextualAction[] = [];
```
abeforgit: You should not call this `actions`, see https://discord.com/channels/480462759797063690/480462759797063692/1493969820981661767 and https://chat.semte.ch/group/ember?msg=ZnRAWo7Bhs98NzrpX

It works here cause it's a tracked prop, and not a getter, but probably best to not fly too close to the sun here

abeforgit (changes requested):

kobemertens:
Alternatively we could use ember runloop and just add `// eslint-disable-next-line ember/no-runloop`?

abeforgit:
> The removal of the lifeline dep introduces a new issue
> 
> ```
> Uncaught Error: Assertion Failed: You attempted to update ``state`` on ``SayView``, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
> ```
> 
> I think this is because changing the focus causes a blur transaction to be dispatched which updates the editor state, which causes ember to complain.

Ok so _now_ we actually see the reason we need something like `ember-lifeline` 
The thing that was bothering me wasn't necessarily the extra dep, but the "it didnt work for some reason and I added this seemingly unrelated dep and now it works"

I'll have a think if I can come up with something better, but we might just have to add it back

abeforgit:
@kobemertens I tried to solve it with a statelistener, which I think should be theoretically possible, but didn't get there. I think perhaps we are too entangled with the state in too many places.
So ember-lifeline is indeed the way to go then, but do add a comment explaining what's happening and why we use this here

PR #1375, opened 2026-05-13
update appuniversum to V4

kobemertens:
### Overview
Update ember-appuniversum to version 4.2.1

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/rare-cobras-end.md
+++ b/.changeset/rare-cobras-end.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Update ember-appuniversum to v4.2.1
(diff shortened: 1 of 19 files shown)


abeforgit (changes requested):
@kobemertens this is breaking (you increase the minimum required version of a peerdependency) but your changeset is marked as 'patch'
is there any way we can support both v3 and v4?

kobemertens:
In order to support both versions we would have to add the `disableAuContent` argument conditionally. Which might be doable, although I personally don't have experience with this.

Or maybe we leave the `disableAuContent` out and accept that using v3 there will be a nested `AuContent`? Not sure if it would be a real issue?

abeforgit:
@kobemertens it looks like we can just leave the arguments in, but add an `@glint-expect-error` comment to make it typecheck
The argument does nothing on v4

abeforgit (approved):

PR #1376, opened 2026-06-02
Improve contextual actions menu placement

kobemertens:
### Overview
Previously, the context menu stayed on the bottom of the text after loading, even if there was more space available above. This change makes the menu always appear on the side with the most space.

### How to test/reproduce
Open the menu where the text is almost at the bottom of the screen, the menu should open on top.

### Checks PR readiness
- [x] changelog
- [x] npm lint


piemonkey (approved):
LGTM

PR #1377, opened 2026-06-02
add sticky groups

kobemertens:
### Overview
Adds the ability for `ContextualActionGroups` to be always displayed at the bottom of the context menu.

Required for https://binnenland.atlassian.net/browse/GN-6164

### How to test/reproduce
The dummy context actions now contain a stickied group called "insert". It should always be visible at the bottom of the menu. When no scroll bar is present, the shadow effect should disappear.

### Challenges/uncertainties
Currently, making a action group sticky does not change the group order. I'm not sure if we want to override the `priority` property here or let the implementer deal with making sure that the stickied group is at the bottom.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/cuddly-bananas-sort.md
+++ b/.changeset/cuddly-bananas-sort.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Improve context actions menu placement
--- a/.changeset/spotty-rings-invent.md
+++ b/.changeset/spotty-rings-invent.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Add `sticky` property to `ContextualActionGroups`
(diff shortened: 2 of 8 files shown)


abeforgit:
@kobemertens not sure if this is ready for review, but looks like the sticky area doesn't work as intended in this configuration
image

kobemertens:
Yeah this is related to this comment:
> Currently, making a action group sticky does not change the group order. I'm not sure if we want to override the priority property here or let the implementer deal with making sure that the stickied group is at the bottom.

And I guess in chrome it sorts differently than in firefox... So maybe forcing the stickied groups to be at the bottom makes sense...

abeforgit:
type errors from ci not your fault, looking into it

abeforgit (approved):
good to go, assuming the height limit and optional group headers will come in a different PR

Also, master has some issues atm so wait to merge

PR #1378, opened 2026-06-02
Fix RDFa aware headings to be not isolating

piemonkey:
### Overview
It's unclear whether there was a strong reason to set isolating, but there seems to be no adverse effects. The behaviour is now as expected in text editors in general.

Part of https://binnenland.atlassian.net/browse/GN-6149
Fixes: https://github.com/lblod/frontend-embeddable-notule-editor/issues/289

### Setup
While this was reported on embeddable, just setting `rdfaAware: true` for `headingWithConfig` is enough to see the issue here.

### How to test/reproduce
Headings should behave as expected. For example, pressing enter at the end of a heading will start a new line that is not a heading.

### Challenges/uncertainties
Finding the issue...

### Checks PR readiness
- [x] changelog
- [x] npm lint
- [x] no new deprecations

### DIFF
--- a/.changeset/slow-suns-strive.md
+++ b/.changeset/slow-suns-strive.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': patch
+---
+
+Fix RDFa aware headings to behave as expected when hitting 'enter' to go onto a newline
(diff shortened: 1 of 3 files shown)


abeforgit (approved):
I think the reason these were isolating is that back when we first started using rdfaAware nodes, we still had the idea that a node being a heading would be semantically significant

As such, we tried to treat them like we do the rdfa-blocks. But then later on we realized that didn't really work.

PR #1381, opened 2026-06-08
Contextual actions rework

kobemertens:
### Overview
This rework includes:
- Per-group loading: Instead of showing a single loading bar until all groups have finished loading, every group has its own loading bar.
- Per-group debounce time: When searching, a group can define its own debounce time. (default = 0ms)
- Allow groups without label: these groups automatically get a separator line between them

https://binnenland.atlassian.net/browse/GN-6142

### How to test/reproduce
The test setup now contains a single sticky group without a label.
The dummy actions in test-app/app/dummy-plugins/expose-contextual-actions.ts now can be configured with a loading time and debounce time.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/many-ends-juggle.md
+++ b/.changeset/many-ends-juggle.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+Contextual actions: add per-group loading, per-group debounce time, allow groups without label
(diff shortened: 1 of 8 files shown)


abeforgit (approved):
Seriously impressive!

PR #1382, opened 2026-06-09
Update PR template to our current needs

kobemertens:
Updates the PR so that it aligns more with our current needs


abeforgit:
@kobemertens small hidden tip: for PRs like these, you can add the `no-changeset` label to the pr itself, to make the CI skip that step

PR #1383, opened 2026-06-11
Extract roadsign-regulation-plugin from plugins repo

piemonkey:
### Overview
This doesn't actually get to the AR-design plugin yet, but the roadsign-regulation plugin should be working in the test app. ~~This is a draft as it hasn't been tested in GN or another consuming app. We'll also need to set up the build scripts for allowing releases and dev releases.~~ Should be ready for review now.

Part of https://binnenland.atlassian.net/browse/GN-6154

### Setup
Shouldn't be necessary.

### How to test/reproduce
The test-app has a new page which includes the plugin.

### Challenges/uncertainties
So many weird type problems and import changes.

### Checks PR readiness
- [ ] Tested on Firefox and Chrome-based browsers
- [x] changelog
- [x] npm lint
- [ ] no new deprecation warnings


packages/say-roadsign-regulation-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,2 @@
+# @lblod/ember-rdfa-editor
```
abeforgit: Seems like this should be say-roadsign-regulation-pugin here

pnpm-workspace.yaml
```diff
@@ -196,3 +194,7 @@ allowBuilds:
   '@parcel/watcher': true
   core-js: true
   esbuild: true
+  unrs-resolver: true
+
+minimumReleaseAgeExclude:
+  - '@lblod/ember-rdfa-editor-lblod-plugins@37.2.1-dev.fe38cae7404ec728921e5296ab029df866b3690f'
```
abeforgit: you can use patterns here, so we can just say '@lblod/ember-rdfa-editor-lblod-plugins'
https://pnpm.io/settings#minimumreleaseageexclude

abeforgit (changes requested):
Incredible work!
Installed and ran without issues 

just 2 small comments, otherwise this is a huge step forwards

pnpm-workspace.yaml
piemonkey: Ah, this was automatically added. Not specifying version makes more sense though.

PR #1385, opened 2026-06-23
Placeholder plugin for empty rdfa blocks

kobemertens:
### Overview
Adds a plugin that renders a prosemirror decoration if a rdfa-block is empty and its `placeholder` attr is set.

https://binnenland.atlassian.net/browse/GN-6165

### How to test/reproduce
Open the editable nodes page, add a block rdfa node and set its `placeholder` attr via the editable nodes plugin.

### Challenges/uncertainties
Should we also use this approach for the existing placeholders in some templates? Should we add the functionality that the context menu is automatically opened when actions are available?

### Checks PR readiness
- [x] changelog
- [x] npm lint


packages/ember-rdfa-editor/src/plugins/empty-block-placeholder/index.ts
```diff
@@ -0,0 +1,73 @@
...
+            return false;
+          }
+          return true;
+        });
+        const decorations = emptyRdfaBlocks.map(({ pos, placeholder }) =>
+          Decoration.widget(pos + 2, () => {
+            const el = document.createElement('span');
+            el.classList.add('mark-highlight-manual');
+            el.classList.add('say-placeholder');
+            // Unnecessary check but otherwise TS complains
```
piemonkey: We could use a different type for the 'processed' `emptyRdfaBlock`s but this way with the extra check is probably easier to read :)

packages/ember-rdfa-editor/package.json
```diff
@@ -73,6 +73,10 @@
       "types": "./declarations/plugins/ember-application/index.d.ts",
       "default": "./dist/plugins/ember-application/index.js"
     },
+    "./plugins/empty-block-placeholder": {
```
piemonkey, line 76: Is there a reason to go for a directory and an index.ts here instead of just `plugins/empty-block-placeholder.ts`?

piemonkey (approved):
Looks good and seems to work well

packages/ember-rdfa-editor/package.json
abeforgit: consistency, I think, all the other plugins are structured this way as well
I'm torn on index files myself tbh, on the one hand they're bad for autoimport and make other tooling more complex, on the other i'm not a mega fan of the `file.ts file/other-related-stuff.ts` combo either, it's a bit cluttered

packages/ember-rdfa-editor/src/plugins/empty-block-placeholder/index.ts
abeforgit: If I read the code right, aren't you collecting all rdfa blocks, regardless of whether they have placeholder attributes or not, and then making a decoration for each of them, just with empty textcontent if they don't have a placeholder?

So the check does seem necessary, but also I don't really understand why you're collecting _all_ the empty rdfa blocks

abeforgit (changes requested):

abeforgit:
otherwise works well, just not sure about the decoration logic, see other comment

packages/ember-rdfa-editor/src/plugins/empty-block-placeholder/index.ts
```diff
@@ -0,0 +1,73 @@
...
+          node: PNode;
+          pos: number;
+          placeholder?: string;
+        }[] = [];
+
+        state.doc.descendants((node, pos) => {
+          if (
+            node.type.name === 'block_rdfa' &&
+            isNodeEmpty(node) &&
+            node.attrs['placeholder'] &&
```
kobemertens, line 39: This line should make sure only nodes with a placeholder attribute should be collected I think

kobemertens: [b200d84](https://github.com/lblod/ember-rdfa-editor/pull/1385/commits/b200d848fc906d4b28db88889b57c2859c03a762) <- restructures the code so the check is not needed since I was indeed only collecting non-null nodes

kobemertens:
> Should we also use this approach for the existing placeholders in some templates? Should we add the functionality that the context menu is automatically opened when actions are available?

Any thoughs on this?

abeforgit:
> > Should we also use this approach for the existing placeholders in some templates? Should we add the functionality that the context menu is automatically opened when actions are available?
> 
> Any thoughs on this?

First question: yes absolutely, I've wanted this for a while already
second: not sure, worth a try to see if it feels right

but neither of them are in scope for this pr

PR #1387, opened 2026-06-25
Fix support for passing decision URIs to AR and roadsign plugins

piemonkey:
### Overview
Turns out these never really worked. They should now.

Part of https://binnenland.atlassian.net/browse/GN-6155
Used by https://github.com/lblod/frontend-embeddable-notule-editor/pull/292

### Setup
You either need to link all the modules (editor and both plugins) to GN or you can test without the actual import part in the test app.

### How to test/reproduce
To test, set a `decisionContext` in the lblod-plugins page of the test app or in the agendapoint-edit component of GN. Insert an AR design and check the output RDFa that it links to the passed URI. You should also be able to pass a `decisionContext` to the roadsign regulation card and it should work as expected.

### Challenges/uncertainties
N/A

### Checks PR readiness
- [ ] Tested on Firefox and Chrome-based browsers
- [x] changelog
- [x] npm lint


packages/ember-rdfa-editor/src/plugins/rdfa-info/utils.ts
```diff
@@ -436,15 +436,11 @@ export function addPropertyToNode({
...
-    if (!resourceNodes?.length) {
-      return { initialState: state, transaction: tr, result: false };
-    }
-
-    const properties = getProperties(resourceNodes[0].value);
-    const updatedProperties = addProperty(properties ?? [], property);
 
     // Update the properties of each node that defines the given resource
     resourceNodes.forEach((node) => {
+      const properties = getProperties(node.value);
```
abeforgit, line 442: wow... yeah I can see how this was causing issues
how did it manifest in this case? just curious

abeforgit (approved):

packages/ember-rdfa-editor/src/plugins/rdfa-info/utils.ts
piemonkey: This change is actually so that this function supports there being no resource node, but setting the backlinks anyway. This is needed for the 'insert freely' case.

Now that I think about it, I'm unsure if this is the best way or if it's best to split out the backlink part of the logic to a separate function and in the 'insert freely' case, calling that function instead of this one.

PR #1393, opened 2026-06-30
LPDC-1479: link-plugin - add support for customizing target attribute of outputted html

elpoelma:
### Overview
This PR adds support for customizing the `target` attribute of the link node html output. This way host applications can adjust the default link opening behaviour.
The default value of the `target` attribute remains `_self`.
[LPDC-1479](https://binnenland.atlassian.net/browse/LPDC-1479)

### How to test/reproduce
- Start the dummy app and open the index page
- Insert a link
- Display the outputted html
- Ensure the link opens in a new tab when simply clicking on it (as the `target` setting is set to `_blank` for the index page configuration)

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/forty-poems-argue.md
+++ b/.changeset/forty-poems-argue.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
+link-plugin: add support for customizing `target` attribute of outputted html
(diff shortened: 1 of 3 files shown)


piemonkey (approved):
Would be nice if it was exposed as an option to users, but that's probably a separate ticket. Looks good!

PR #1395, opened 2026-07-06
Slash commands inline search

kobemertens:
### Overview
Allow inline search for the slash commands

### How to test/reproduce
On the dummy plugins page, typing a slash should open the slash menu, and subsequently typing should search the menu. This works almost exactly like in Notion.

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/.changeset/public-bushes-hug.md
+++ b/.changeset/public-bushes-hug.md
@@ -1,0 +1,5 @@
+---
+'@lblod/ember-rdfa-editor': minor
+---
+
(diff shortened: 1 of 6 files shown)


lagartoverde (approved):
Good job maybe nitpicky when we have many actions the search should also maybe filter on that, so if I write marcode only marcode invoegen should be selected.
But I think this is something to think about in the future, just wrote it here to keep in mind

PR #1397, opened 2026-07-29
Remove the embroider instructions

Windvis:
These are no longer needed since v11 of the editor.

I don't think this needs a changelog entry either?


abeforgit:
Thank you!

PR #1399, opened 2026-08-04
Feat/make ar work with regulatory

lagartoverde:
### Overview
This PR mainly does 2 things, first fixes a bug on inserting a AR design at cursor position (previously only used in the preview mode) which made articles be inside each other.
Second it adds all the logic needed to insert AR designs without a besluit being present,

https://binnenland.atlassian.net/browse/GN-6156

### Setup
None

### How to test/reproduce
The test page is setup so now the AR designs are added in this new way. So just empty the editor and insert a new AR design, this should work as expected

### Challenges/uncertainties
The test page change should be reverted before merging the PR is just done for easier testing

### Checks PR readiness
- [x] changelog
- [x] npm lint

### DIFF
--- a/packages/say-roadsign-regulation-plugin/src/plugin/actions/insert-measure.ts
+++ b/packages/say-roadsign-regulation-plugin/src/plugin/actions/insert-measure.ts
@@ -282,9 +282,6 @@ export default function insertMeasure({
         }),
       ),
     ]);
-    console.log('multiple insertion');
-    console.log(args.multipleInsertion);
-    console.log(transaction.selection);
     if (!args.multipleInsertion) {
       console.log('setting new selection');
 
--- a/packages/say-roadsign-regulation-plugin/src/plugin/actions/insert-measure.ts
+++ b/packages/say-roadsign-regulation-plugin/src/plugin/actions/insert-measure.ts
@@ -281,9 +281,11 @@ export default function insertMeasure({
       ),
     ]);
     if (!args.multipleInsertion) {
-      transaction.setSelection(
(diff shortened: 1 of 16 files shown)


packages/say-roadsign-regulation-plugin/src/plugin/actions/insert-measure.ts
```diff
@@ -276,9 +280,11 @@ export default function insertMeasure({
         }),
       ),
     ]);
-    transaction.setSelection(
-      Selection.fromJSON(transaction.doc, resultingSelection.toJSON()),
-    );
+    if (!args.multipleInsertion) {
+      transaction.setSelection(
```
piemonkey: For me, while the selection looks good, it might need a `scrollIntoView()` and a `controller.focus()` to work as expected (at least when tested in Firefox).

piemonkey (commented):
Seems to work well, but I'm not sure about only supporting inserting on the cursor. This makes it possible to put articles inside articles as well as not matching the besluit behaviour. Ideally we'd use the logic used by the 'insert article' sidebar button.

abeforgit:
I agree with @piemonkey, ideally we use the logic from the "insert article" from the RB app when you're working on an RB document:

image

You could use this to find the initial position, and then insert all the articles as you're already doing

lagartoverde:
Done, I did a simplified version of it, as I talked with Rich and we noted that the only case we would have to avoid is the insert article into article 
No need for calculate hierarchy

piemonkey (approved):
Seems to work well with the new logic

PR #1407, opened 2026-09-15
LPDC-1740: link-plugin - adjust DEFAULT_REGEX to also accept capital letters in URL schemes

elpoelma:
### Overview
This PR adjusts the default regex used by `link_input_rule` to also match URLs when (part of) the URL scheme contains capital letters. E.g. `HtTp://`. The capital letters will automatically be converted to lowercase letters by the default link parser.

[LPDC-1740](https://binnenland.atlassian.net/browse/LPDC-1740?atlOrigin=eyJpIjoiYmYwZWMzZjg2ZGJhNGY5Zjk0ODA1ZThhYzI1YTczYjUiLCJwIjoiaiJ9)

### How to test/reproduce
- Start the test app
- Type a plain text link with capital letters in the scheme, e.g. `HTTP://example.org/`
- Ensure that a link object is created with text: `HTTP://example.org/` and href:  `http://example.org/` 

### Checks PR readiness
- [x] changelog
- [x] npm lint


packages/ember-rdfa-editor/src/plugins/link/parser.ts
```diff
@@ -46,44 +46,42 @@ export const defaultLinkParser = ({
...
   }
 
   const phoneNumber = parsePhoneNumber(input, defaultCountryCode);
   if (phoneNumber) {
     const phoneUri = phoneNumber.getURI();
-    return input.startsWith('sms:')
-      ? phoneUri.replace('tel:', 'sms:')
-      : phoneUri;
+    return /^sms:/i.test(input)
+      ? new URL(phoneUri.replace('tel:', 'sms:'))
```
Windvis: Not related to this PR change, but this code was a bit confusing for me without the context in this old comment: https://github.com/lblod/ember-rdfa-editor/blame/9f599d38db748760308bdd70f4921ccae02dded6/packages/ember-rdfa-editor/src/plugins/link/parser.ts#L42

Not sure why that comment was removed, but adding it again might make it less confusing for future readers 😅?

Windvis (approved):

abeforgit (approved):

packages/ember-rdfa-editor/src/plugins/link/parser.ts
elpoelma: oh right, I'm not really sure either... I'll re-add it! Good catch 😃

